PR #9, opened 2019-12-09
Depend on linfa-clustering by path

remram44:
This makes no difference for people getting it from crates.io but simplifies local development greatly.

See [Specifying path dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies) in the Cargo Book.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ keywords = ["machine-learning", "linfa", "ai", "ml"]
 categories = ["algorithms", "mathematics", "science"]
 
 [dependencies]
-linfa-clustering = "0.1"
+linfa-clustering = { path = "linfa-clustering", version = "0.1" }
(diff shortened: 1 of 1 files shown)


LukeMathWalker:
I did this originally, but I moved to depending on a version to release `linfa` on crates.io. I'd be interested to see if `cargo publish` is happy with this when I try to release the next version (path + version).
Anyway, we can get it in for ease of local development and we'll sort it out when we need to get a release out.
Thanks for filing this 👍

remram44:
It works because Cargo rewrites the Cargo.toml file to only have the version, when it packages into a crate (if you run `cargo package` you can see this crate, it has a different `Cargo.toml` and yours as `Cargo.toml.orig`). They are pretty smart :+1:

Issue #7, opened 2019-12-01
Roadmap


LukeMathWalker:
In terms of functionality, the mid-term end goal is to achieve an offering of ML algorithms and pre-processing routines comparable to what is currently available in Python's [`scikit-learn`](https://scikit-learn.org/stable/).

These algorithms can either be:
- re-implemented in Rust;
- re-exported from an existing Rust crate, if available on [crates.io](crates.io) with a compatible interface.

In no particular order, focusing on the main gaps:
- Clustering:
   - [ ] DBSCAN;
   - [ ] Spectral clustering;
   - [ ] Hierarchical clustering;
   - [ ] OPTICS.
- Preprocessing:
   - [ ] PCA
   - [ ] ICA
   - [ ] Normalisation
   - [ ] CountVectoriser
   - [ ] TFIDF
- Supervised Learning:
   - [ ] Linear regression;
   - [ ] Ridge regression;
   - [ ] LASSO;
   - [ ] Support vector machines;
   - [ ] Nearest Neighbours;
   - [ ] Gaussian processes;
   - [ ] Decision trees;
   - [ ] Random forest
   - [ ] Naive Bayes

The collection is on purpose loose and non-exhaustive - if there is an ML algorithm that you find yourself using often on a day to day, please feel free to contribute it :100:

Nimpruda:
Hi, I'm eager to help I'll take Linear regression, Lasso and ridge.

LukeMathWalker:
Cool! I worked a bit on linear regression a while ago - you can find a very vanilla implementation of it here: https://github.com/rust-ndarray/ndarray-examples/tree/master/linear_regression @Nimpruda

InCogNiTo124:
What does `Normalization` mean, is it like `StandardScaler` or something else?

LukeMathWalker:
Exactly @InCogNiTo124.

ADMoreau:
This is an interesting project and I will work on the PCA implementation

nestordemeure:
I am the author of the [friedrich](https://crates.io/crates/friedrich) crate which implements Gaussian Processes.

While it is still a work in progress, it is fully featured and I would be happy to help integrate it into the project if you have directions to do so.

LukeMathWalker:
That would be awesome @nestordemeure - I'll have a look at the project and I'll get back to you! Should I open an issue on `friedrich`'s repository when I am ready? Or would you prefer it to be tracked here on the `linfa` repository?

[...]

Issue #8, opened 2019-12-08
Clarify conventions for public interfaces


bplevin36:
It would be useful, especially to potential contributors, to have a unified description of how public interfaces should be structured.  At first glance, I assumed we would be attempting to stay as close to sklearn's conventions as is reasonable with Rust's conventions and syntax.  Looking at the KMeans implementation, however, shows significant departure from sklearn's parameter naming scheme, and introduces several additional types for Hyperparameter management.

I think it's important in the long run for the public interfaces to be both intuitive and consistent.  With that in mind, I think we should:

1. Start a discussion about design choices for the public interfaces.  I personally am not sure that the utility of introducing HyperParams structs and accompanying factory functions justifies the additional API complexity.  But I could absolutely be wrong about that.  I'd just like to see some rationale.

2. Write up the conclusions of that discussion in a design doc.  This doesn't need to be that complex or in-depth, just a basic statement of conventions and design philosophy to make it easier for contributors.

LukeMathWalker:
I agree in principle @bplevin36, but I disagree in practice - if nothing else, because I tried to do this already in https://github.com/rust-ml/classical-ml-discussion/pull/2 and I quickly realised that it's difficult to have design discussions when there are not enough concrete examples. 
They are difficult to follow and it's difficult to understand if you are going in the right direction or not.
It's extremely valuable to be able to quickly benchmark against existing functionality what it would look like to go for a certain trait definition instead of another. 

For the current stage of the project, I would focus on algorithm implementation without spending too much time on traits/interfaces until we have at least 3/4 algorithms in each relevant class (supervised learning, unsupervised learning, pre-processing functions, etc.). 
Refactoring to a common interface takes a fraction of the effort of the initial heavy-lifting required to write, test and document a new model implementation.

I do as well understand that a **minimal** set of directives/guidelines is required to avoid having contributors wander in incompatible directions.
At the moment, I would probably suggest the following as contributor guidelines:

- when you need to reach for a data-structure, favor re-use over adding a new dependency (e.g. `ndarray` for dense matrices);
- leverage the strength of the type system to ensure compile-time correctness;
- don't abuse the type system - it's ok if some checks have to be performed at runtime (no `typenum` 😅 )

I would add that imitating the API of `scikit-learn` is not a goal for `linfa`: `scikit-learn` is a great library, but its first release came out in 2007 and the authors themselves would probably make different choices if they had a chance to redesign the API today. 
Rust is not Python - we should leverage its own strength.

[...]

PR #10, opened 2019-12-12
[WIP] Adds Linear, Ridge and Lasso regressions

Nimpruda:
see #7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ categories = ["algorithms", "mathematics", "science"]
 
 [dependencies]
 linfa-clustering = "0.1"
+linfa-supervised = {path = "linfa-supervised"}
 
 [dev-dependencies]
 ndarray = { version = "0.13" , features = ["rayon", "serde", "approx"]}
@@ -22,7 +23,8 @@ rand_isaac = "0.2.0"
 ndarray-npy = { version = "0.5", default-features = false }
 
 [workspace]
-members = ["linfa-clustering"]
+members = ["linfa-clustering", "linfa-supervised"]
 
 [profile.release]
 opt-level = 3
+lto = true
--- a/linfa-supervised/Cargo.toml
+++ b/linfa-supervised/Cargo.toml
@@ -8,9 +8,9 @@ license = "MIT/Apache-2.0"
 
 repository = "https://github.com/LukeMathWalker/linfa"
 
-keywords = ["supervised", "machine-learning", "linfa", "regression", "linear"]
+keywords = ["supervised", "machine-learning", "linfa", "regression", "linear", "ridge", "lasso"]
 categories = ["algorithms", "mathematics", "science"]
 
 [dependencies]
-ndarray = { version = "0.13" , features = ["rayon", "approx"] }
-ndarray-linalg = "0.12"
+ndarray = { version = "0.13" , features = ["rayon"] }
+ndarray-linalg = { version = "0.12", features = ["openblas"] }
--- a/linfa-supervised/Cargo.toml
+++ b/linfa-supervised/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "linfa-supervised"
+version = "0.1.0"
+authors = ["Khalil HADJI <khalil.hadji@outlook.fr>"]
+edition = "2018"
+description = "A collection of supervised learning algorithms"
+license = "MIT/Apache-2.0"
+
+repository = "https://github.com/LukeMathWalker/linfa"
+
+keywords = ["supervised", "machine-learning", "linfa", "regression", "linear"]
+categories = ["algorithms", "mathematics", "science"]
+
+[dependencies]
+ndarray = { version = "0.13" , features = ["rayon", "approx"] }
+ndarray-linalg = "0.12"
--- a/linfa-supervised/examples/main.rs
+++ b/linfa-supervised/examples/main.rs
@@ -1,0 +1,40 @@
+use linfa_supervised::LinearRegression;
+use linfa_supervised::RidgeRegression;
+use ndarray::array;
+
+fn linear_regression() {
+    let mut linear_regression = LinearRegression::new(false);
+    let x = array![[1.0], [2.0], [3.0], [4.0]];
+    let y = array![1.0, 2.0, 3.0, 4.0];
+    linear_regression.fit(&x, &y);
+    let x_hat = array![[6.0], [7.0]];
+    println!("{:#?}", linear_regression.predict(&x_hat));
+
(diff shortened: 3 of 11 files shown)


Cargo.toml
```diff
@@ -13,6 +13,8 @@ keywords = ["machine-learning", "linfa", "ai", "ml"]
 categories = ["algorithms", "mathematics", "science"]
 
 [dependencies]
+
+linfa-supervised = {path = "linfa-supervised"}
```
remram44: You need to add a version number, of publishing the `linfa` crate to crates.io will fail :wink:
```suggestion
linfa-supervised = { path = "linfa-supervised", version = "0.1" }
```

linfa-supervised/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+license = "MIT/Apache-2.0"
+
+repository = "https://github.com/LukeMathWalker/linfa"
+
+keywords = ["supervised", "machine-learning", "linfa", "regression", "linear", "ridge", "lasso"]
+categories = ["algorithms", "mathematics", "science"]
+
+[dependencies]
+ndarray = { version = "0.13" , features = ["rayon"] }
+ndarray-linalg = { version = "0.12", features = ["openblas"] }
```
LukeMathWalker: We don't want to force the usage of a certain BLAS implementation over the other ones (e.g. `intel-mkl`) - I would suggest to use a features-strategy similar to what I put down here: https://github.com/rust-ndarray/ndarray-examples/blob/d7d43e018cade1f8bf0d92220081ba66963bab07/linear_regression/Cargo.toml#L8

linfa-supervised/examples/main.rs
```diff
@@ -0,0 +1,40 @@
...
+
+fn linear_regression() {
+    let mut linear_regression = LinearRegression::new(false);
+    let x = array![[1.0], [2.0], [3.0], [4.0]];
+    let y = array![1.0, 2.0, 3.0, 4.0];
+    linear_regression.fit(&x, &y);
+    let x_hat = array![[6.0], [7.0]];
+    println!("{:#?}", linear_regression.predict(&x_hat));
+
+    let mut linear_regression2 = LinearRegression::new(true);
```
LukeMathWalker: I would suggest splitting this example (and the one below) in two separate examples, with a descriptive name (e.g. with/without intercept).
It would be ideal to have them in separate files as well under the `examples` folder.

linfa-supervised/src/linear_regression/algorithm.rs
```diff
@@ -0,0 +1,119 @@
+#![allow(non_snake_case)]
+use ndarray::{stack, Array, Array1, ArrayBase, Axis, Data, Ix1, Ix2};
+use ndarray_linalg::Solve;
+/* I will probably change the implementation for an enum for more type safety.
+I have to make sure, it is a great idea when it comes to pyhton interoperability
```
LukeMathWalker: There is considerable freedom in how to wrap the Rust version for Python consumption - as detailed in #8, we shouldn't let Python move our Rust design in directions which are not idiomatic. The wrapping code can do the bridging when required 😀 

So I'd definitely suggest to go with the commented out version, which uses an enum (`LinearRegression::new(false)` is much more confusing than `LinearRegression::new(Intercept::NoIntercept)`).

linfa-supervised/src/ridge_regression/algorithm.rs
```diff
@@ -0,0 +1,44 @@
+#![allow(non_snake_case)]
+use ndarray::{Array, Array1, ArrayBase, Data, Ix1, Ix2};
+use ndarray_linalg::Solve;
+/* The difference between a linear regression and a Ridge regression is
+ that ridge regression has an L2 penalisation term to having some features
```
LukeMathWalker: Typo? `avoid having`?

linfa-supervised/src/ridge_regression/algorithm.rs
```diff
@@ -0,0 +1,44 @@
...
+ b = (X^T X + aI)X^T y with a being the regularisation/penalisation term
+*/
+
+pub struct RidgeRegression {
+    beta: Option<Array1<f64>>,
+    alpha: f64,
+}
+
+impl RidgeRegression {
+    pub fn new(alpha: f64) -> RidgeRegression {
```
LukeMathWalker: Shouldn't we have an `intercept` parameter here as well?

linfa-supervised/src/utils.rs
```diff
@@ -0,0 +1,6 @@
+trait GradientDescent {
```
LukeMathWalker: Is this used anywhere?

linfa-supervised/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "linfa-supervised"
```
LukeMathWalker: I would probably prefer this to be named `linfa-linear` - we will have many more algorithm belonging to the family of supervised learning and we don't want to have all of them in the same sub-crate.

linfa-supervised/src/linear_regression/algorithm.rs
```diff
@@ -0,0 +1,119 @@
...
+        A: Data<Elem = f64>,
+    {
+        let (n_samples, _) = X.dim();
+
+        // If we are fitting the intercept, we need an additional column
+        if self.fit_intercept {
+            let dummy_column: Array<f64, _> = Array::ones((n_samples, 1));
+            let X = stack(Axis(1), &[dummy_column.view(), X.view()]).unwrap();
+            match &self.beta {
+                None => panic!("The linear regression estimator has to be fitted first!"),
```
LukeMathWalker: This is not ideal - can we refactor the way we build `LinearRegression` to make sure that `beta` is always there when a user calls `predict`?

The easiest way to do this is re-using the same approach I put down in `KMeans`: a structure to hold the hyperparameters of the model (built with/without builder pattern, depending on how many hyperparameters we have there) and a `fit` method on it that returns a fitted `LinearRegression` - basically, using `fit` as constructor. In this way we can be sure that `beta` is there and we have one less `panic` condition 😁

LukeMathWalker:
Thanks for working on it @Nimpruda! 
I left some comments here and there to give you pointers on the next step to make this PR ready to be merged - it's already half-way there 😁 

I see we have some issues in CI due to `ndarray-linalg` requiring a BLAS implementation to build correctly - I'll work in a separate PR to get this issue solved so that your branch can compile correctly on Travis 👍

Nimpruda:
I'll work on it whenever I have time! Thanks for all the feedback

bytesnake:
superseded by #20

PR #12, opened 2019-12-24
WIP: Initial commit of DBScan

xd009642:
Looked to the k-means implementation to keep the design consistent. Comments need to be filled in and tests. Also distance is currently just euclidean distance but it's probably a good idea for there to some sort of distance trait or enum so people can pass in what distance function they want for this or other algorithms.

I'll carry on filling in the rest just figured early feedback was better than later :smile:


xd009642:
Also, the dbscan implementation in sklearn isn't one that you can fit then predict, fitting is part of the prediction stage so I've replicated that. Incremental DBScan is an extension over the original algorithm

LukeMathWalker:
Nice! I'll have a look straight away :grin: 

There are other interesting implementation in Python-land that we should keep in mind - in particular, HDBSCAN. 
They also support assigning a cluster to a new point after the initial fitting round - see [here](https://hdbscan.readthedocs.io/en/latest/faq.html#q-i-want-to-predict-the-cluster-of-a-new-unseen-point-how-do-i-do-this).

linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+fn find_neighbors(
+    candidate: &ArrayBase<impl Data<Elem = f64> + Sync, Ix1>,
+    observations: &ArrayBase<impl Data<Elem = f64> + Sync, Ix2>,
+    eps: f64,
+) -> Vec<usize> {
+    let mut res = vec![];
+    for (i, obs) in observations.axis_iter(Axis(1)).enumerate() {
+        if distance(candidate, &obs) < eps {
```
LukeMathWalker: You can use [`sq_l2_dist`](https://docs.rs/ndarray-stats/0.3.0/ndarray_stats/trait.DeviationExt.html#tymethod.sq_l2_dist) here - no need to reimplement it.

linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -0,0 +1,65 @@
+use crate::dbscan::hyperparameters::DBScanHyperParams;
+use ndarray::{Array1, ArrayBase, Axis, Data, Ix1, Ix2};
+
+pub fn predict(
+    hyperparameters: DBScanHyperParams,
+    observations: &ArrayBase<impl Data<Elem = f64> + Sync, Ix2>,
```
LukeMathWalker: Do we need a `Sync` bound here (and in the other functions)? At a first glance it doesn't seem to me that we are using any threading right now.

linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -0,0 +1,65 @@
...
+    let mut latest_id = 0;
+    for (i, obs) in observations.axis_iter(Axis(1)).enumerate() {
+        if result[i].is_some() {
+            continue;
+        }
+        let n = find_neighbors(&obs, observations, hyperparameters.tolerance());
+        if n.len() < hyperparameters.minimum_points() {
+            continue;
+        }
+        // Now go over the neighbours adding them to the cluster
```
LukeMathWalker: Before initialising the `search_queue`, shouldn't we perform the cluster assignment? I.e. `result[i] = latest_id`?
Otherwise `obs` will end up in `search_queue` and we might be duplicating a bunch of work.

LukeMathWalker: Basically, as soon as we know that it has enough neighbors.

linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -0,0 +1,65 @@
...
+                .filter(|x| result[[**x]].is_none() && !search_queue.contains(x))
+                .copied()
+                .collect::<Vec<_>>();
+
+            search_queue.append(&mut n);
+        }
+        latest_id += 1;
+    }
+    result
+}
```
LukeMathWalker: This looks good! 
Improvements we could make from a readability perspective:
- `n` -> `neighbours`
- `cand` -> `candidate`
- `result` -> `cluster_memberships`
- `latest_id` -> `current_cluster_id`

linfa-clustering/src/dbscan/algorithm.rs
xd009642: Didn't realise that existed! (Think I now have to do some changes to `ndarray-vision` to use it there :sweat_smile: )

linfa-clustering/src/dbscan/algorithm.rs
xd009642: No we don't, removed

linfa-clustering/src/dbscan/algorithm.rs
xd009642: Done :+1:

linfa-clustering/src/dbscan/algorithm.rs
xd009642: Will do, though I've been avoiding writing `neighbours` as it's the UK spelling. Though I'm happy to move everything to the superior UK spelling :wink:

linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    #[test]
+    fn non_cluster_points() {
+        let params = DBScanHyperParams::new(4).build();
+
+        let data: Array2<f64> = Array2::zeros((2, 3));
+
+        let labels = predict(&params, &data);
+        assert!(labels.iter().all(|x| x.is_none()));
```
LukeMathWalker: This should probably be its own test - we are basically testing that if the dataset is smaller than `min_points` than every point is a noise point, is that correct?

linfa-clustering/src/dbscan/algorithm.rs
LukeMathWalker: If you put the assignment before `search_queue`, you can avoid `&& **x != i` in the filter statement :grin:

linfa-clustering/src/dbscan/algorithm.rs
LukeMathWalker: I mean, I am not an English native speaker, so I am happy with either as long as its consistent across the crate :stuck_out_tongue:

linfa-clustering/src/dbscan/algorithm.rs
xd009642: yeah I can move it into it's own

LukeMathWalker:
It would be interesting to add a benchmark for DBSCAN as well - I believe we can do some optimisations in a couple of places, but I'd avoid starting with them before we can measure if the gain is real :+1:

xd009642:
So I realised when writing a quick benchmark that predict was a bad function name as a free function because of the public reexports, so I've currently renamed it to `dbscan` benchmark coming soon (basically copying the k_means one)

LukeMathWalker:
Yeah, we can work on the naming - I would probably suggest to wrap it in a struct anyway, for ease of saving/loading as well as future extensibility. But we can figure this out once we nailed down the algorithm implementation (I think we are close :grin:)

xd009642:
Right benchmark added which should look very familiar

linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+        let neighbors = find_neighbors(&obs, observations, hyperparameters.tolerance());
+        if neighbors.len() < hyperparameters.minimum_points() {
+            continue;
+        }
+        // Now go over the neighbours adding them to the cluster
+        cluster_memberships[i] = Some(current_cluster_id);
+        let mut search_queue = neighbors
+            .iter()
+            .filter(|x| cluster_memberships[[**x]].is_none())
```
LukeMathWalker: Given that we always go and access the row corresponding to an index in the neighbors collection, I think it would be interesting to try to return `Vec<(usize, &ArrayBase<impl Data<Elem=64>, Ix1>` from `find_neighbors`.
In that way we should be able to save on a bunch of index accesses, bound checks and so forth.

xd009642: Putting the reference to the `ArrayBase` in the vector means lifetimes are needed. Tried to do it and then I need to add the Sized trait to things etc.

xd009642:
```Rust
error[E0038]: the trait `ndarray::data_traits::Data` cannot be made into an object
  --> linfa-clustering/src/dbscan/algorithm.rs:49:1
   |
49 | / fn find_neighbors<'a>(
50 | |     candidate: &'a ArrayBase<impl Data<Elem = f64>, Ix1>,
51 | |     observations: &'a ArrayBase<impl Data<Elem = f64>, Ix2>,
52 | |     eps: f64,
...  |
60 | |     res
61 | | }
   | |_^ the trait `ndarray::data_traits::Data` cannot be made into an object
   |
   = note: the trait cannot require that `Self : Sized`
```

LukeMathWalker: Just out of curiosity, could you try using generic notation instead of `impl` notation for that function?

xd009642: Got it! And some improvement, here's the criterion printout:
```
     Running /home/xd009642/code/rust/linfa/target/release/deps/dbscan-abaeb88fc2d9076d
dbscan/10               time:   [580.82 ns 582.51 ns 584.38 ns]                       
                        change: [+1.4560% +2.0619% +2.6529%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe
dbscan/100              time:   [3.8319 us 3.8423 us 3.8564 us]                        
                        change: [-0.3080% +0.0472% +0.3940%] (p = 0.80 > 0.05)
                        No change in performance detected.
Found 14 outliers among 100 measurements (14.00%)
  9 (9.00%) high mild
  5 (5.00%) high severe
dbscan/1000             time:   [36.916 us 36.989 us 37.069 us]                         
                        change: [-6.6344% -5.4607% -4.3047%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  7 (7.00%) high mild
  3 (3.00%) high severe
dbscan/10000            time:   [375.52 us 376.70 us 378.03 us]                         
                        change: [-11.490% -9.4285% -7.4747%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe
```

xd009642:
Here's a picture of the change for 10,000 points

Screenshot from 2019-12-24 17-04-37

linfa-clustering/src/dbscan/algorithm.rs
LukeMathWalker: That looks nice :grin: 
The other optimisation I was considering is adding `cluster_memberships` as a parameter to `find_neighbors` and iterating over both the observations and the memberships using `Zip`, thus returning directly `search_queue`. In this way we avoid having to allocate two vectors (first for neighbors than for the indexes) and we also save a bunch of indexed accesses.

xd009642: Just directly returning the search queue would change the behaviour when it's looking for the first point in the `cluster_id` because the number of neighbours could then drop below `min_points`. There will be a way around that though I'll have a play/think

xd009642: Figured it out, easily enough. Another improvement though not as dramatic this time.

Screenshot from 2019-12-24 18-50-15

linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -0,0 +1,116 @@
...
+    observations: &'a ArrayBase<impl Data<Elem = f64>, Ix2>,
+    eps: f64,
+    clusters: &Array1<Option<usize>>,
+) -> (usize, Vec<(usize, ArrayView<'a, f64, Ix1>)>) {
+    let mut res = vec![];
+    let mut count = 0;
+    for (i, obs) in observations.axis_iter(Axis(1)).enumerate() {
+        if candidate.l2_dist(&obs).unwrap() < eps {
+            count += 1;
+            if clusters[[i]].is_none() {
```
LukeMathWalker: We are still using an index to access the cluster from `clusters`.
I expect a more significant performance improvement using `Zip`, so that we already have the value at hand.

linfa-clustering/src/dbscan/algorithm.rs
LukeMathWalker: Left a comment which explains better what I was trying to achieve :pray:

linfa-clustering/src/dbscan/algorithm.rs
xd009642: Ah I see. Tried it out and didn't see an improvement except with the small values:

```
dbscan/10               time:   [597.95 ns 599.53 ns 601.28 ns]                       
                        change: [-6.4851% -6.0215% -5.5629%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  6 (6.00%) high mild
  7 (7.00%) high severe
dbscan/100              time:   [3.8862 us 3.8933 us 3.9013 us]                        
                        change: [-0.2848% +0.1601% +0.6346%] (p = 0.51 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
dbscan/1000             time:   [36.933 us 37.005 us 37.095 us]                         
                        change: [-0.1611% +0.1888% +0.5086%] (p = 0.30 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  5 (5.00%) high mild
  3 (3.00%) high severe
dbscan/10000            time:   [373.45 us 374.56 us 375.95 us]                         
                        change: [-0.5209% +0.1337% +0.9362%] (p = 0.73 > 0.05)
                        No change in performance detected.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe
```

xd009642: I used the `zip` method in the standard library `Iterator` trait not `ndarray::Zip` as I didn't expect there to be a different and the iterator one was easier to write quicker. I can try again with `ndarray::Zip` if there's any difference in implementation that could make it faster

Screenshot from 2019-12-24 19-30-45

LukeMathWalker: Just given it a shot - no significant differences.
The one thing we could do though is removing that `count` variable - accessing the length of a vector should be `O(1)` if I am not mistaken, so no need to keep track of it separately (in this way we can simply return a `Vec` from that function and use `.len()` to get `neighbors_count`).

LukeMathWalker:
I think we are there from an algorithmic point of view :grinning: 

Next steps to get this ready to be merged:
- an example in the `examples` category;
- wrapping it in a struct for consistency/serialisation/future extensibility.

Then we are good to go :rocket:

linfa-clustering/src/dbscan/algorithm.rs
xd009642: The count is there because count can be greater than the length of the vector. Remember how the list of neighbours was originally filtered to get the search queue because it could have points already allocated. The returned vector is the subset of neighbours that aren't yet allocated to a cluster

LukeMathWalker: Yes, you are right :+1:

xd009642:
> I think we are there from an algorithmic point of view

I realised a mistake in my implementation of the algorithm! Only minor but the number of neighbours needs to be taken into account for each element in the search queue. I've added that and an example. Currently writing the doc comments then I'll push something

linfa-clustering/src/dbscan/hyperparameters.rs
```diff
@@ -0,0 +1,60 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+/// The set of hyperparameters that can be specified for the execution of
+/// the [DBSCAN algorithm](struct.Dbscan.html).
+pub struct DbscanHyperParams {
+    /// Distance between points for them to be considered neighbours.
+    tolerance: f64,
```
LukeMathWalker: `DbscanHyperParams`'s fields are private - we need to copy these doc comments over to corresponding methods, both in the builder and accessor methods.
We should also document the default value in the builder methods.

xd009642: Done :smile:

LukeMathWalker:
Merged - thanks for all your work here @xd009642! :pray:

PR #17, opened 2020-03-18
Add diffusion map dimensionality reduction and simple Principal Component Analysis

bytesnake:
Spectral Clustering consists of two steps. First a dimensionality reduction finds important pattern in a graph between the points (by projecting the graph), then a clustering algorithm runs on the embedding. The technique is non-linear because cluster can be distributed such that they are nested inside each other, for example swiss-rolls or nested rings. A linear projection (like PCA) fails in such cases.
Gaussian and polynomial kernels are implemented and projected with diffusion maps. `LOBPCG` is used for finding the `k` largest eigenvalues.

Things todo before merging:
* [x] wait for LOPBCG PR getting merged https://github.com/rust-ndarray/ndarray-linalg/pull/184
* [x] implement similarity matrix kernel function (with fast nearest neighbour algorithms and sparsity constraints)
* [x] simple PCA implementation (@ADMoreau)


bytesnake:
haven't looked at this for ages, but should be ready to merge. Perhaps I will find time over the weekend to give it a second glance

quietlychris:
@bytesnake If you don't have permissions to merge this, just let me know and I can either add you as a rust-ml owner (if you're interested in that), or otherwise would just be happy to make the merge whenever you're ready.

bytesnake:
@quietlychris yup you can add me and I will see how I can be of help

PR #20, opened 2020-05-20
Add linfa-linear package with ordinary least squares regression

paulkoerbitz:
This PR adds the linfa-linear crate with an ordinary least squares regression model.

It is in spirit similar to [PR 10](https://github.com/LukeMathWalker/linfa/pull/10) which looks like it has stalled. It also addresses the issues discussed in this PR and ads example code and tests.


linfa-linear/src/lib.rs
```diff
@@ -0,0 +1,366 @@
...
+}
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum Options {
+    None,
+    FitIntercept,
+    FitInterceptAndNormalize,
+}
+
+fn fit_intercept(options: Options) -> bool {
```
bytesnake, line 57: can you implement this as a member of `Options`?

linfa-linear/src/lib.rs
```diff
@@ -0,0 +1,366 @@
...
+    /// To change this, call `fit_intercept()` or
+    /// `fit_intercept_and_normalize()` before calling `fit()`.
+    pub fn new() -> LinearRegression {
+        LinearRegression {
+            options: Options::None,
+        }
+    }
+
+    /// Configure the linear regression model to fit an intercept.
+    pub fn fit_intercept(mut self) -> Self {
```
bytesnake, line 84: I don't really like the name, because `fit_intercept` sounds like a special function of `fit`, but is actually part of the builder. Perhaps it would be easier to have a `pub fn options(&mut self, options: Options)` or alternatively `with_intercept`.

bytesnake:
considering that the required changes are minimal, I will merge this PR now. If you have something to amend, please do so (otherwise any PR for LASSO regression is welcome too :smile: )

PR #23, opened 2020-07-15
[WIP] Add common metrics for model evaluations

bytesnake:
This PR adds different metrices for classification and regression tasks. 
Common metrics for classification are 
 * [x] accuracy, precision, recall, F1-score
 * [x]  ROC, AUC ROC
 * [x] Matthew Correlation Coefficient (MCC) 
 * [x] confusion matrix

Common metrics for regression are 
 * [x] max/absolute/square/log error, explained variance


bytesnake:
My current goal for the interface is something like this
```rust
prediction
  .apply_threshold(0.5)
  .weight_samples(weights)
  .accuracy(ground_truth)
```

wangjiawen2013:
Hi, @bytesnake 
the roc of linfa logistic is not the same as that of python, do you know why ?
```
let roc = predictions.roc(&ground_truths);  // for rust
fpr, tpr, thresh = roc_curve(groundtruth, predictions) // for python
```

wangjiawen2013:
This is from python, it's different from  https://docs.rs/linfa/latest/src/linfa/metrics_classification.rs.html#433 , which shows an example of `test_roc_curve()`
Out[20]: 
   FPR   TPR  Threshold
0  0.0  0.00        1.9
1  0.0  0.25        0.9
2  0.0  0.75        0.7
3  0.5  0.75        0.5
4  0.5  1.00        0.3
5  1.0  1.00        0.1

PR #27, opened 2020-07-29
Add Support Vector Machines with SMO

bytesnake:
This PR adds support for SVM with [Sequential Minimal Optimization](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tr-98-14.pdf) for C/nu/one-class classification and epsilon/nu regression. The algorithm is similar to that of [libsvm](https://www.csie.ntu.edu.tw/~cjlin/libsvm/).

 * [x] Implementation
 * [x] At least one test for each flavor
 * [x] Optimize discriminant for linear kernel
 * [x] Example, see [here](https://github.com/rust-ml/linfa/blob/ba2197ed0a26bb321ff5bbe71769915d8c2a5be5/linfa-svm/examples/winequality.rs) for wine quality classifier
 * [x] Documentation
 * [x] More testing

The API is intentionally spartan right now, because it is not clear how multi-label classification or datasets will be structured in the future. Further the interaction between a kernel matrix and streaming data is problematic, and I think it will be necessary to create special traits for partially trainable models (like NN) and one-fit models, which needs the whole dataset a-priori.
I left out all multi-label extension, because it would be nice to create a wrapper, which can transform any binary classification model into a multi-label classification one. Also bagging would be a nice extension.

I added tests with linearly separable classes, convoluted rings and polynomials in order to test different kernel methods.

Two things I learned while implementing:
1. You can do `.collect::<Array1<_>>()`, which is immensely useful
2. With [NdProducer](https://docs.rs/ndarray/0.13.1/ndarray/trait.NdProducer.html) you can accept `ArrayView`s as well as slices

would be nice to collect useful recipes somewhere :)

There is also a case where clippy is not super useful. I'm iterating over three different variables with the same index in the optimization, but clippy gives me a warning, that I should iterate over the first and use an index for the remaining .. which I think is by no means more readable
```
warning: the loop variable `k` is used to index `dist_j`
   --> linfa-svm/src/solver_smo.rs:346:26
    |
346 |                 for k in 0..self.nactive() {
    |                          ^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_range_loop
help: consider using an iterator
    |
346 |                 for (k, <item>) in dist_j.iter().enumerate().take(self.nactive()) {
    |                     ^^^^^^^^^^^    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```


linfa-kernel/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "linfa-kernel"
+version = "0.1.0"
+authors = ["Lorenz Schmidt <bytesnake@mailbox.org>"]
+description = "Kernel methods for non-linear algorithms"
+edition = "2018"
+license = "MIT/Apache-2.0"
+
+repository = "https://github.com/LukeMathWalker/linfa"
```
quietlychris, line 9: Should this be `repository = "https://github.com/rust-ml/linfa"`?

linfa-svm/examples/winequality.rs
```diff
@@ -0,0 +1,85 @@
...
+    let array = reader.deserialize_array2_dynamic()?;
+    Ok(array)
+}
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // read in wine-quality dataset
+    let data = read_array("../datasets/winequality-red.csv.gz")?;
+
+    // calculate number of training and validation points
+    let npoints = data.len_of(Axis(0));
```
quietlychris, line 31: It might useful to add a comment like 
```
// .csv data is two-dimensional: Axis(0) denotes y-axis (rows), Axis(1) denotes x-axis (columns)
```
to make this more clear to people that aren't familiar with the ndarray's slicing/axis conventions

linfa-kernel/Cargo.toml
bytesnake: yes that's out of date

linfa-svm/examples/winequality.rs
bytesnake: yeah you're right will improve the comments .. they were meant as a remainder for myself, but the audience is obviously different here

bytesnake: okay added some comments in the example

quietlychris: Looked through most recent commits, the new comments and updated fields look great

bytesnake:
will wait until Paul commented, but I'm happy to move forward then :tada: after the merge we finally have at least one example in each class of transformers/unsupervised/supervised methods and experiment with the API

quietlychris:
Sounds great!

PR #34, opened 2020-08-08
Run Clippy in CI

paulkoerbitz:
This PR ads a job to run clippy in the CI. Clippy runs over all targets and features (so that it also checks test code and - in the future - conditionally compiled code). For now, clippy is run with default settings which IMO gave reasonable results on the linfa code base. The job is set to fail if any warnings are given (as it would otherwise be useless to catch transgressions).

There is one rule which gave three warnings I didn't fix: new_ret_no_self, which checks if a method called `new` returns a value of the type it's implemented on. I do think this is a reasonable rule (and I do find the places it warns about at least slightly confusing/awkward), but I didn't want to change the API without prior discussion in this housework PR. So I added `#[allow(clippy::new_ret_no_self)]` to the offending places and added a comment.

I would propose that we treat clippy this way going forward: Review the suggestions and adopt them where it makes sense (which is hopefully going to be the vast majority of times), but don't follow them slavishly - if a rule doesn't make sense we can ignore it via an `#[allow(...)` attribute and add a comment explaining why this is so. (I'm intending to add such stuff to a contributing doc going forward.) If we find a rule that doesn't make sense or consistently gives bad advice we disable it.

CC @quietlychris @bytesnake 

PS: To see the rules currently enabled go [here](https://rust-lang.github.io/rust-clippy/master/) and uncheck "Pedantic" "Cargo" and "Nursery" on the RHS and the "Allow" level on the LHS.

### DIFF
--- a/linfa-clustering/src/dbscan/hyperparameters.rs
+++ b/linfa-clustering/src/dbscan/hyperparameters.rs
@@ -39,6 +39,8 @@ impl DbscanHyperParams {
     ///
     /// Defaults are provided if the optional parameters are not specified:
     /// * `tolerance = 1e-4`
+    // Violates the convention that new should return a value of type `Self`
+    #[allow(clippy::new_ret_no_self)]
     pub fn new(min_points: usize) -> DbscanHyperParamsBuilder {
         DbscanHyperParamsBuilder {
             min_points,
--- a/src/metrics_classification.rs
+++ b/src/metrics_classification.rs
@@ -103,7 +103,7 @@ pub struct ConfusionMatrix<A> {
 
 impl<A> ConfusionMatrix<A> {
     fn is_binary(&self) -> bool {
-        self.matrix.shape() == &[2, 2]
+        self.matrix.shape() == [2, 2]
     }
 
     /// Precision score, the number of correct classifications for the first class divided by total
@@ -292,11 +292,11 @@ impl<A> ConfusionMatrix<A> {
 impl<A: fmt::Display> fmt::Debug for ConfusionMatrix<A> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let len = self.matrix.len_of(Axis(0));
-        write!(f, "\n")?;
+        writeln!(f)?;
         for _ in 0..len * 4 + 1 {
             write!(f, "-")?;
         }
-        write!(f, "\n")?;
+        writeln!(f)?;
 
         for i in 0..len {
             write!(f, "| ")?;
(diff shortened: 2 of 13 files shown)


bytesnake:
sorry for the delay, will look into this now

linfa-clustering/src/dbscan/hyperparameters.rs
```diff
@@ -39,6 +39,8 @@ impl DbscanHyperParams {
     ///
     /// Defaults are provided if the optional parameters are not specified:
     /// * `tolerance = 1e-4`
+    // Violates the convention that new should return a value of type `Self`
```
bytesnake, line 42: I would add a `IntoValidatedParams` trait, and have a single structure describing the parameters. The receiver function can then call `validate` of `IntoValidatedParams`, or just panics in the setter functions. But I think as well, that we should add exceptions here for now

src/metrics_classification.rs
```diff
@@ -493,19 +493,48 @@ impl<A: NdFloat, D: Data<Elem = A>> BinaryClassification<A> for ArrayBase<D, Ix1
...
     }
 
-    fn assert_eq_iter<A: std::fmt::Debug + PartialEq>(a: impl Iterator<Item = A>, b: &[A]) {
-        assert_eq!(a.collect::<Vec<_>>(), b);
+    fn assert_eq_iter<A: std::fmt::Debug + PartialEq + AbsDiffEq + Clone>(
+        a: impl Iterator<Item = A>,
+        b: &[A],
+    ) {
+        let v: Vec<A> = a.collect();
+        assert_eq_iters(v.iter(), b);
```
bytesnake, line 516: could you not just pass ´a`, because it's already an iterator

paulkoerbitz: I did try that, however, `a` is an iterator producing (owned) `A`'s, not `&A`'s. Is there a way to still make this work?

paulkoerbitz:
Thanks for the review! 😀

src/metrics_classification.rs
bytesnake: with `Borrow` you can accept `A` and `&A`

```rust
fn assert_eq_iter<A: std::fmt::Debug + PartialEq + AbsDiffEq + Clone, B: Borrow<A>>(
      a: impl Iterator<Item = B>, 
      b: &[A],
) { 
    assert_eq_iters(a, b);                                                                                                                                                             
}  
fn assert_eq_iters<'a, A, B>( 
    a: impl IntoIterator<Item = B>, 
    b: impl IntoIterator<Item = &'a A>, 
) where
    A: 'a + std::fmt::Debug + PartialEq + AbsDiffEq,
    B: Borrow<A>,
 {
    let mut a_iter = a.into_iter();
    let mut b_iter = b.into_iter();
    loop {
        match (a_iter.next(), b_iter.next()) {
            (None, None) => break,
            (Some(a_item), Some(b_item)) => {
                abs_diff_eq!(a_item.borrow(), b_item);
            }
            _ => {
                panic!("assert_eq_iters: iterators had different lengths");
            }
        }
    }
}
```

paulkoerbitz: Awesome, thanks a lot!

PR #39, opened 2020-08-28
test DecisionTree fit method

fgadaleta:
DecisionTree fit doesn't seem to return the same result from a known Python equivalent (sklearn code below)

```Python
clf = tree.DecisionTreeClassifier()                                                                                                        
X = [ [1,2,3], [1, 2, 3.5], [1.2, 3, 4]  ]                                                                                                 
clf = clf.fit(X, [0,0,1])                                                                                                                  
clf.predict(X)                                                                                                                             
  array([0, 0, 1])
```
In this case, predicted labels should equal the ground truth. In the Rust implementation I cannot print that.


paulkoerbitz:
Thanks for contributing this. Indeed concerning. Would you be willing to investigate why this problem occurs?

PR #43, opened 2020-09-03
[WIP] Feature/linfa ensemble random forest

fgadaleta:
Implemented random forest baseline with tests and bench


bytesnake:
will look into this tomorrow :+1:

linfa-ensemble/benches/random_forest.rs
```diff
@@ -0,0 +1,50 @@
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use linfa_trees::DecisionTreeParams;
+use ndarray::{Array, Array1};
+use linfa_ensemble::{RandomForest, RandomForestParamsBuilder, MaxFeatures};
+
+
+fn random_forest_bench(c: &mut Criterion) {
+    // Load data
```
bytesnake, line 8: is this just random data or from some dataset?

linfa-ensemble/benches/random_forest.rs
```diff
@@ -0,0 +1,50 @@
...
+                        .min_samples_leaf(2 as u64)
+                        .build();
+    // Define parameters of random forest
+    let trees_set_sizes = vec![10, 100, 500, 1000];
+    // Benchmark training time 10 times for each training sample size
+    let mut group = c.benchmark_group("random_forest");
+    group.sample_size(10);
+
+    for ntrees in trees_set_sizes.iter() {
+        let xtrain = Array::from(data.clone()).into_shape((10, 5)).unwrap();
```
bytesnake, line 32: you could use the [array](https://docs.rs/ndarray/0.13.1/ndarray/macro.array.html) macro for the initialization and move it out of the loop. If you have ownership issues, then create a `ArrayView` with `.view()`.

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl RandomForest {
+    pub fn fit(
+        hyperparameters: RandomForestParams,
+        x: &ArrayBase<impl Data<Elem = f64>, Ix2>,
+        y: &ArrayBase<impl Data<Elem = u64>, Ix1>,
+    ) -> Self {
+
+    let n_estimators = hyperparameters.n_estimators;
```
bytesnake, line 23: there is a missing tab here :D

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,122 @@
...
+    let mut trees: Vec<DecisionTree> = Vec::with_capacity(n_estimators);
+    let single_tree_params = hyperparameters.tree_hyperparameters;
+
+    //TODO check bootstrap
+    let _bootstrap = hyperparameters.bootstrap;
```
bytesnake, line 28: this parameter controls whether to sample with replacement or without?

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    for _ in 0..n_estimators {
+        // Bagging here
+        let rnd_idx = Array::random((1, x.nrows()), Uniform::new(0, x.nrows())).into_raw_vec();
```
bytesnake, line 32: perhaps it would make sense to let the user choose a sample size smaller than the original one

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        let mut predictions: Array2<u64> = Array2::zeros((ntrees, x.nrows()));
+
+        for i in 0..ntrees {
+            let single_pred = self.trees[i].predict(&x);
+            dbg!("single pred: ", &single_pred);
+
+            // TODO can we make this more idiomatic rust
+            for j in 0..single_pred.len() {
+                predictions[[i, j]] = single_pred[j];
```
bytesnake, line 58: I don't know whether you can collect to `Array2`, but it's easier to write that with iterators.
```rust
self.trees.iter().map(|tree| tree.predict(&x)).collect::<Array2<_>>()
```
otherwise you can do something like [this](https://stackoverflow.com/questions/53698503/how-to-convert-a-collection-of-vecndarrayarray1-into-an-array2)

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+        }
+
+        let mut result: Vec<u64> = Vec::with_capacity(x.nrows());
+        for j in 0..predictions.ncols() {
+            let mut counter_stats: HashMap<u64, u64> = HashMap::new();
```
bytesnake, line 64: perhaps a more concise way is
```rust
let mut counter_stats = vec![HashMap::new(); predictions.ncols()];
self.trees.iter().map(|tree| {
   for (i, pred) in tree.predict(&x).into_iter().enumerate() {
      counter_stats[i].entry(pred).or_insert(0) += 1;
   }
});
```

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,122 @@
...
+            for i in 0..ntrees {
+                *counter_stats.entry(predictions[[i,j]]).or_insert(0) += 1;
+            }
+
+            let final_pred = counter_stats
```
bytesnake, line 69: you could shorten the collect phase
```rust
let final_pred = counter_stats.into_iter().map(|x| x.iter().max_by(..)..).collect::<Array1<_>>()
```

bytesnake (commented):
I mostly added some suggestions for places where you can add iterators instead of `for` loops. You should also make the algorithm generic over [Float](https://docs.rs/num-traits/0.2.12/num_traits/float/trait.Float.html) and labels.
In the end it would make sense to have a general `Bagging` implementation here, and add a more specialized version in `linfa-trees`, called `RandomForest`. 
Thanks for the work :+1:

linfa-ensemble/benches/random_forest.rs
fgadaleta: It's vectors for which I have a ground truth available. Once I know the single tree implementation is correct, we can replace this with other validation data.

fgadaleta:
> I mostly added some suggestions for places where you can add iterators instead of `for` loops. You should also make the algorithm generic over [Float](https://docs.rs/num-traits/0.2.12/num_traits/float/trait.Float.html) and labels.
> In the end it would make sense to have a general `Bagging` implementation here, and add a more specialized version in `linfa-trees`, called `RandomForest`.
> Thanks for the work +1

Wonderful comments. Will start fixing tomorrow ;)

linfa-ensemble/src/random_forest/algorithm.rs
fgadaleta: why?
the official implementation of RF with bagging uses the same sample size with replacement (to encourage diversity in the sample to fit the single tree).
If such sample is too small, there is a risk of fitting identical trees.

linfa-ensemble/src/random_forest/algorithm.rs
fgadaleta: yes. Just a placeholder now as functionality not yet implemented.

linfa-ensemble/src/random_forest/algorithm.rs
bytesnake: not necessary and rather an edge case, I was just thinking about a use-case where you have a very large dataset and want to speed up the inference

fgadaleta: Got it. Then I can define an `u64` to indicate the number of rows of data to fit trees against.

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -243,6 +265,53 @@ impl DecisionTree {
...
+            let s = queue.pop();
+            if let Some(node) = s {
+                // count only internal nodes (where features are)
+                if !node.leaf_node {
+                    // add feature index to list of used features
+                    fitted_features.push(node.feature_idx);
+                }
+
+                // get children and enque them
+                let lc = match node.left_child {
```
bytesnake, line 291: you could also do
```rust
for node in &[node.left_child, node.right_child] {
   if let Some(ref node) = node {
      if !visited.contains(node) {
          visited.insert(node);
          queue.push(node);
      } 
   }
}
```
or
```rust
&[node.left_child, node.right_child].iter().filter_map(|x| x)
   .filter(|node| !visited.contains(node))
   .for_each(|node| {
      visited.insert(node);
      queue.insert(node);
   })
```

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,134 @@
...
+                .unwrap();
+
+            result.push(*final_pred);
+        }
+        Array1::from(result)
+    }
+
+    /// Collect features from each tree in the forest and return hashmap(feature_idx: counts)
+    ///
+    pub fn feature_importances(&self) -> HashMap<usize, usize> {
```
bytesnake, line 76: :+1:

bytesnake (commented):
added some remarks on the feature importance stuff :)

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -243,6 +265,53 @@ impl DecisionTree {
...
+    ///
+    pub fn features(&self) -> Vec<usize> {
+        // features visited and counted
+        let mut visited: HashSet<TreeNode> = HashSet::new();
+        // queue of nodes yet to explore
+        let mut queue: Vec<TreeNode> = vec![];
+        // vector of feature indexes to return
+        let mut fitted_features: Vec<usize> = vec![];
+        // starting node
+        let root = self.root_node.clone();
```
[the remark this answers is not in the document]
fgadaleta: interesting. 
commit on your way :)

fgadaleta:
can we merge and close this?

bytesnake:
mh for me a `linfa-ensemble` with just random forest is a bit too empty, I think there are two ways forward:
 * implement ensemble learning for generic fittable algorithms, this depends on #45 and will take some time
 * or move the `RandomForest` implementation to `linfa-trees` and accept this PR

fgadaleta:
I am implemeting VotingClassifier in the same crate
All the family of XGB and extraRF could be under the same crate.
I think putting all that under linfa-trees can be misleading.

bytesnake:
well if they all are using decision trees, then I would put them under `linfa-trees` :sweat_smile: 

okay, if it's important for you I will accept, but at least improve the documentation a bit. First it is required for every subcrate to add a `README.md` describing its purpose and goals, then the link in the super crate is missing in the `README.md` too. Also add some comments to the `RandomForest` struct, you can follow these https://doc.rust-lang.org/stable/rust-by-example/meta/doc.html

fgadaleta:
By looking at the list of ensemble models indeed most of them are tree based. Hence it makes sense to refactor them there. I only found metaclassifier and votingclassifier not to be necessarily tree based (but just ensemble of generic weak learners)
Let me add the votingclassifier and documentation to the same PR and we move from there. Does this sound reasonable? :)

bytesnake:
yep sounds good :+1: I have zero experience in the field of ensemble learning, are you saying that most of them are tree based, because of properties of decision trees necessary for the ensemble learning process. Or that they normally, e.g. the most popular, are tree based?

fgadaleta:
Because they use trees as the single model to ensemble on.
With this said, there are exceptions. For instance a voting classifiers can be composed of heterogeneous models (not necessarily trees). Same goes for a Stacking classifier (can also be based on logistic regression) , bagging classifier (can be based on K-Nearest neighbors).
These models are still considered ensemble methods (just not tree-based).

paulkoerbitz (changes requested):
I've left a few comments, I haven't taken a very thorough look at this yet.

Unfortunately I think this PR brings up some design decisions that we need to make (e.g. `Predictor` / `Classifier` trait), which IMO need some discussion and there is some design work to do. I also think we need some more thorough test cases. Things like `unimplemented!` should not be merged.

linfa-predictor/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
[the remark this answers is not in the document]
bytesnake: Yep working on it here https://github.com/rust-ml/linfa/pull/45 :+1:

linfa-predictor/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+use ndarray::{Array1, ArrayBase, Data, Ix2};
+
+/// Trait every predictor should implement
+pub trait Predictor {
```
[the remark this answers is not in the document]
bytesnake: I think that `Predictor` is an accurate name, but should have a Target type parameter, specifying whether it predicts labels or continuous values. https://github.com/rust-ml/linfa/pull/45/files#diff-44e8606db87b39b048179b1f7af8d244R19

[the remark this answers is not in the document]
bytesnake: I would be okay with merging this as experimental, *when* the trait is included locally, until we figured out how to design the architecture linfa wide. A separate crate for all traits is nonsense, this should be in the base crate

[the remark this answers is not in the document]
bytesnake: the implementation can then be specialized at any point like 
```rust
impl<L: Label> Predict<Array2<f32>, Vec<L>> ... {
}
```
this accepts only labels, e.g. integers, strings and booleans and not more

linfa-predictor/src/lib.rs
```diff
@@ -0,0 +1,9 @@
...
+    /// predict class for each sample
+    fn predict(&self, x: &ArrayBase<impl Data<Elem = f64>, Ix2>) -> Array1<u64>;
```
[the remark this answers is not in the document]
bytesnake: same as above, but I'm now thinking whether you can implement a trait twice each with different return values? :thinking: Would be nice to have a 
```rust
impl<F: Float, D: Data> Predictor<D, Array1<F>> for .. {
}
impl<L: Label, D: Data> Predictor<D, Array1<L>> for .. {
}
```
for the same type and then decide on the required return type, which implementation is used. But lets move that to the PR :D

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -45,6 +49,24 @@ struct TreeNode {
...
+        stuff.push(self.leaf_node as u64);
+        stuff.hash(state);
+    }
+}
+
+impl Eq for TreeNode {}
+
+impl PartialEq for TreeNode {
+    fn eq(&self, other: &Self) -> bool {
+        self.feature_idx == other.feature_idx
```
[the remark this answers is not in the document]
bytesnake: mh to be absolutely correct, we need either to traverse the tree and compare nodes until the root node is reached, or introduce a unique index for each node. Note that multiple nodes in the tree can use the same feature index at different locations and would be equal in this definition. I would just introduce a counter in the `fit` function and use the value for `PartialEq` and `Hash` implementations.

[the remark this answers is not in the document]
fgadaleta: yeah I thought about that and was that close to add a node_id to the `Treenode`. But I actually had a doubt about it as I am not 100% sure the same feature_idx can be present multiple times (just on different split values) in the same tree.

linfa-predictor/src/lib.rs
```diff
@@ -0,0 +1,9 @@
...
+    /// predict probability of each possible class for each sample
+    fn predict_proba(&self, x: &ArrayBase<impl Data<Elem = f64>, Ix2>) -> Vec<Array1<f64>>;
```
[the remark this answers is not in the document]
fgadaleta: Noted. 
(I was keeping a similar naming convention to sklearn)

linfa-predictor/Cargo.toml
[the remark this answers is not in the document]
fgadaleta: I have implemented a `LinfaError` struct that is now returned from functions returning `Result` 
Specifically, decision trees should not implement `predict_probabilities()`. In the next commit I gently raise the error instead of the `unimplemented!()`

`Predictor` trait is currently in its own crate. I personally do not think that makes sense (it should be part of linfa, regardless). I have placed it as importable crate for convenience now, but needs to be moved

linfa-predictor/src/lib.rs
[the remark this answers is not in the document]
fgadaleta: This is interesting. Both a `Classifier` and `Regressor` are `Predictors` :) 
How about an `enum` ?

linfa-trees/src/decision_trees/algorithm.rs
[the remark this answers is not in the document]
paulkoerbitz: Re counter: not sure if this is a big deal (probably not), but many random forests enable parallelisation and the counter would need to be synced. An alternative could be to give key spaces to the different threads (e.g. have a globally assigned prefix for each thread and per-thread counter suffix).

Probably not an issue right now, just thinking out loud.

linfa-predictor/src/lib.rs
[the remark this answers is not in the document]
paulkoerbitz: OK, that would be fine with me as well.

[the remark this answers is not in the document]
paulkoerbitz: OK, I agree if the predictor is generic in its output then the name `Predictor` makes sense.

But I think predicting probabilities doesn't make sense for all predictors, so should be another trait.

linfa-predictor/Cargo.toml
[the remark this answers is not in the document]
mossbanay: > Specifically, decision trees should not implement `predict_probabilities()`
Do you mean that
1. This is actually genuinely not implemented and we should handle the case more gracefully
2. Making a prediction of class probabilities is not a sensible operation for decision trees?

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,191 @@
...
+use linfa_predictor::{LinfaError, Predictor, ProbabilisticPredictor};
+use linfa_trees::DecisionTree;
+use ndarray::Array;
+use ndarray::Axis;
+use ndarray::{Array1, ArrayBase, Data, Ix1, Ix2};
+use ndarray_rand::rand_distr::Uniform;
+use ndarray_rand::RandomExt;
+use std::collections::HashMap;
+
+/// A random forest is composed of independent decision trees performing a prediction and collecting each of them
```
mossbanay, line 11: nit: Maybe something to the effect of "whose predictions are aggregated" instead of "collecting"

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,191 @@
...
+    ///
+    /// * `x` - A 2D array of floating point elements
+    ///
+    ///
+    fn predict(
+        &self,
+        x: &ArrayBase<impl Data<Elem = f64>, Ix2>,
+    ) -> Result<Array1<u64>, LinfaError> {
+        let ntrees = self.hyperparameters.n_estimators;
+        assert!(ntrees > 0, "Run .fit() method first");
```
mossbanay, line 30: I think this has been discussed in the comments already but just a note here to replace with something like LinfaError instead of a panic.

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,191 @@
...
+                .unwrap();
+
+            result.push(*final_pred);
+        }
+        Ok(Array1::from(result))
+    }
+}
+
+impl ProbabilisticPredictor for RandomForest {
+    /// Return probability of predicted class for each sample, calculated as the rate of independent trees that
```
mossbanay, line 60: nit: s/rate/proportion

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,191 @@
...
+    pub fn fit(
+        hyperparameters: RandomForestParams,
+        x: &ArrayBase<impl Data<Elem = f64>, Ix2>,
+        y: &ArrayBase<impl Data<Elem = u64>, Ix1>,
+        max_n_rows: Option<usize>,
+    ) -> Self {
+        let n_estimators = hyperparameters.n_estimators;
+        let mut trees: Vec<DecisionTree> = Vec::with_capacity(n_estimators);
+        let single_tree_params = hyperparameters.tree_hyperparameters;
+        let max_n_rows = max_n_rows.unwrap_or_else(|| x.nrows());
```
mossbanay, line 108: Can't we just do `unwrap_or(x.nrows())` here?

fgadaleta:
One simply cannot predict probabilities with a decision tree (and should not) .
A decision tree is very unstable and can change dramatically for small perturbations of the input data.
If you want something that looks like probabilities you need to reduce the depth of the tree. And still all observations in the same node will have the same probability.
This is due to the fact that a decision tree generates probabilities from the number of samples of a class k on a leaf node, over the number of samples in that leaf.
When the tree changes so dramatically from small input data perturbations, all those "probabilities" change completely.
Only in ensemble methods (and random forest in particular) it makes sense to predict probabilities.

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        //TODO check bootstrap
```
mossbanay, line 110: Pretty important part of the algorithm is TODO here :smile: 
The bagging over samples looks good to me. When I was reading through the `sklearn` code I noticed a distinct absence of documentation internally. It could be nice to add a line or two here, something like,

Random forests work by bagging over samples (sampling from the original dataset with replacement) and taming a subset of features (sampling from the original features without replacement). Here we generate a random vector of sample indices `rnd_idx` and then selecting those rows from our original to get `xsample` and `ysample`.

Likewise here we'll need to generate something like a mask to decide whether to keep each feature. I'm actually rusty on how this is decided (i.e. whether it's just 50/50 whether to keep each feature or something else).

linfa-ensemble/src/random_forest/algorithm.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /// Collect features from each tree in the forest and return hashmap(feature_idx: counts)
+    ///
+    pub fn feature_importances(&self) -> HashMap<usize, usize> {
+        let mut counter: HashMap<usize, usize> = HashMap::new();
+        for st in &self.trees {
+            // features in the single tree
+            let st_feats = st.features();
+            for f in st_feats.iter() {
```
mossbanay, line 134: nit: As a user I would almost always want to see these ordered by index. Perhaps this should return a `Vec`?

Other note: there are alternative definitions of feature importance (e.g. Gini importance). I mention this not because we should implement that right away but because `usize` might not be the only sensible type. However that said, I would think that because these are pretty specific to random forest they can just be different methods with different return types.

linfa-ensemble/examples/random_forest.rs
```diff
@@ -0,0 +1 @@
+fn main() {}
```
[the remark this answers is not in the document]
mossbanay: Oops looks like this is in `linfa-ensemble/src/random_forest/examples/random_forest.rs`. In which case we should move that here?...

linfa-ensemble/src/random_forest/hyperparameters.rs
```diff
@@ -0,0 +1,80 @@
...
+#[derive(Clone, Copy)]
+pub struct RandomForestParams {
+    // number of decision trees to fit
+    pub n_estimators: usize,
+    // single tree hyperparameters
+    pub tree_hyperparameters: DecisionTreeParams,
+    // max number of features for the ensemble
+    pub max_features: u64,
+    // use all dataset or not
+    pub bootstrap: bool,
```
mossbanay, line 12: I think this doc is slightly misleading / also the flag is unused at the moment.

linfa-ensemble/src/random_forest/hyperparameters.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl RandomForestParamsBuilder {
+    pub fn new(
+        tree_hyperparameters: DecisionTreeParams,
+        n_estimators: usize,
+    ) -> RandomForestParamsBuilder {
+        RandomForestParamsBuilder {
+            tree_hyperparameters,
+            n_estimators,
+            max_features: Some(MaxFeatures::Auto),
```
mossbanay, line 37: If we initialise these to a (default) value then I think we don't need the `Option`. At the moment we sort of have two default values, one here and one in `build` when it does the `unwrap_or`.

linfa-ensemble/src/random_forest/hyperparameters.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn bootstrap(mut self, bootstrap: Option<bool>) -> Self {
+        self.bootstrap = bootstrap;
+        self
+    }
+
+    pub fn build(&self) -> RandomForestParams {
+        // TODO
+        let max_features = self.max_features.unwrap_or(MaxFeatures::Auto);
+
+        let max_features = match max_features {
```
mossbanay, line 66: TODO

linfa-ensemble/src/voting_classifier/algorithm.rs
```diff
@@ -0,0 +1,109 @@
+use linfa_predictor::{LinfaError, Predictor};
+use ndarray::{Array1, ArrayBase, Data, Ix2};
+use std::collections::HashMap;
+
+#[derive(Clone)]
+pub enum Voting {
+    Hard,
+    // Soft,
```
mossbanay, line 8: nit: If it's not implemented yet that's fine, we should document it in the roadmap instead of here where it's easy to forget about.

linfa-ensemble/src/voting_classifier/hyperparameters.rs
```diff
@@ -0,0 +1,3 @@
+// use linfa_trees::DecisionTreeParams;
+
+
```
mossbanay, line 3: I think the `Voting` value should live here with a builder.

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -45,6 +49,24 @@ struct TreeNode {
     prediction: u64,
 }
 
+impl Hash for TreeNode {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let mut stuff: Vec<u64> = vec![];
```
mossbanay, line 54: nit: Is there a better name than `stuff` here? I haven't done an `impl Hash` before don't know the convention.

mossbanay (commented):
I won't weigh in on `linfa-predictor` further than I think it's a good idea in principle and might take a few iterations to get to a point where we are happy with it.
The voting classifier and random forest look close but just need a bit of polish in some places. The biggest omission I can see is it doesn't look like we subset the features, which is a pretty core part of random forests.

mossbanay:
> One simply cannot predict probabilities with a decision tree (and should not) . A decision tree is very unstable and can change dramatically for small perturbations of the input data. If you want something that looks like probabilities you need to reduce the depth of the tree. And still all observations in the same node will have the same probability. This is due to the fact that a decision tree generates probabilities from the number of samples of a class k on a leaf node, over the number of samples in that leaf. When the tree changes so dramatically from small input data perturbations, all those "probabilities" change completely. Only in ensemble methods (and random forest in particular) it makes sense to predict probabilities.

I'm not saying using a decision tree is the _best option_ to predict probabilities but it's certainly valid. You could make the same argument about decision trees being unstable for regression right?

In some cases you have dense enough datasets where these aren't as big of an issue. Alternatively you can construct a shallow tree that tends to have more samples in the leaf nodes. Think about models used for inference rather than just prediction, knowing the distribution of the classes in each leaf node can be very useful.

linfa-ensemble/examples/random_forest.rs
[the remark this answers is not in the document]
fgadaleta: yes it's the same. 
I can move the test to example. But then there will be no test or a duplicate.

linfa-ensemble/README.md
```diff
@@ -20,6 +16,9 @@ There is an example in the `examples/` directory how to use random forest models
 $ cargo run --release --example random_forest
 ```
 
+The current benchmark performs random forest with 10, 100, 500, 1000 independent trees to predict labels and compare such results to ground truth.
+
+
```
mossbanay, line 21: Is it just `cargo bench` to run them?

fgadaleta:
yes

linfa-ensemble/src/random_forest/algorithm.rs
mossbanay: Sounds like the norm is to uniformly select `sqrt(p)` features for classification or `floor(p/3)` for regression (without replacement) [reference](https://en.wikipedia.org/wiki/Random_forest#From_bagging_to_random_forests).

fgadaleta: With bagging, one just samples with replacement and re-run the single trees with full features. 
What you mention with `sqrt(p)` is the maximum number of features one can force each single tree to "focus" on.
The standard bagging implementation of random forest however, uses all available features. Which is why I left the  `_bootstrap` as TODO.

fgadaleta: cleaning up doc indeed it can be misleading

mossbanay: If you don't take a subset of the features then it's not a random forest, it's just bagged decision trees.

mossbanay:
Hey, how's this going? Any luck with the sub-sampling of features? Let me know if you need a hand I'd be happy to help. This is pretty close to getting over the line and into the crate :smile:

bytesnake:
this PR seems to stale, continue work in #60

PR #46, opened 2020-09-20
Elastic Net

paulkoerbitz:
First implementation of elastic net regression, which encompasses Lasso and Ridge regression. Uses coordinate descent optimisation algorithm.

One question is if we should really make this its own sub-package or put this in linfa-linear. This package does not have significant extra dependencies (in fact, it has fewer dependencies than linfa-linear).

This PR is WIP, I would like to add the following points:
- [x] wrappers for Lasso and Ridge Regression
- [x] dual gap convergence check
- [x] Add readme or move into linfa_linear
- [x] more thorough tests
- [x] Generalize over Float type


linfa-linear/src/elastic.rs
```diff
@@ -0,0 +1,572 @@
...
+            intercept,
+            parameters,
+            duality_gap,
+            n_steps,
+        })
+    }
+
+    /// Compute the intercept as the mean of `y` and center `y` if an intercept should
+    /// be used, use `0.0` as intercept and leave `y` unchanged otherwise.
+    fn compute_intercept<'a>(&self, y: &'a Array1<F>) -> (F, Cow<'a, Array1<F>>) {
```
bytesnake, line 128: I think there is a `ndarray::CowArrray` for this usecase

linfa-linear/src/elastic.rs
```diff
@@ -0,0 +1,572 @@
...
+    let n_samples = F::from(x.shape()[0]).unwrap();
+    let n_features = x.shape()[1];
+    // the parameters of the model
+    let mut w = Array1::<F>::zeros(n_features);
+    // the residuals: `y - X*w` (since w=0, this is just `y` for now),
+    // the residuals are updated during the algorithm as the parameters change
+    let mut r = y.clone();
+    let mut n_steps = 0u32;
+    let norm_cols_x = x.map_axis(Axis(0), |col| {
+        col.fold(F::zero(), |sum_sq, &x| sum_sq + x * x)
```
bytesnake, line 331:
```rust
col.mapv_into(|x| x*x).sum()
```
should also work here

bytesnake (approved):
looking good to me, two ideas:
 * add some comments how the duality gap works
 * add a test where one column correlates with the target perfectly and everything else is gaussian noise

paulkoerbitz:
Thanks for the comments @bytesnake! I'll address them shortly, also added more tests now to gain confidence. Otherwise I think this is OK to submit (once comments are addressed).

linfa-linear/src/elastic.rs
paulkoerbitz: I think this requires ownership/mutability (for `mapv_inplace`) of the underlying data, but I'm working on a shared reference here. I would also not want to copy the column previous to manipulating it to avoid the extra allocation.

paulkoerbitz: But I guess `col.dot(&col)` should work.

linfa-linear/src/elastic.rs
paulkoerbitz: Thanks for pointing this out. I've tried changing the code to using CowArray, but in this case I found it inconvenient. 

CowArray wraps either a `ArrayBase<ViewRepr, ...>` or an `ArrayBase<OwnedRepr, ...>`, which is semantically the same as Cow, but slightly different: For the view case, I have to take a slice and then convert that into the owned array. Then the subsequent code takes a shared ref to an owned array (this could of course be changed as well....). In the end I found this inconvenient (in this instance) and left this like it is.

linfa-linear/src/elastic.rs
bytesnake: yeah that's true, you could do `x.mapv(|y| y*y).sum_axis(Axis(1))`, but this would copy the `ndarray` as well

linfa-linear/src/elastic.rs
bytesnake: wait when the coordinate descent routine takes the intercept as reference to an owned one, the `Cow` wrapper is always converted to an owned object. Why can't you then not just return ` -> (F, Array1<F>)` and clone the intercept for the second case implicitly?

bytesnake:
Merged in #76

PR #51, opened 2020-10-06
Gaussian Naive Bayes (GaussianNB)

VasanthakumarV:
Port of [GaussianNB](https://scikit-learn.org/stable/modules/generated/sklearn.naive_bayes.GaussianNB.html)

### DIFF
--- a/linfa-bayes/Cargo.toml
+++ b/linfa-bayes/Cargo.toml
@@ -18,3 +18,7 @@ num-traits = "0.1.32"
 
 [dev-dependencies]
 approx = "0.3"
+linfa = { path = ".." }
+csv = "1.1"
+ndarray-csv = "0.4"
+flate2 = "1.0"
--- a/linfa-bayes/Cargo.toml
+++ b/linfa-bayes/Cargo.toml
@@ -13,11 +13,9 @@ categories = ["algorithms", "mathematics", "science"]
 [dependencies]
 ndarray = { version = "0.13" , features = ["blas", "approx"]}
 ndarray-stats = "0.3"
-linfa = { version = "0.2.1", path=".." }
+linfa = { version = "0.2.1", path = ".." }
 
 [dev-dependencies]
 approx = "0.3"
 linfa = { path = ".." }
-csv = "1.1"
-ndarray-csv = "0.4"
-flate2 = "1.0"
+linfa-datasets = { version = "0.2.0", path = "../datasets", features = ["winequality"] }
--- a/linfa-bayes/Cargo.toml
+++ b/linfa-bayes/Cargo.toml
@@ -15,3 +15,6 @@ ndarray = { version = "0.13" , features = ["blas"]}
 ndarray-linalg = { version = "0.12", features = ["openblas"] }
 ndarray-stats = "0.3"
 num-traits = "0.1.32"
+
+[dev-dependencies]
+approx = "0.3"
--- a/linfa-bayes/Cargo.toml
+++ b/linfa-bayes/Cargo.toml
@@ -11,10 +11,9 @@ keywords = ["factorization", "machine-learning", "linfa", "unsupervised"]
 categories = ["algorithms", "mathematics", "science"]
 
 [dependencies]
-ndarray = { version = "0.13" , features = ["blas"]}
-ndarray-linalg = { version = "0.12", features = ["openblas"] }
+ndarray = { version = "0.13" , features = ["blas", "approx"]}
 ndarray-stats = "0.3"
-num-traits = "0.1.32"
+linfa = { version = "0.2.1", path=".." }
 
 [dev-dependencies]
 approx = "0.3"
--- a/linfa-bayes/Cargo.toml
+++ b/linfa-bayes/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "linfa-bayes"
+version = "0.1.0"
+authors = ["VasanthakumarV <vasanth260m12@gmail.com>"]
+description = "Collection of Naive Bayes Algorithms"
+edition = "2018"
+license = "MIT/Apache-2.0"
+repository = "https://github.com/rust-ml/linfa"
+readme = "README.md"
+keywords = ["factorization", "machine-learning", "linfa", "unsupervised"]
+categories = ["algorithms", "mathematics", "science"]
+
+[dependencies]
+ndarray = { version = "0.13" , features = ["blas"]}
+ndarray-linalg = { version = "0.12", features = ["openblas"] }
+ndarray-stats = "0.3"
+num-traits = "0.1.32"
--- a/linfa-bayes/README.md
+++ b/linfa-bayes/README.md
@@ -1,0 +1,27 @@
+# Naive Bayes
+
+`linfa-bayes` aims to provide pure Rust implementations of Naive Bayes algorithms. 
+
+## The Big Picture
+
+`linfa-bayes` is a crate in the [`linfa`](https://crates.io/crates/linfa) ecosystem, an effort to create a toolkit for classical Machine Learning implemented in pure Rust, akin to Python's `scikit-learn`.
+
+## Current state
+
+`linfa-bayes` currently provides an implementation of the following methods: 
+
+- Gaussian Naive Bayes (GaussianNB)
+
+## Examples
+
+There is an usage example in the `examples/` directory. To run, use:
+
+```bash
+$ cargo run --example winequality 
+```
+
+## License
+Dual-licensed to be compatible with the Rust project.
+
+Licensed under the Apache License, Version 2.0 <http://www.apache.org/licenses/LICENSE-2.0> or the MIT license <http://opensource.org/licenses/MIT>, at your option. This file may not be copied, modified, or distributed except according to those terms.
+
--- a/linfa-bayes/examples/winequality.rs
+++ b/linfa-bayes/examples/winequality.rs
@@ -9,8 +9,10 @@ use flate2::read::GzDecoder;
 use ndarray::{Array1, Array2, Axis};
 use ndarray_csv::Array2Reader;
 
-use linfa::metrics::IntoConfusionMatrix;
-use linfa_bayes::{FittedGaussianNb, GaussianNb};
+use linfa::dataset::Dataset;
+use linfa::metrics::ToConfusionMatrix;
+use linfa::traits::{Fit, Predict};
+use linfa_bayes::GaussianNbParams;
 
 fn main() -> Result<(), Box<dyn Error>> {
     // Read in the wine-quality dataset from dataset path
@@ -18,70 +20,35 @@ fn main() -> Result<(), Box<dyn Error>> {
     let dataset = read_array("../datasets/winequality-red.csv.gz")?;
 
     // Training data goes along the first axis (rows)
-    let npoints = dataset.len_of(Axis(0));
-    // Take 90% of the dataset as training data, and the remaining 10% as validation data
-    // `floor` is used here to explicitely down-round the number
-    let ntrain = (npoints as f32 * 0.9).floor() as usize;
-
     // The first 11 columns are features used in training and the last columns are targets
     let (data, targets) = dataset.view().split_at(Axis(1), 11);
     // group targets below 6.5 as bad wine and above as good wine
-    let targets = targets.into_iter().map(|x| *x > 6.5f64).collect::<Vec<_>>();
-    // split into training and validation data
-    let (train_data, train_targets) = (data.slice(s!(0..ntrain, ..)), &targets[0..ntrain]);
-    let (valid_data, valid_targets) = (data.slice(s!(ntrain.., ..)), &targets[ntrain..]);
-
-    let train_data = train_data.to_owned();
-    let valid_data = valid_data.to_owned();
-
-    fn tag_classes(x: bool) -> f64 {
-        if x {
-            1.
+    fn tag_classes(x: &f64) -> usize {
+        if *x > 6.5 {
+            1
         } else {
-            0.
+            0
         }
     };
+    let targets: Array1<_> = targets.into_iter().map(tag_classes).collect();
 
-    // Map targets from boolean to float
-    let train_targets: Array1<f64> = train_targets
-        .into_iter()
-        .cloned()
-        .map(tag_classes)
-        .collect();
-    let valid_targets: Array1<f64> = valid_targets
-        .into_iter()
-        .cloned()
-        .map(tag_classes)
-        .collect();
+    // Take 90% of the dataset as training data, and the remaining 10% as validation data
+    // `floor` is used here to explicitely down-round the number
+    let npoints = dataset.len_of(Axis(0));
 
-    // Initialize the model
-    let mut gnb: GaussianNb<f64> = GaussianNb::default();
-    let mut model = FittedGaussianNb::unfitted();
+    let ntrain = (npoints as f32 * 0.9).floor() as usize;
+    let (train_data, train_targets) = (data.slice(s!(0..ntrain, ..)), targets.slice(s!(0..ntrain)));
+    let (valid_data, valid_targets) = (data.slice(s!(ntrain.., ..)), targets.slice(s!(ntrain..)));
 
-    // Train the model using the incremental learning api
-    // `fit` method is also available for training using all data
-    for (x, y) in train_data
-        .axis_chunks_iter(Axis(0), 120)
-        .zip(train_targets.axis_chunks_iter(Axis(0), 120))
-    {
-        // The `&array![0., 1.]` represents the unique classes available
-        // in the dataset, this argument is required when training incrementally
-        model = gnb.fit_with(model, &x, &y, &array![0., 1.])?;
-    }
+    let train = Dataset::new(train_data, train_targets);
 
-    // Calculation predictions on the validation set
-    let prediction = model.predict(&valid_data)?;
+    // Train the model
+    let model = GaussianNbParams::params().fit(&train)?;
 
-    // We convert the predictions and the validation target as string for
-    // compatibility with the confusion matrix api
-    let prediction_str: Vec<_> = prediction.to_vec().iter().map(|x| x.to_string()).collect();
-    let valid_targets_str: Vec<_> = valid_targets
-        .to_vec()
-        .iter()
-        .map(|x| x.to_string())
-        .collect();
+    let pred = model.predict(valid_data)?;
 
-    let cm = prediction_str.into_confusion_matrix(&valid_targets_str);
+    // Construct confusion matrix
+    let cm = pred.confusion_matrix(valid_targets);
 
     // classes    | 0          | 1
     // 0          | 131        | 7
--- a/linfa-bayes/examples/winequality.rs
+++ b/linfa-bayes/examples/winequality.rs
@@ -1,6 +1,3 @@
-#[macro_use]
-extern crate ndarray;
-
 use std::error::Error;
 use std::fs::File;
 
@@ -22,39 +19,35 @@ fn main() -> Result<(), Box<dyn Error>> {
     // Training data goes along the first axis (rows)
     // The first 11 columns are features used in training and the last columns are targets
     let (data, targets) = dataset.view().split_at(Axis(1), 11);
+    let targets = targets.into_iter().collect::<Array1<_>>();
     // group targets below 6.5 as bad wine and above as good wine
-    fn tag_classes(x: &f64) -> usize {
-        if *x > 6.5 {
+    fn tag_classes(x: &&f64) -> usize {
+        if **x > 6.5 {
             1
         } else {
             0
         }
     };
-    let targets: Array1<_> = targets.into_iter().map(tag_classes).collect();
+
+    let dataset = Dataset::new(data, targets).map_targets(tag_classes);
 
     // Take 90% of the dataset as training data, and the remaining 10% as validation data
-    // `floor` is used here to explicitely down-round the number
-    let npoints = dataset.len_of(Axis(0));
-
-    let ntrain = (npoints as f32 * 0.9).floor() as usize;
-    let (train_data, train_targets) = (data.slice(s!(0..ntrain, ..)), targets.slice(s!(0..ntrain)));
-    let (valid_data, valid_targets) = (data.slice(s!(ntrain.., ..)), targets.slice(s!(ntrain..)));
-
-    let train = Dataset::new(train_data, train_targets);
+    let (train, valid) = dataset.split_with_ratio(0.9);
 
     // Train the model
     let model = GaussianNbParams::params().fit(&train)?;
 
-    let pred = model.predict(valid_data)?;
+    //let pred = model.predict(valid_data)?;
+    let pred = model.predict(valid.records)?;
 
     // Construct confusion matrix
-    let cm = pred.confusion_matrix(valid_targets);
+    let cm = pred.confusion_matrix(&valid);
 
-    // classes    | 0          | 1
-    // 0          | 131        | 7
-    // 1          | 12         | 10
+    // classes    | 1          | 0
+    // 1          | 10         | 12
+    // 0          | 7          | 130
     //
-    // accuracy 0.88125, MCC 0.45128104
+    // accuracy 0.8805031, MCC 0.45080978
     println!("{:?}", cm);
     println!("accuracy {}, MCC {}", cm.accuracy(), cm.mcc());
 
--- a/linfa-bayes/examples/winequality.rs
+++ b/linfa-bayes/examples/winequality.rs
@@ -11,22 +11,6 @@ use ndarray_csv::Array2Reader;
 
 use linfa::metrics::IntoConfusionMatrix;
 use linfa_bayes::GaussianNb;
-//use linfa_kernel::Kernel;
-//use linfa_svm::{SVClassify, SolverParams};
-
-/// Extract a gziped CSV file and return as dataset
-fn read_array(path: &str) -> Result<Array2<f64>, Box<dyn Error>> {
-    // unzip file
-    let file = GzDecoder::new(File::open(path)?);
-    // create a CSV reader with headers and `;` as delimiter
-    let mut reader = ReaderBuilder::new()
-        .has_headers(true)
-        .delimiter(b';')
-        .from_reader(file);
-    // extract ndarray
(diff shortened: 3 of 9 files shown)


bytesnake:
will look into this in the weekend :)

linfa-bayes/src/error.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use ndarray::ShapeError;
+use ndarray_stats::errors::MinMaxError;
+
+pub type Result<T> = std::result::Result<T, BayesError>;
+
+#[derive(Debug)]
+pub enum BayesError {
+    /// Error when accessing an untrained model
+    UntrainedModel(String),
```
bytesnake, line 11: it should not be possible to run in a situation where an error is thrown because a model is not initialized. Let me example why I think so:
 * each model should have two structs, one where the training hyperparameters are stored, and one which contains a fitted model
 * the function `Parameters::fit` should construct a fitted model from a set of hyperparameters
 * the function `Parameters::fit_with` should take a `Option<Model>` and return a new model with an updated state

A training loop would look then like this:
```
let hparams = GNBayes::params();
create_data().iter()
    .fold(None, |model, data| Some(hparams.fit_with(model, data)?))
```
or
```
let hparams = GNBayes::params();
let mut model = None;
for data in create_data() {
     model = Some(hparams.fit_with(model, data)?);
}
```
what do you think?

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -0,0 +1,544 @@
...
+    /// # }
+    /// ```
+    pub fn partial_fit(
+        &mut self,
+        x: &ArrayView2<A>,
+        y: &ArrayView1<A>,
+        classes: &Array1<A>,
+    ) -> Result<()> {
+        // If the ratio of the variance between dimensions is too small, it will cause
+        // numberical errors. We address this by artificially boosting the variance
```
bytesnake, line 152: typo

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -0,0 +1,544 @@
...
+    fn joint_log_likelihood(&self, x: &Array2<A>) -> Array2<A> {
+        let mut joint_log_likelihood = Array2::zeros((x.nrows(), self.classes.len()));
+
+        for i in 0..self.classes.len() {
+            let jointi = self.priors.get(i).unwrap().ln();
+
+            let mut nij = self
+                .sigma
+                .row(i)
+                .mapv(|x| A::from(2.).unwrap() * A::from(std::f64::consts::PI).unwrap() * x)
```
bytesnake, line 414: `2.0f64 * std::f64::consts::PI`

bytesnake (commented):
For me it makes more sense to structure it into `ModelParameter` -> `FittedModel`, than `ProbablyFittedModel` -> `FittedModel`. Added a comment how the API would look like, what do you think?

linfa-bayes/src/gaussian_nb.rs
VasanthakumarV: Fixed

linfa-bayes/src/gaussian_nb.rs
VasanthakumarV: Fixed

linfa-bayes/examples/winequality.rs
```diff
@@ -0,0 +1,109 @@
...
+    let valid_targets: Array1<f64> = valid_targets
+        .into_iter()
+        .cloned()
+        .map(tag_classes)
+        .collect();
+
+    // Initialize the model
+    let mut model: GaussianNb<f64> = GaussianNb::new();
+
+    // Trian the model using the incremental learning api
```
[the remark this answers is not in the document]
VasanthakumarV: Fixed

linfa-bayes/README.md
```diff
@@ -0,0 +1,27 @@
+# Naive Bayes
```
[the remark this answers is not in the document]
VasanthakumarV: Added linfa-bayes to the overview table at the front

linfa-bayes/examples/winequality.rs
```diff
@@ -0,0 +1,109 @@
...
+    let targets = targets.into_iter().map(|x| *x > 6.5f64).collect::<Vec<_>>();
+    // split into training and validation data
+    let (train_data, train_targets) = (data.slice(s!(0..ntrain, ..)), &targets[0..ntrain]);
+    let (valid_data, valid_targets) = (data.slice(s!(ntrain.., ..)), &targets[ntrain..]);
+
+    let train_data = train_data.to_owned();
+    let valid_data = valid_data.to_owned();
+
+    fn tag_classes(x: bool) -> f64 {
+        if x {
```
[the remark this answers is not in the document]
VasanthakumarV: The model expects a number for the target/labels and not String. I was hoping we will create a label encoder for the conversions.

linfa-bayes/src/error.rs
VasanthakumarV: I like the functional style, might require involved refactoring, will start working on it tomorrow.

VasanthakumarV: I took a stab at it today, please look at [example](https://github.com/VasanthakumarV/linfa/blob/feature/bayes/linfa-bayes/examples/winequality.rs#L57) to see how the incremental learning api changed.

I had to duplicate one parameter ([`priors`](https://github.com/VasanthakumarV/linfa/blob/feature/bayes/linfa-bayes/src/gaussian_nb.rs#L91)) between the two structs, which I don't feel good about.

Also do let me know what naming convention I should follow, for the structs and methods.

I think having one struct encompassing all methods (fit, fit_with and predict) will give the users the best experience.
Are there benefits to separating the predict function from training?

bytesnake: > I took a stab at it today, please look at [example](https://github.com/VasanthakumarV/linfa/blob/feature/bayes/linfa-bayes/examples/winequality.rs#L57) to see how the incremental learning api changed.
> 
 * new should be named `default_params` and `unfitted` to give an indication what they are doing
 * the `partial_fit` could consume a model and produce one (for me its simpler to reason about the behaviour)

Another point are labels. At the moment you're reserving an additional field to `partial_fit` for them, but I think they should be implicitly stored in the targets. One way to solve this, is to have a trait for targets and a corresponding function for labels. For a Vec/Array1 this performs the `unique` function. A special struct `TargetsWithLabels` would wrap a Vec/Array1 and its corresponding labels. In this case the function would return the provided labels (and perhaps performs a check operation to see that they are complete) I have experimented with this [here](https://github.com/rust-ml/linfa/pull/45/files#diff-1b4495edaff386d67486c019daec29fdR17). After we merged the PR and figured out how the API should look like, I would like to update the syntax.

> I had to duplicate one parameter ([`priors`](https://github.com/VasanthakumarV/linfa/blob/feature/bayes/linfa-bayes/src/gaussian_nb.rs#L91)) between the two structs, which I don't feel good about.
> 
well they are different, the first one is a prior prior, and the second one an estimated prior :D 

> Also do let me know what naming convention I should follow, for the structs and methods.
> 
mh don't know, I think `partial_fit` should be called `update` or `fit_with` when the prior model is optional

> I think having one struct encompassing all methods (fit, fit_with and predict) will give the users the best experience.
> Are there benefits to separating the predict function from training?

yes that you can't produce untrained models. If you have `Model::params -> Params`, `Params::fit -> Model`, `Params::fit_with -> Model` only one struct is visible to the user, the separation is only an implementation detail
```rust
let model = GaussianNB::params()
   .with_priors(..)
   .var_smoothing(0.5)
   .fit(&x, &y, &labels);
```

VasanthakumarV: Thank you for the detailed comments,
- `new` methods in the [example](https://github.com/VasanthakumarV/linfa/blob/feature/bayes/linfa-bayes/examples/winequality.rs#L57) is now `default` and `unfitted`
- `partial_fit` is now `fit_with`
- Do we come back to fix the labels/classes argument in the `fit_with` method after merging the traits PR?
- I still had to use the `UntrainedModel` error because `unfitted` method returns a model object, using Option to avoid this isn't very elegant, because we have to return a model object wrapped in Option and eventually the user has to unwrap it

bytesnake: > Thank you for the detailed comments,
> 
>     * `new` methods in the [example](https://github.com/VasanthakumarV/linfa/blob/feature/bayes/linfa-bayes/examples/winequality.rs#L57) is now `default` and `unfitted`
> 
>     * `partial_fit` is now `fit_with`
> 
>     * `fit_with` consumes a model and produces one
> 
okay :+1: 

>     * Do we come back to fix the labels/classes argument in the `fit_with` method after merging the traits PR?
> 
yes definitely, I'm playing around atm how the new traits and SVM classifier are interacting, if you want to speed the process up, do a review of the PR. 

>     * I still had to use the `UntrainedModel` error because `unfitted` method returns a model object, using Option to avoid this isn't very elegant, because we have to return a model object wrapped in Option and eventually the user has to unwrap it
yeah that is true, either its an external error or an internal one .. once we have proper error handling in the top level, a `UntrainedModel` error is due

I will look into the remaining parts tomorrow again, and when everything is fine, merge the PR

bytesnake:
Could you take a look at https://github.com/rust-ml/linfa/blob/fdf241127fe9058ae98c67a603a2bbb8eccc561c/CONTRIBUTE.md, I'm currently writing a small how-to-contribute document for the upcoming interface changes. If you find something unclear, please let me know so I can update the document.

linfa-bayes/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+edition = "2018"
+license = "MIT/Apache-2.0"
+repository = "https://github.com/rust-ml/linfa"
+readme = "README.md"
+keywords = ["factorization", "machine-learning", "linfa", "unsupervised"]
+categories = ["algorithms", "mathematics", "science"]
+
+[dependencies]
+ndarray = { version = "0.13" , features = ["blas"]}
+ndarray-linalg = { version = "0.12", features = ["openblas"] }
```
bytesnake, line 15: you don't need the `ndarray-linalg` dependency, because you don't call any subroutine in your code

linfa-bayes/examples/winequality.rs
```diff
@@ -0,0 +1,109 @@
...
+fn main() -> Result<(), Box<dyn Error>> {
+    // Read in the wine-quality dataset from dataset path
+    // The `.csv` data is two dimensional: Axis(0) denotes y-axis (rows), Axis(1) denotes x-axis (columns)
+    let dataset = read_array("../datasets/winequality-red.csv.gz")?;
+
+    // Training data goes along the first axis (rows)
+    let npoints = dataset.len_of(Axis(0));
+    // Take 90% of the dataset as training data, and the remaining 10% as validation data
+    // `floor` is used here to explicitely down-round the number
+    let ntrain = (npoints as f32 * 0.9).floor() as usize;
```
bytesnake, line 24: can you use the new dataset interface for the splitting here, here is an example: https://github.com/rust-ml/linfa/blob/master/linfa-svm/examples/winequality.rs

linfa-bayes/src/error.rs
```diff
@@ -0,0 +1,46 @@
+use std::fmt;
+
+use ndarray::ShapeError;
+use ndarray_stats::errors::MinMaxError;
+
+pub type Result<T> = std::result::Result<T, BayesError>;
+
+#[derive(Debug)]
+pub enum BayesError {
```
bytesnake, line 9: there is now a global error type for algorithms, which you can extend https://github.com/rust-ml/linfa/blob/master/src/error.rs

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -0,0 +1,546 @@
...
+
+use ndarray::{s, Array1, Array2, ArrayView1, ArrayView2, Axis};
+use ndarray_stats::QuantileExt;
+
+use crate::error::{BayesError, Result};
+use crate::Float;
+
+/// Gaussian Naive Bayes (GaussianNB)
+#[derive(Debug)]
+pub struct GaussianNb<A> {
```
bytesnake, line 15: this should be `GaussianNbParams`

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -0,0 +1,546 @@
...
+    ///
+    /// let mut clf = GaussianNb::new();
+    /// let fitted_clf = clf.fit(&x, &y)?;
+    /// let pred = fitted_clf.predict(&x)?;
+    ///
+    /// assert_eq!(pred, y);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn fit(&mut self, x: &Array2<A>, y: &Array1<A>) -> Result<FittedGaussianNb<A>> {
```
bytesnake, line 87: with merging #55 a new fitting API is introduced, this needs an update here

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -0,0 +1,546 @@
...
+            .enumerate()
+            .for_each(|(i, &r)| xsubset.row_mut(i).assign(&x.slice(s![r, ..])));
+
+        Ok(xsubset)
+    }
+}
+
+/// Fitted GaussianNB for predicting classes
+#[derive(Debug)]
+pub struct FittedGaussianNb<A> {
```
bytesnake, line 364: this should just be `GaussianNb`, which was previously the hyperparmeter struct

linfa-bayes/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+mod error;
+mod gaussian_nb;
+
+use ndarray::NdFloat;
+use ndarray_linalg::Lapack;
+use num_traits::FromPrimitive;
+
+pub use gaussian_nb::{FittedGaussianNb, GaussianNb};
+
+pub trait Float:
```
bytesnake, line 10: there is now a `linfa::Float` trait

bytesnake (commented):
I updated the review with regards to merging #55 
If you have any questions how to implement the new traits, please let me know!

VasanthakumarV:
@bytesnake Thank you for the detailed comments, will try to make the changes by next week.

linfa-bayes/Cargo.toml
VasanthakumarV: Removed `ndarray-linalg`

linfa-bayes/examples/winequality.rs
VasanthakumarV: Some errors like `BayesError::Priors` are very specific to the algorithm, do you still want to move them over to the root?

linfa-bayes/src/error.rs
VasanthakumarV: Some errors like `BayesError::Priors` are very specific to the algorithm, do you still want to move them over to the root?

linfa-bayes/src/gaussian_nb.rs
VasanthakumarV: Changed

linfa-bayes/src/gaussian_nb.rs
VasanthakumarV: Changed

linfa-bayes/src/gaussian_nb.rs
VasanthakumarV: Changed

linfa-bayes/src/lib.rs
VasanthakumarV: Changed

linfa-bayes/examples/winequality.rs
VasanthakumarV: The `split_with_ratio` method for `Dataset` returns a slice for the target attribute, now should I,

- convert slice to ndarray and proceed? or
- make the algorithm work with a slice target?

The second option will make linfa more ergonomic to use, but I want to know your thoughts

VasanthakumarV:
@bytesnake For incrementally learning the algorithm needs the definite set of classes supplied by the user, currently I supply them as an argument in the `fit_with` method, this prevents me from implementing the `linfa::traits::IncrementalFit` trait.

linfa-bayes/examples/winequality.rs
bytesnake: yeah the second would be more useful

linfa-bayes/src/error.rs
bytesnake: no I changed my mind on this. We should use an error library which offers extensible errors. For the time being just use your own error type here

bytesnake:
I think you can use the [fn labels(&self) -> Vec<T::Elem>](https://github.com/rust-ml/linfa/blob/master/src/dataset/mod.rs#L89) returning a vector of the unique classes. For this you have to restrict your targets like this:
```rust
impl<F: Float, R: Records<Elem = F>, T: Labels<Elem = usize>> IncrementalFit<R, T> for GaussianNbParams<F> {
}
```
in the future the trait could also be extended to accept any countable type:
```rust
impl<F: Float, L: Label, R: Records<Elem = F>, T: Label<Elem = L>> IncrementalFit<R, T> for GaussianNbParams<F, L> {
}
```

bytesnake:
oh and you could estimate the priors from the samples as well, there is the [Dataset::frequencies_with_mask](https://github.com/rust-ml/linfa/blob/master/src/dataset/impl_dataset.rs#L189) function. It was introduced for decision trees, where you have to mask the data at each split point. You could either pass a `true` slice or introduce a function just called `frequencies`. This also takes the sample weights in consideration.

linfa-bayes/examples/winequality.rs
VasanthakumarV: The example now used `split_with_ratio` method

VasanthakumarV:
@bytesnake I have started using the `Labels` trait for the dataset targets, and now I am using `labels` method for accessing unique entries. 

But the issue with incremental learning in naive-bayes is that we need to know the complete list of labels while training on batches of data, the first batch of data that is fed to the model will almost never have encountered all the labels, that is the reason why I have an extra argument in the function signature.

Also I see [`Iter`](https://github.com/rust-ml/linfa/blob/master/src/dataset/iter.rs) data structure, do you think we will be able to create chunks of Datasets using this?, this will be very useful for the incremental learning api.

VasanthakumarV:
> oh and you could estimate the priors from the samples as well, there is the [Dataset::frequencies_with_mask](https://github.com/rust-ml/linfa/blob/master/src/dataset/impl_dataset.rs#L189) function. It was introduced for decision trees, where you have to mask the data at each split point. You could either pass a `true` slice or introduce a function just called `frequencies`. This also takes the sample weights in consideration.

@bytesnake Currently I calculate priors along with mean and variance for each class, which requires masked indexing for each class.
So more than [Dataset::frequencies_with_mask](https://github.com/rust-ml/linfa/blob/master/src/dataset/impl_dataset.rs#L189), a convenience function for masked indexing or regular indexing will be useful, my [implementation](https://github.com/rust-ml/linfa/blob/19b07470710272fc5ad4a380db20a0032265d9da/linfa-bayes/src/gaussian_nb.rs#L345) is a little ugly, but can work on a more generic one if you think it will be useful.

bytesnake:
@VasanthakumarV I think that an incremental model should collect its label incrementally as well. Would it be possible to form the union of previous labels with new encounter ones and update the model correspondingly. Otherwise you could also pass the labels as a hyper-parameter and return an error when a different label is encountered.

bytesnake:
> So more than [Dataset::frequencies_with_mask](https://github.com/rust-ml/linfa/blob/master/src/dataset/impl_dataset.rs#L189), a convenience function for masked indexing or regular indexing will be useful, my [implementation](https://github.com/rust-ml/linfa/blob/19b07470710272fc5ad4a380db20a0032265d9da/linfa-bayes/src/gaussian_nb.rs#L345) is a little ugly, but can work on a more generic one if you think it will be useful.

the API around dataset is very sparse right now, any update is welcome and a function which can select samples based on the class should definitely be in there somewhere :+1: It would be even more useful if you could somehow produce a view on the sub-selected samples, but I don't think that is possible right now.
I want to start a PR, renaming `Dataset` -> `DatasetBase` and implementing `Dataset` with `Records = Array2<F>` and `Targets = Array1<L>` to simplify the implementation details.

VasanthakumarV:
> @VasanthakumarV I think that an incremental model should collect its label incrementally as well. Would it be possible to form the union of previous labels with new encounter ones and update the model correspondingly. Otherwise you could also pass the labels as a hyper-parameter and return an error when a different label is encountered.

@bytesnake I went with the first approach from a usability standpoint. 
To achieve this I replaced ndarray matrix with a hashtable, this greatly reduced the complexity of the training step, introduced a little bit of complexity to the prediction step.

To implement the incremental learning API, I had to change IncrementalFit API - [link](https://github.com/rust-ml/linfa/pull/51/files#diff-172747fc43e38d33de743b2d5f564ad6440570670af7aa912235399448a55c7cL34). 
[Example](https://github.com/rust-ml/linfa/blob/5c04b813d25540aaa826d8c5d5dcbb2021625ce9/linfa-bayes/src/gaussian_nb.rs#L121) - exhibits what it takes to train a model incrementally, I hope this matches the expectation we had for this.

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -97,32 +83,29 @@ where
...
+        model = self.fit_with(model, dataset)?;
+
+        Ok(model.unwrap())
     }
 }
 
-impl<A: Float> GaussianNbParams<A> {
+impl<A, L> IncrementalFit<'_, ArrayView2<'_, A>, L, BayesError> for GaussianNbParams
+where
+    A: Float + PartialEq + PartialOrd,
```
bytesnake, line 97: you can just bound this in `linfa::Float`, it wasn't done to this point because no other algorithm used partial equality/ordering

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -134,173 +117,105 @@ impl<A: Float> GaussianNbParams<A> {
...
-            // sure it is correct
-            if let Some(ref priors) = model.priors {
-                if priors.len() != nclasses {
-                    return Err(BayesError::Priors(format!(
-                        "The number of priors: ({}), does not match the number of classes: ({})",
-                        priors.len(),
-                        nclasses
-                    )));
-                }
+            * *x.var_axis(Axis(0), A::from(0.).unwrap()).max().unwrap();
```
bytesnake, line 154: use `A::zero()` here

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -134,173 +117,105 @@ impl<A: Float> GaussianNbParams<A> {
...
-                    )));
-                }
-
-                if priors.iter().any(|x| x < &A::from(0.).unwrap()) {
-                    return Err(BayesError::Priors(
-                        "Class priors cannot have negative values".to_string(),
-                    ));
-                }
-            }
+        if let Some(temp) = model_in {
```
bytesnake, line 156: why not using 
```rust
let model = match model_in {
...
};
```
here

linfa-bayes/src/gaussian_nb.rs
```diff
@@ -488,73 +413,50 @@ mod tests {
...
         let y = array![1, 1, 1, 2, 2, 2];
-        let classes = vec![1, 2];
 
         let clf = GaussianNbParams::params();
-        let mut model = GaussianNb::unfitted();
+
+        let mut model = None;
 
         for (x, y) in x
             .axis_chunks_iter(Axis(0), 2)
```
bytesnake, line 422: in the future there should be a `Dataset::chunks_iter` for this :D

bytesnake (commented):
thanks for all the work, will merge after adressing these small nitpicks :+1:

bytesnake:
ready to merge?

VasanthakumarV:
> ready to merge?
I really don't like the [changes](https://github.com/rust-ml/linfa/pull/51/files#diff-172747fc43e38d33de743b2d5f564ad6440570670af7aa912235399448a55c7cR34) I made to IncrementalFIt trait, especially hardcoding Result on the output. I might take another day to make it better.

VasanthakumarV:
@bytesnake What do you think about the following trait for IncrementalFit?, two associated types to capture the asymmetry.

```rust
pub trait IncrementalFit<'a, R: Records, T: Targets> {
    type Input: 'a;
    type Output: 'a;

    fn fit_with(&self, model: Self::Input, dataset: &'a Dataset<R, T>) -> Self::Output;
}

// Implementation
impl<A, L> IncrementalFit<'_, ArrayView2<'_, A>, L> for GaussianNbParams
where
    A: Float,
    L: Labels<Elem = usize>,
{
    type Input = Option<GaussianNb<A>>;
    type Output = Result<Option<GaussianNb<A>>>;

    fn fit_with(&self, model_in: Self::Input, dataset: &Dataset<ArrayView2<A>, L>) -> Self::Output {
        ...
    }
    ...
}

```

The problem here is the flexibility, no rule is enforced, but the user has the freedom to either return an Option or an Option wrapped in a result should there be a need.

bytesnake:
didn't see this yesterday, this definitely gives you more flexibility. I had another thing in my mind, which I pushed here https://github.com/VasanthakumarV/linfa/pull/1
Shouldn't be the output `type Output = Option<Result<GaussianNb<A>>>;`?

VasanthakumarV:
> Shouldn't be the output `type Output = Option<Result<GaussianNb<A>>>;`?

it is `Result<Option<GaussianNb<A>>>` so that the error if any can be bubbled up and the underlying model can be fed back in for the next iteration

bytesnake:
I'm back from christmas and after the break I think your idea is good, especially in combination with `try_fold`. We should revise the decision for `Self::Object` once we have `Result<T, anyhow::Error>`

VasanthakumarV:
@bytesnake do you want to rename the associated type's names from `Input`, `Output` to `ObjectIn`, `ObjectOut`, or something else?

bytesnake:
@VasanthakumarV no `Input` and `Output` is fine and transparent, otherwise this could be `InputObj`, `ObjectIn` or other combinations :sweat_smile:

VasanthakumarV:
> @VasanthakumarV no `Input` and `Output` is fine and transparent, otherwise this could be `InputObj`, `ObjectIn` or other combinations 😅

I made them `ObjectIn` and `ObjectOut` for consistency with `Fit` trait. Let me know if you want any changes made.

bytesnake (approved):

PR #54, opened 2020-11-01
WIP OPTICS clustering

xd009642:
Still working out how best to test it but it seems mostly there. Wiki for details https://en.wikipedia.org/wiki/OPTICS_algorithm there's also a paper I used as reference at the start I'll see if I can dig that out if anyone wants. Also I think I may have to change some of the neighbour stuff to be more correct. I'll figure it out 👀

### DIFF
--- a/linfa-ica/src/fast_ica.rs
+++ b/linfa-ica/src/fast_ica.rs
@@ -272,7 +272,7 @@ impl GFunc {
 
     #[allow(clippy::manual_range_contains)]
     fn logcosh<A: Float>(x: &Array2<A>, alpha: f64) -> Result<(Array2<A>, Array1<A>)> {
-        if !(alpha >= 1. && alpha <= 2.) {
+        if alpha < 1. || alpha > 2. {
             return Err(FastIcaError::InvalidValue(format!(
                 "alpha must be between 1 and 2 inclusive, got {}",
                 alpha
--- a/linfa-svm/src/solver_smo.rs
+++ b/linfa-svm/src/solver_smo.rs
@@ -99,7 +99,7 @@ impl<'a, A: Float, K: 'a + Permutable<'a, A>> SolverState<'a, A, K> {
             .collect::<Vec<_>>();
 
         // initialize full active set
-        let active_set = (0..alpha.len()).map(|i| i).collect::<Vec<_>>();
+        let active_set = (0..alpha.len()).collect::<Vec<_>>();
 
         // initialize gradient
         let mut gradient = p.clone();
--- a/src/error.rs
+++ b/src/error.rs
@@ -22,7 +22,7 @@ impl fmt::Display for Error {
             Error::Parameters(msg) => write!(f, "Parameter: {}", msg),
             Error::NdShape(msg) => write!(f, "NdArray shape: {}", msg),
             Error::Priors(msg) => write!(f, "Priors: {}", msg),
-            Error::NotConverged(msg) => write!(f, "Not converged: {}", msg)
+            Error::NotConverged(msg) => write!(f, "Not converged: {}", msg),
         }
     }
 }


bytesnake:
Thank you! I will look into this at the end of the week.
Could you take a look at https://github.com/rust-ml/linfa/blob/fdf241127fe9058ae98c67a603a2bbb8eccc561c/CONTRIBUTE.md, I'm currently writing a small how-to-contribute document for the upcoming interface changes. If you find something unclear, please let me know so I can update the document. The interface is a bit different, to the original K-Means implementation because of the new traits. If you find some drawbacks in ergonomic, please let me know as well.

xd009642:
@bytesnake sure, I had a brief look and it looks fine but I'll see if I have any more concrete feedback after using it. I'll try and rebase and start updating it some point this week so you might want to save giving it too strong a look until that's done :sweat_smile:. Plus I still need to figure out adequate tests :thinking: 

An interesting aside of OPTICS, is you can use the distance metrics it outputs to calculate an equivalent to the output of some clustering algorithms like DBSCAN. It's naturally slower to do OPTICS then from that get the DBSCAN clusters but it might be a nice demonstration of the correctness of both algorithms :thinking:

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,236 @@
...
+)]
+/// OPTICS (Ordering Points To Identify Clustering Structure) is a clustering algorithm that
+/// doesn't explicitly cluster the data but instead creates an "augmented ordering" of the dataset
+/// representing it's density-based clustering structure. This ordering contains information which
+/// is equivalent to the density-based clusterings and can then be used for automatic and
+/// interactive cluster analysis.
+pub struct Optics;
+
+impl Optics {
+    pub fn predict<'a>(
```
bytesnake, line 116: `linfa::Transformer` can be used here

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,236 @@
...
+            let neighbors = find_neighbors(
+                &obs.1.observation,
+                observations,
+                hyperparameters.tolerance(),
+            );
+            let n = obs.1;
+            n.set_core_distance(hyperparameters.minimum_points(), &neighbors);
+            result.orderings.push(n.sample());
+            if n.c_distance.is_some() {
+                let mut seeds = Vec::new();
```
bytesnake, line 149: can you add a small comment what seeds are doing?

bytesnake (commented):
can you update to the new traits #55? The implementation looks straighforward, but I will read through some paper first. Also some tests are missing

xd009642:
Yeah so far just done the rebase, gonna try to get onto moving to new traits etc this weekend or next week. Just have some tarpaulin stuff I need to do first :sweat_smile:

xd009642:
I've moved over to the transformer trait. One comment though I think potentially it should return a result type. Especially since existing code does tend to unwrap the l2_dist result and we all know how tricky floating point numbers can be. Either a result or decide for each algorithm if a floating point op fails what the appropriate value to replace it with so we can change them to `unwrap_or`.

xd009642:
Also just noticed when I ran cargo test on the root there's dependencies in tests for the main linfa crate that aren't in the Cargo.toml? I'm patching locally can open another PR, commit here or let someone else do it

bytesnake:
> I've moved over to the transformer trait. One comment though I think potentially it should return a result type. Especially since existing code does tend to unwrap the l2_dist result and we all know how tricky floating point numbers can be. Either a result or decide for each algorithm if a floating point op fails what the appropriate value to replace it with so we can change them to `unwrap_or`.

I think `Transformer` will most of the time return a `Result<_>` type, but I think it does not make much difference to have the type implicitly in the trait

bytesnake:
> Also just noticed when I ran cargo test on the root there's dependencies in tests for the main linfa crate that aren't in the Cargo.toml? I'm patching locally can open another PR, commit here or let someone else do it

wait are some dependencies missing in `dev-dependencies`? I haven't encountered any errors in travis and locally

xd009642:
Fair enough, I did some similar traits in a work project and explicitly did the Result and then had an infallible type which was automatically generated when the error type was `std::convert::Infallible`

xd009642:
> > Also just noticed when I ran cargo test on the root there's dependencies in tests for the main linfa crate that aren't in the Cargo.toml? I'm patching locally can open another PR, commit here or let someone else do it
> 
> wait are some dependencies missing in `dev-dependencies`? I haven't encountered any errors in travis and locally

hm looking at the diff it looks like it was removed between now and master, but I don't recall changing that Cargo.toml. What's the branch the traits are in so I can diff against that?

xd009642:
Also I think the only thing I need to do bar docs, tidy-up and any review comments is to actually do the tests properly. I'll look for an existing implementation work out the optics result then encode that in the test for similarity

bytesnake:
> > > Also just noticed when I ran cargo test on the root there's dependencies in tests for the main linfa crate that aren't in the Cargo.toml? I'm patching locally can open another PR, commit here or let someone else do it
> > 
> > 
> > wait are some dependencies missing in `dev-dependencies`? I haven't encountered any errors in travis and locally
> 
> hm looking at the diff it looks like it was removed between now and master, but I don't recall changing that Cargo.toml. What's the branch the traits are in so I can diff against that?

can you post the error you're encountering? The `traits` branch is already merged into master, perhaps the error goes away, when you rebase your PR to the newest commit?

xd009642:
Oh maybe I'll rebase and see if that fixes it, probably just a merge fuck up on my end

xd009642:
Okay something has gone seriously wrong I tried a rebase --onto and it worked with no complaints. test passed, push... Saw this PR got closed and then realised it deleted all of my work...

shit shit shit.

gimme a moment

PR #58, opened 2020-11-11
Feature/optics

xd009642:
@bytesnake after a rebase nightmare I'm back


linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,247 @@
...
+        let mut points = observations
+            .axis_iter(Axis(0))
+            .enumerate()
+            .map(|(i, x)| Neighbor::new(i, x))
+            .collect::<Vec<_>>();
+
+        let mut processed = HashSet::new();
+        while processed.len() != points.len() {
+            let obs = points
+                .iter_mut()
```
bytesnake, line 143: is there a reason why you iterating in each step through all points, i.e. can it happen that point is "un-processed"? Otherwise you can just do:
```rust
let mut i = 0;
loop {
    if points[i].processed {
        i += 1;
        continue
    }
    ...
    i += 1;
}
```
also no need for `iter_mut` here

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,247 @@
...
+    fn transform(&self, observations: &'a ArrayBase<D, Ix2>) -> OpticsResult<'a, F> {
+        let mut result = OpticsResult { orderings: vec![] };
+
+        let mut points = observations
+            .axis_iter(Axis(0))
+            .enumerate()
+            .map(|(i, x)| Neighbor::new(i, x))
+            .collect::<Vec<_>>();
+
+        let mut processed = HashSet::new();
```
bytesnake, line 140: do you need a separate `processed` field in Neighbor and a `processed` hashset?

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,247 @@
...
+                .unwrap();
+            if processed.contains(&obs.0) {
+                // We've processed this point so can move on
+                continue;
+            }
+            processed.insert(obs.0);
+            let neighbors = find_neighbors(&obs.1.observation, observations, self.tolerance());
+            let n = obs.1;
+            n.set_core_distance(self.minimum_points(), &neighbors);
+            result.orderings.push(n.sample());
```
bytesnake, line 155: does this actually works, how can you ensure that `n` outlives `n.samples`?

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,247 @@
...
+            if n.c_distance.is_some() {
+                // Here we get a list of "density reachable" samples that haven't been processed
+                // and sort them by reachability so we can process the closest ones first.
+                let mut seeds = Vec::new();
+                get_seeds(n.clone(), &neighbors, &mut points, &processed, &mut seeds);
+                while !seeds.is_empty() {
+                    seeds.sort_unstable_by(|a, b| b.cmp(a));
```
bytesnake, line 162: `seeds` are the indices of points which have been added to a core the first time. It was unclear to me why ordering the indices also orders the distance in the core. Could you just add a sentence, which says that `get_neighbors` returns ordered element

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,247 @@
...
+    res
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use ndarray::Array2;
+
+    #[test]
+    fn simple_optics() {
```
bytesnake, line 239: you can add tests for the following categories:
 * test single functions, i.e. check that `find_neighbors` and `get_seeds` are working as intended
 * do qualitative testing, i.e. construct simple example where the ordering is clear, probably with only a handful of points 
 * do quantitative testing, i.e. generate random data and prove effectiveness in separating them

bytesnake (commented):
okay I finally got some time to read the OPTIC's paper and did a second round of comments. The name 'core_distance' was really confusing to me, because it could also be the distance to an assigned distribution in sense of KMeans :/ 
Performance-wise you could add a faster neighborhood algorithm, for example [hnsw](https://github.com/rust-cv/hnsw). I did something similar for sparse kernel matrices, can be found [here](https://github.com/rust-ml/linfa/blob/master/linfa-kernel/src/sparse.rs). I also don't think you need the second for loop, blowing up your runtime performance. 
The second half of the paper deals with the optimal cluster points, that is still missing, but I really like the way to illustrate a N-dimensional distribution in a simple reachability plot.
Sorry for the trouble with the rebasing :eyes:

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,247 @@
...
+///
+/// OPTICS cluster analysis can be used to derive clusters equivalent to the output of other
+/// clustering algorithms such as DBSCAN. However, due to it's more complicated neighborhood
+/// queries it typically has a higher computational cost that other more specific algorithms.
+pub struct Optics;
+
+#[derive(Clone, Debug)]
+// This is a struct as in future may want to implement methods on it to get certain metrics from
+// the optics cluster distances.
+pub struct OpticsResult<'a, F: Float> {
```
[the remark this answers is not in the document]
xd009642: True, maybe `OpticsAnalysis` instead

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,247 @@
...
+            observation,
+            processed: false,
+            c_distance: None,
+            r_distance: None,
+        }
+    }
+
+    /// Set the core distance given the minimum points in a cluster and the points neighbors
+    fn set_core_distance(&mut self, min_pts: usize, neighbors: &[Neighbor<'a, F>]) {
+        self.c_distance = neighbors
```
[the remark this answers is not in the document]
xd009642: But if the point isn't a core point the distance should be `None`, I figured this was a reasonable way to avoid misuse of the function without sacrificing performance

linfa-clustering/src/optics/algorithm.rs
xd009642: I figured the lifetime ensured that but I might be wrong

linfa-clustering/src/optics/algorithm.rs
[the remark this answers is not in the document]
bytesnake: ah I see, perhaps you can use the `slice::get` syntax:
```rust
self.c_distance = neighbors.get(min_pts-1)
    .and_then(|x| self.observation.l2_dist(x))
    .ok();
```
A possible error originating from the `l2_dist` is converted into `None` with the last line

linfa-clustering/src/optics/algorithm.rs
bytesnake: if it compiles then it should work :eyes:

xd009642:
After being way too overloaded with work/life I'm now back to try and implement the feedback :eyes:

xd009642:
Right HNSW done, I need to go through the processed comments and remind myself how OPTICS works as it's been a while and do tests. I did have a go on implementing on `ArrayView` and found myself in a hellish nightmare of lifetime errors, I'll try again once I've addressed the other stuff and am sure it's working as hopefully the code will simplify a bit - or I'll remember why I did what I did all those months ago.

bytesnake:
> After being way too overloaded with work/life I'm now back to try and implement the feedback eyes

yes know this feeling, for me it helps to realize that work will finish at some point in the future and therefore not pressure myself too much

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,291 @@
...
+use ndarray::{ArrayBase, ArrayView, ArrayView1, Axis, Data, Ix1, Ix2};
+use ndarray_stats::DeviationExt;
+#[cfg(feature = "serde")]
+use serde_crate::{Deserialize, Serialize};
+use space::MetricPoint;
+use std::cmp::{max, Ordering};
+use std::collections::BTreeSet;
+
+/// Implementation of euclidean distance for ndarray
+struct Euclidean<'a, F>(ArrayView1<'a, F>);
```
bytesnake, line 15: there is also the plan to unify these in `linfa-nn` https://github.com/rust-ml/linfa/issues/119#issuecomment-841839071

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+#[derive(Clone)]
+struct Neighbor<'a, F: Float> {
+    /// Index of the observation in the dataset
+    index: usize,
+    /// The observation
+    observation: ArrayView<'a, F, Ix1>,
+    /// The core distance
+    c_distance: Option<FloatOrd<f64>>,
```
bytesnake, line 70: can you use standardized naming, i.e. `c_distance` or `core_distance`

bytesnake: ah this is in conflict with the member right? I think `c_distance` should be o.k. with a comment above

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,291 @@
...
+        loop {
+            if index == points.len() {
+                break;
+            } else if processed.contains(&index) {
+                index += 1;
+                continue;
+            }
+            let mut expected = index + 1;
+            let mut points_index = 0;
+            for index in processed.range((index + 1)..) {
```
bytesnake, line 159: this find the next lowest index (starting at `index+1`) which is not yet processed right? If yes can you add a comment for that

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,291 @@
...
+            n.set_core_distance(self.minimum_points(), &neighbors);
+            result.orderings.push(n.sample());
+            if n.c_distance.is_some() {
+                // Here we get a list of "density reachable" samples that haven't been processed
+                // and sort them by reachability so we can process the closest ones first.
+                let mut seeds = Vec::new();
+                get_seeds(n.clone(), &neighbors, &mut points, &processed, &mut seeds);
+                while !seeds.is_empty() {
+                    seeds.sort_unstable_by(|a, b| b.cmp(a));
+                    let (i, min_point) = seeds
```
bytesnake, line 183: if the seeds are sorted why query for the minimal value and not just take the first item?

bytesnake: ah the above sorts indices and this the actual values

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,291 @@
...
+                        .iter()
+                        .enumerate()
+                        .min_by(|(_, a), (_, b)| points[**a].cmp(&points[**b]))
+                        .unwrap();
+                    let n = &mut points[*min_point];
+                    seeds.remove(i);
```
bytesnake, line 189: this may be slow perhaps also use `BTreeSet` here? (this would give you sorted indices for free)

algorithms/linfa-clustering/src/optics/hyperparameters.rs
```diff
@@ -0,0 +1,85 @@
...
+        self.tolerance
+    }
+
+    /// Minimum number of a points in a neighborhood around a point for it to
+    /// not be considered noise
+    pub fn minimum_points(&self) -> usize {
+        self.min_points
+    }
+
+    fn build(tolerance: f64, min_points: usize) -> Self {
```
bytesnake, line 72: we've moved in various places to a lazy variant of this:
 * add a function `validate -> Result<..>` which checks all hyperparameter for validity
 * return a `Result` inside `Transformer` or `Fit` and call `self.params.validate()?` at the beginning
 * no need to call build anymore and have two variants of hyperparameters

bytesnake (commented):
:+1:

algorithms/linfa-clustering/Cargo.toml
```diff
@@ -28,17 +28,20 @@ default-features = false
 features = ["std", "derive"]
 
 [dependencies]
+hnsw = "0.6"
```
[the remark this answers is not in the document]
xd009642: I just partially copied the hnsw from linfa-kernel while I figured out the API, I'll update later

algorithms/linfa-clustering/src/optics/algorithm.rs
xd009642: Once they're unified I can change over, otherwise if this is done and merged first I guess it will be unified when the great unification happens

algorithms/linfa-clustering/src/optics/algorithm.rs
xd009642: Will do :+1:

algorithms/linfa-clustering/Cargo.toml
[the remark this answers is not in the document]
bytesnake: I still have to update the dependency in linfa-kernel

algorithms/linfa-clustering/src/optics/algorithm.rs
xd009642: Given that the seeds are only points reachable from the current processed point, I'd expect more often than not the seed list won't be large enough to see any efficiency gains from `BTreeSet` versus the insertion cost. 

I would expect some better performance to move creation of seeds to outside the outer loop and clearing it at the end of each iteration though so I'll do that

algorithms/linfa-clustering/src/optics/algorithm.rs
bytesnake: sure!

xd009642:
What's the minimum supported rust version policy? Updating HNSW seems to have caused some failures due to const generics being stabilised

linfa-clustering/src/optics/algorithm.rs
xd009642: I need to do some testing for `get_seeds` but I've added some consistency tests to make sure certain properties are kept with the returned values. I'll add some more tests tomorrow if I have time but it looks like it's coming along :grin:

linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+}
+
+impl Optics {
+    pub fn params(min_points: usize) -> OpticsHyperParamsBuilder {
+        OpticsHyperParams::new(min_points)
+    }
+}
+
+impl<'a, F: Float, D: Data<Elem = F>> Transformer<&'a ArrayBase<D, Ix2>, OpticsResult<'a, F>>
```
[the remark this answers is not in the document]
xd009642: So I'm struggling a bit with the `ArrayView` implementation because `OpticsAnalysis` doesn't own the data and it complains that `find_neighbors` borrows the data. I think I might need to seriously refactor what I've done to get round the generic/lifetime hole I've dug myself into :grimacing: 

```rust
error[E0515]: cannot return value referencing function parameter `observations`
   --> algorithms/linfa-clustering/src/optics/algorithm.rs:209:9
    |
143 |         let mut points = observations
    |                          ------------ `observations` is borrowed here
...
209 |         Ok(result)
    |         ^^^^^^^^^^ returns a value referencing data owned by the current function

error[E0515]: cannot return value referencing function parameter `observations`
   --> algorithms/linfa-clustering/src/optics/algorithm.rs:209:9
    |
174 |                 &observations,
    |                 ------------- `observations` is borrowed here
...
209 |         Ok(result)
    |         ^^^^^^^^^^ returns a value referencing data owned by the current function

error[E0515]: cannot return value referencing function parameter `observations`
   --> algorithms/linfa-clustering/src/optics/algorithm.rs:209:9
    |
194 |                     let neighbors = find_neighbors(&n.observation, &observations, self.tolerance());
    |                                                                    ------------- `observations` is borrowed here
...
209 |         Ok(result)
    |         ^^^^^^^^^^ returns a value referencing data owned by the current function
```

YuhanLiin:
If you use `linfa-nn` for `find_neighbours` you can eliminate the HNSW dependency while retaining good perf. That'll solve the MSRV issue.

bytesnake:
> What's the minimum supported rust version policy? Updating HNSW seems to have caused some failures due to const generics being stabilised.
> 
> Edit: CI currently has it to 1.42, min const generics needs 1.49

a change of the MSRV should result in a bump in the minor version. Otherwise I think we should adopt const generic support, because it's a central feature in numerical optimizations

bytesnake:
> If you use `linfa-nn` for `find_neighbours` you can eliminate the HNSW dependency while retaining good perf. That'll solve the MSRV issue.

the author answered one review comment: `Once they're unified I can change over, otherwise if this is done and merged first I guess it will be unified when the great unification happens`

linfa-clustering/src/optics/algorithm.rs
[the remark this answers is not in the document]
bytesnake: did you solve this? Otherwise I can take a look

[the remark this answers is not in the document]
xd009642: Nope didn't manage to solve it, wanted to try and do it without having to change too much of the existing code. But I might have to bite the bullet and just make some significant changes :eyes: 

I did consider most of the lifetime stuff could be removed if the `OpticsAnalysis` didn't have the row views ordered with distances but stored the index. But dunno how forward compatible/nice-to-use that will be

.github/workflows/checking.yml
```diff
@@ -33,7 +33,7 @@ jobs:
           args: --workspace
 
       - name: Run cargo check (with serde)
-        if: ${{ matrix.toolchain != '1.42.0' }} # The following args don't work on older versions of Cargo
+        if: ${{ matrix.toolchain != '1.49.0' }} # The following args don't work on older versions of Cargo
```
YuhanLiin, line 36: If we're bumping MSRV we probably don't need this condition, which was only put in because the old versions of Cargo couldn't handle the `cargo check` command with serde.

algorithms/linfa-clustering/src/optics/hyperparameters.rs
```diff
@@ -0,0 +1,69 @@
...
+    pub fn with_tolerance(mut self, tolerance: f64) -> Self {
+        self.tolerance = tolerance;
+        self
+    }
+
+    /// Two points are considered neighbors if the euclidean distance between
+    /// them is below the tolerance
+    pub fn tolerance(&self) -> f64 {
+        self.tolerance
+    }
```
YuhanLiin, line 47: Can you rename these methods to `tolerance` and `get_tolerance` respectively? `with_tolerance` seems like a weird name for a builder method.

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,498 @@
...
+)]
+/// OPTICS (Ordering Points To Identify Clustering Structure) is a clustering algorithm that
+/// doesn't explicitly cluster the data but instead creates an "augmented ordering" of the dataset
+/// representing it's density-based clustering structure. This ordering contains information which
+/// is equivalent to the density-based clusterings and can then be used for automatic and
+/// interactive cluster analysis.
+///
+/// OPTICS cluster analysis can be used to derive clusters equivalent to the output of other
+/// clustering algorithms such as DBSCAN. However, due to it's more complicated neighborhood
+/// queries it typically has a higher computational cost that other more specific algorithms.
```
YuhanLiin, line 40: change "that" to "than"

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,498 @@
...
+pub struct Optics;
+
+#[derive(Clone, Debug)]
+// This is a struct as in future may want to implement methods on it to get certain metrics from
```
YuhanLiin, line 44: "as in future may" => "as in the future we may"

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,498 @@
+use crate::optics::errors::Result;
+use crate::optics::hyperparameters::OpticsHyperParams;
+use float_ord::FloatOrd;
```
YuhanLiin, line 3: `FloatOrd` ordering semantics don't handle NaN at all, since NaN is listed as the largest and smallest value in the docs. Assuming that you don't actually want NaNs in your input at all, the `noisy_float` crate is the better choice. The type `noisy_float::NoisyFloat` implements `Ord` and also panics when you attempt to construct it with NaN in debug mode, so it does a better job of upholding the implicit assumption of no NaNs.

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,498 @@
...
+}
+
+impl Optics {
+    pub fn params(min_points: usize) -> OpticsHyperParams {
+        OpticsHyperParams::new(min_points)
+    }
+}
+
+impl<'a, F: Float, D: Data<Elem = F>>
+    Transformer<&'a ArrayBase<D, Ix2>, Result<OpticsAnalysis<'a, F>>> for OpticsHyperParams
```
YuhanLiin, line 137: Did you manage to fix the lifetime issue here? If not then I suggest you remove the `ArrayViews` from `OpticsAnalysis` and just use the indices, since that's what DBSCAN does. Also the trait here should be `Predict`, which is what's used by DBSCAN right now.

algorithms/linfa-clustering/src/optics/hyperparameters.rs
xd009642: Sure, I just did it to match the new lazy builder pattern using the `gaussian_mixture` module as reference, so someone might want to do a PR to change all the `with_*` methods there.

algorithms/linfa-clustering/src/optics/algorithm.rs
xd009642: OPTICS is more "meta" than DBSCAN, for example DBSCAN predicts clusters, whereas OPTICS generates some metadata for each data point that could be used to implement something like DBSCAN, it doesn't actually do any clustering

YuhanLiin: Now that I think about it you're right. OPTICs returns distance info on every point so it makes more sense to use `Transformer`.

algorithms/linfa-clustering/src/optics/hyperparameters.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    /// Two points are considered neighbors if the euclidean distance between
+    /// them is below the tolerance
+    pub fn get_tolerance(&self) -> f64 {
+        self.tolerance
+    }
+
+    /// Minimum number of a points in a neighborhood around a point for it to
+    /// not be considered noise
+    pub fn minimum_points(&self) -> usize {
```
YuhanLiin, line 51: Can you also change this to get_minimum_points for consistency?

xd009642:
Annoyingly CI failed due to failing to download one of the dependencies from crates.io. I've been getting this on a few different projects on github actions in the past week so wonder if they've got connection issues :thinking: if someone wants to hit rerun on that job as I seem unable to trigger a rerun

xd009642:
Although actually I'm going to make the members of the `OpticsAnalysis` private and put methods on it to access things :eyes:

xd009642:
@bytesnake @YuhanLiin I've moved `Sample` and `OpticsAnalysis` to `src/optics/analysis.rs` as there may in future be algorithms implemented on them. Also made the fields non-public to avoid a maintenance headache of avoiding breaking changes and added accessors and updated documentation.

bytesnake:
you have to raise the MSRV for testing to 1.51 as well

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,492 @@
...
+    fn set_core_distance<F: Float>(
+        &mut self,
+        min_pts: usize,
+        neighbors: &[Neighbor],
+        dataset: ArrayView<F, Ix2>,
+    ) {
+        let observation = dataset.row(self.index);
+        self.c_distance = neighbors
+            .get(min_pts - 1)
+            .map(|x| n64(observation.l2_dist(&dataset.row(x.index)).unwrap()));
```
bytesnake, line 74: gets a tiny bit simpler if you move`index` outside the map and `n64(..)` into a separate map

xd009642:
Ah I just did the resolve merge conflicts via the github UI and it didn't show that file as conflicted, pushing now

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,492 @@
...
+        .iter()
+        .enumerate()
+        .find(|(_, x)| x.distance > eps)
+        .map(|(i, _)| i);
+    if let Some(i) = out_of_bounds {
+        // once shrink_to is stablised can switch to that
+        neighbours.resize_with(i, space::Neighbor::invalid);
+    }
+
+    let mut result = vec![];
```
bytesnake, line 269:
```rust
neighbours.into_iter().map(|x| x.index).map(|idx| {
        let observation = observations.row(idx);
        let distance = candidate.l2_dist(&observation).unwrap();
        Neighbor {
            index: idx,
            r_distance: Some(n64(distance)),
            c_distance: None,
        }
}).collect()

**Review by bytesnake — APPROVED** [review]

File: algorithms/linfa-clustering/src/optics/algorithm.rs (hunk 23 above)
Comment (xd009642) [reply algorithms/linfa-clustering/src/optics/algorithm.rs:74]: You can't move index out of the map, because it comes from `|x|`.  I guess I could do:

```
        self.c_distance = neighbors
            .get(min_pts - 1)
            .map(|x| x.index)
            .map(|x| observation.l2_dist(&dataset.row(x)).unwrap())
            .map(|x| n64(x));
```

**Comment by xd009642** [comment]
@bytesnake ready for another review :eyes:

File: algorithms/linfa-clustering/src/optics/errors.rs [hunk 25]
```diff
@@ -0,0 +1,10 @@
+use thiserror::Error;
+pub type Result<T> = std::result::Result<T, OpticsError>;
+
+/// An error when modeling a GMM algorithm
```
bytesnake, line 4: not GMM but OPTICS

algorithms/linfa-clustering/src/lib.rs
```diff
@@ -28,10 +28,12 @@ mod dbscan;
 mod gaussian_mixture;
 #[allow(clippy::new_ret_no_self)]
 mod k_means;
+mod optics;
```
bytesnake, line 31: you have to add the OPTICS to the documentation at the top

algorithms/linfa-clustering/src/optics/algorithm.rs
```diff
@@ -0,0 +1,496 @@
...
+)]
+/// OPTICS (Ordering Points To Identify Clustering Structure) is a clustering algorithm that
+/// doesn't explicitly cluster the data but instead creates an "augmented ordering" of the dataset
+/// representing it's density-based clustering structure. This ordering contains information which
+/// is equivalent to the density-based clusterings and can then be used for automatic and
+/// interactive cluster analysis.
+///
+/// OPTICS cluster analysis can be used to derive clusters equivalent to the output of other
+/// clustering algorithms such as DBSCAN. However, due to it's more complicated neighborhood
+/// queries it typically has a higher computational cost than other more specific algorithms.
```
bytesnake, line 41: can you also add a link here to another resource for people who want to read more (e.g. the original paper or a wikipedia article)

bytesnake (approved):
added more comments, things which are missing:
 * an example demonstrating how OPTICS is used
 * mention of OPTICS in the ReadMe and at `docs/website/content/about.md`

xd009642:
Done the comments, I'll do the example later on today hopefully

bytesnake:
After merging I'll try to come up with something for my research group, perhaps we can look at acoustical scene embeddings. Cheers :beers:

xd009642:
@bytesnake I've added in the example. I save the reachability distances, dataset and indexes to npy arrays as that's all you need to recreate the example on wikipedia (computing the spanning tree from the distances is an exercise left to the reader :wink: ) 

We can see the 3 peaks and 4 regions of lower distances indicating the 4 clusters and their most unreachable points. `generate_blobs` didn't see to have anything for adding outliers so it doesn't really demonstrate anything with them, but I suppose I could set a tighter epsilon to get an equivalent output if that's desired. But I think this is good enough.

Also the plots I generated from the output:

dataset

reachability

bytesnake (approved):
The example is looking good and all the items of the review are ticked off. I think the PR is ready to merge?

bytesnake:
one thing: can you add the python script to reproduce those plots? I know it's a simple scatter and chart plot but still useful when somebody want to reproduce them

xd009642:
@bytesnake sure, do you have a preference on location? Because right now I just have it in the root of linfa-clustering

bytesnake:
perhaps the best place is alongside the example to `examples/optics_plot.py`

xd009642:
Done :+1:

xd009642:
@bytesnake for some reason the CI failed on a bunch of unrelated linfa crates I haven't touched i.e. linfa-linear...

bytesnake:
mhh let me see

bytesnake:
the `serde` features does not get activated, but this doesn't trigger any error on my machine. Can you change linfa's import to
```
linfa = { version = "0.4.0", path = "../..", features=["serde"] }
```
for `linfa-linear` and `linfa-logistic` and see whether this fixes the error. They both depend on `argmin` which requirs serde atm

xd009642:
@bytesnake looks like that fixed it, IO can see some of the previously failing jobs now pass :+1:

PR #60, opened 2020-11-20
Linear decision trees improvements

bytesnake:
This PR continues the work of #43 

 * [x] Rebase to master branch
 * [x] Use new traits for `linfa-trees`
 * [x] Add more tests to `linfa-trees`
 * [x] Add function to generate tikz code describing the tree


mossbanay:
I have exams this week and next but I'm interested in helping out with this afterward. There are only a few changes left mentioned in the previous PR.

bytesnake:
I would implement the `VotingClassifier` as a simple wrapper, taking targets and producing new targets. For example:
```rust
let params = DecisionTrees::params()..;
dataset.bootstrap(1024, 12) // subsample with replacement 12 sets, each 1024 samples
    .map(|ds| params.fit(&ds))
    .map(|obj| obj.predict(&valid))
    .fold(VotingClassifier::new(), |v, p| v.add(&p))
    .hard();
```
what do you think?

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -208,7 +208,7 @@ impl<F: Float, L: Label> TreeNode<F, L> {
 
         for i in 0..x.nrows() {
             if mask.mask[i] {
-                if x[(i, best_feature_idx)] < best_split_value {
+                if x[(i, best_feature_idx)] <= best_split_value {
```
mossbanay: Isn't this an arbitrary choice?

bytesnake: the equality was introduced to satisfy a test with only synthetic values. But `best_split_value` should actually be some kind of middle point between previous and next optimal split decision

mossbanay: Ah okay I see

bytesnake: the best_split_value is now the midpoint

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -627,19 +635,101 @@ mod tests {
...
         let targets = array![0, 0, 1];
 
         let dataset = Dataset::new(data.clone(), targets);
         let model = DecisionTree::params(2).max_depth(Some(1)).fit(&dataset);
 
         assert_eq!(&model.predict(data.clone()), &[0, 0, 1]);
     }
+
+    #[test]
+    /// Small toy dataset from sklearn
```
mossbanay, line 658: Was there sklearn code used to generate this? Even if it's in python would be nice to have documentation on how to reproduce it.

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -627,19 +635,101 @@ mod tests {
...
+        );
     }
 
     #[test]
     /// Small perfectly separable dataset test
     ///
     /// This dataset of three elements is perfectly using the second feature.
     fn perfectly_separable_small() {
-        let data = array![[1., 2., 3.5], [1., 2., 3.5], [1., 3., 3.5]];
+        let data = array![[1., 2., 3.], [1., 2., 3.5], [1., 3., 4.]];
```
mossbanay, line 648: This should be separable on the second or third feature. if we make third feature 3, 4, 3.5 then I think it's just the the second feature that should be used. I think I did the same thing by accident when tweaking the numbers.

mossbanay (commented):
Tests & lints want a review?

mossbanay (approved):
Test & lints want an _approval_?

linfa-trees/src/decision_trees/algorithm.rs
bytesnake: they were taken from here https://github.com/scikit-learn/scikit-learn/blob/0fb307bf39bbdacd6ed713c00724f8f871d60370/sklearn/tree/tests/test_tree.py#L67

linfa-trees/src/decision_trees/algorithm.rs
bytesnake: it should only use the third feature in this case. Update in the latest commit to [3,4,3.5]

linfa-trees/src/decision_trees/algorithm.rs
mossbanay: Ah nice!

mossbanay:
For looking at random forests (which I think should be in a separate PR to these improvements) we can either:
1. Pass in a mask over the features
2. Generate datasets with the features masked out
It's not immediately clear to me which would be better in the long-term so I suggest we pursue (2) for the moment since it doesn't clutter the API for now.

bytesnake:
they were moved to https://github.com/rust-ml/linfa/pull/66

mossbanay (approved):
Happy to merge this then continue on with RFs in that other PR

bytesnake:
awesome :+1: I will just write a quick function which can generates a tikz styled tree (for example [here](https://texample.net/tikz/examples/tree/)) and then merge

bytesnake:
from the example, this is not as pretty as possible, but for now is sufficient
decision-tree

mossbanay:
Nice work! That looks great. One of the more premium features in ML libraries for sure.

PR #61, opened 2020-11-22
Add KMeans n_init option and errors handling

relf:
This PR adds the <code>n_init</code> option to make multiple runs of KMeans (10 by default as in scikit-learn) as the output depends on initial centroids random initialization. 

The best output in terms of inertia (sum of squared distances of the closest centroid for all observations) is returned.
I've added some errors handling and a test, see below the visualizations:
* first clustering with one iteration 
kmeans1
* then clustering with 10 iterations (much better!)
kmeans2

### DIFF
--- a/linfa-clustering/src/k_means/algorithm.rs
+++ b/linfa-clustering/src/k_means/algorithm.rs
@@ -1,3 +1,4 @@
+use crate::k_means::errors::{KMeansError, Result};
 use crate::k_means::helpers::IncrementalMean;
 use crate::k_means::hyperparameters::{KMeansHyperParams, KMeansHyperParamsBuilder};
 use linfa::{
@@ -95,7 +96,7 @@ use serde_crate::{Deserialize, Serialize};
 /// // Let's configure and run our K-means algorithm
 /// // We use the builder pattern to specify the hyperparameters
 /// // `n_clusters` is the only mandatory parameter.
-/// // If you don't specify the others (e.g. `tolerance` or `max_n_iterations`)
+/// // If you don't specify the others (e.g. `n_init`, `tolerance`, `max_n_iterations`)
 /// // default values will be used.
 /// let n_clusters = expected_centroids.len_of(Axis(0));
 /// let model = KMeans::params(n_clusters)
@@ -151,7 +152,7 @@ impl<F: Float> KMeans<F> {
 impl<'a, F: Float, R: Rng + Clone, D: Data<Elem = F>, T: Targets> Fit<'a, ArrayBase<D, Ix2>, T>
     for KMeansHyperParams<F, R>
 {
-    type Object = KMeans<F>;
+    type Object = Result<KMeans<F>>;
 
     /// Given an input matrix `observations`, with shape `(n_observations, n_features)`,
     /// `fit` identifies `n_clusters` centroids based on the training data distribution.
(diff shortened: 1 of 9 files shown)


linfa-clustering/src/k_means/algorithm.rs
```diff
@@ -162,31 +164,50 @@ impl<'a, F: Float, R: Rng + Clone, D: Data<Elem = F>, T: Targets> Fit<'a, ArrayB
...
+                let distance = centroids
+                    .sq_l2_dist(&new_centroids)
+                    .expect("Failed to compute distance");
+                centroids = new_centroids;
+                if distance < self.tolerance() {
+                    converged_iter = Some(n_iter);
+                    break;
+                }
+            }
+            if inertia < min_inertia {
```
bytesnake, line 192: can you add a reference where inertia is coming from? The `inertia` is just the total sum of error for each cluster assignment?

linfa-clustering/src/k_means/hyperparameters.rs
```diff
@@ -30,13 +32,23 @@ pub struct KMeansHyperParams<F: Float, R: Rng> {
...
     n_clusters: usize,
     rng: R,
 }
 
 impl<F: Float, R: Rng + Clone> KMeansHyperParamsBuilder<F, R> {
+    /// Set the value of `n_init`.
+    ///
+    /// The final results will be the best output of n_init consecutive runs in terms of inertia
+    /// (sum of squared distances to the closest centroid for all observations in the training set)
+    pub fn n_init(mut self, n_init: u64) -> Self {
```
bytesnake, line 47: I think that `n_init` suggests that some pre-initialization happens before the actual K-Means. What do you think about a name like `num_runs`?

relf: I should have referenced the [scikit-learn implementation](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html). So I think it is better to keep the same name (it si also used in GMM).

relf: ... But I understand your point. In that case I would go for <code>n_runs</code> for consistency and change it also in GMM. What do you think?

bytesnake: sounds good to me :+1: we don't have to follow Scikit-Learn here and probably should try to improve the interface

bytesnake:
in future we could also implement this and perhaps avoid the restarting algorithm https://www.opdc.ox.ac.uk/_asset/file/what-to-do-when-k-means-clustering-fails-a-simple-yet-principled-alternative-algorithm.pdf 
> This novel algorithm which we call MAP-DP (maximum a-posteriori Dirichletprocess mixtures), is statistically rigorous as it is based on nonparametric Bayesian Dirichlet process mixture modeling. This approach allows us to overcome most of the limitations imposed byK-means. The number of clusters K is estimated from the data instead of being fixed a-priori as in K-means. In addition, while K-means is restricted to continuous data, the MAP-DP framework can be applied to many kinds of data, for example, binary, count or ordinal data. Also, it can efficiently separate outliers from the data.

linfa-clustering/src/k_means/algorithm.rs
relf: I have added some comments: we keep the centroids which minimize the inertia (sum of squared distances of the closest centroid for all observations) over the n runs

linfa-clustering/src/k_means/hyperparameters.rs
```diff
@@ -95,13 +112,19 @@ impl<F: Float, R: Rng + Clone> KMeansHyperParams<F, R> {
 
     pub fn new_with_rng(n_clusters: usize, rng: R) -> KMeansHyperParamsBuilder<F, R> {
         KMeansHyperParamsBuilder {
+            n_runs: 10,
```
bytesnake, line 115: I would use `n_runs: 1` as initialization

bytesnake: otherwise the algorithm would be 10 times slower in default settings

bytesnake: ah just saw that Sklearn is doing the same thing ..

bytesnake (approved):

linfa-clustering/src/k_means/hyperparameters.rs
relf: Yeah... I suppose it is a right trade-off drawn from experience.

PR #62, opened 2020-11-24
Fix GMM regularization covariance term default value

relf:
After some tests, I've forgotten to set it back to 1e-6 (as scikit-learn does). The default reg_covar value should be small but definitely not zero!

### DIFF
--- a/linfa-clustering/src/gaussian_mixture/algorithm.rs
+++ b/linfa-clustering/src/gaussian_mixture/algorithm.rs
@@ -100,6 +100,7 @@ use serde_crate::{Deserialize, Serialize};
 /// // We can retrieve the actual centroid of the closest cluster using `.centroids()` (alias of .means())
 /// let closest_centroid = &gmm.centroids().index_axis(Axis(0), dataset.targets()[0]);
 /// ```
+#[derive(Debug, PartialEq)]
 pub struct GaussianMixtureModel<F: Float> {
     covar_type: GmmCovarType,
     weights: Array1<F>,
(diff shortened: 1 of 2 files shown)


bytesnake:
can you add a test where manually setting the regularizing to 0.0 results in a failure for the decomposition? Would be nice to catch that in further evaluation

linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -548,6 +550,48 @@ mod tests {
...
+        });
+        y
+    }
+
+    #[test]
+    fn test_zeroed_reg_covar_failure() {
+        let mut rng = Isaac64Rng::seed_from_u64(42);
+        let xt = Array2::random_using((50, 1), Uniform::new(0., 1.), &mut rng);
+        let yt = function_test_1d(&xt);
+        let data = stack(Axis(1), &[xt.view(), yt.view()]).unwrap();
```
bytesnake, line 572: out of curiosity: that Cov(X, X^2) = 0 is obvious here for X ~ U(1, 0), but what's for the other two cases?

bytesnake (approved):

linfa-clustering/src/gaussian_mixture/algorithm.rs
bytesnake, line 572: out of curiosity, how does this work: houldn't be Cov(X, X) = 1/12 here because of the uniform distribution?

relf: Sorry, to be frank, I didn't do the maths. I happen to work with that test function. I tested it with scikit-learn GMM and notice the regularization glitch (it fails in scikit-learn when reg_covar is 0 too). I took for granted that the regularization improves the numerical stability avoiding problems with small eigenvalues and did not get into the details. But now... :eyes:

bytesnake: I added a second test where the covariance is zero because all points are just repeated.

Issue #70, opened 2020-12-11
Dataset improvements


bytesnake:
Most of the time the dataset is implemented for `ArrayBase<D, Ix2>` for records and `ArrayBase<D, Ix1>` for targets. We could simplify a lot of `impl`s by:
 * [ ] renaming `Dataset` to `DatasetBase`
 * [ ] implement `Dataset` and `DatasetView`

Implement for new types
 * [ ] k-folding
 * [ ] bootstrapping
 * [ ] shuffling

Sauro98:
Hi, I would love to help with this, but I can't say that I understand completely what you have in mind. Could you explain it a little more? Is the following what you had in mind?
* Keep what's already implemented for `Dataset` but rename the struct `DatasetBase` (kind of like ndarray with Array and ArrayBase?)
* Add new types `Dataset` and `DatasetView` that use `ArrayBase` and `ArrayView` for the records and targets like you described. Would one do this by `pub type Dataset = DatasetBase<ArrayBase<D, Ix2>,ArrayBase<D, Ix1>>` ?

bytesnake:
both points are right, we would just follow the ndarray model more closely. Because of the coming holidays, my plan was to schedule that for beginning of next year. You can give it a try, though it's rather diligence work as all algorithms have to be updated :D

Sauro98:
Well I'll try then and I'll let you know how it goes 👍🏻

Sauro98:
Hi, is there some specific reason why some datasets in the `datasets` crate are returned using a `Vec` for the targets instead of an `Array1`?

bytesnake:
no this should default to `Array1`

Sauro98:
I have one doubt about `iter_fold`. Shouldn't iterator items possibly outlive the object from which the iterator is created? I mean, one would have to:

*  take a mutable reference to the original dataset
*  split that reference into mutable slices
*  for each item in the iterator swap the content of some slices and then recombine the slices in two arrays

But at that point, the content of the various items would overlap, and calling collect on that iterator would yield a mess right? I mean if one wanted to use two items from the iterator at the same time then there would be a conflict on the contents of the slices composing the two items right?

Wouldn't such implementation need something like [this proposal](https://github.com/rust-lang/rfcs/blob/master/text/1598-generic_associated_types.md#motivation)?

I'm asking because as a rust noob I just can't wrap my head around it

bytesnake:
[...]

PR #74, opened 2020-12-20
Dataset improvements

Sauro98:
Hi, I'm opening the PR a little early because I'd like to ask some questions. This PR wants to make the changes proposed in #70.

* struct `Dataset` has been renamed `DatasetBase`
* All crates have been updated according to the renaming of `Dataset` into `DatasetBase`
* New types have been added:
    * `Dataset`: owned `Array2` as records, owned `Array1` as targets. The current implementation allows to write `Dataset<f24,f24>` or `Dataset<f24,usize>` and so on. Is this in line with the objective of the proposed changes?
    * `DatasetView`: view of an `Array2` as records, view of an `Array1` as targets.  The current implementation allows to write `DatasetView<f24,f24>` or `DatasetView<f24,usize>` and so on. Calling `view()` on a `Dataset` produces a `DatasetView`.
 * Functions inside sub-crate `datasets` have been changed to return an instance of `Dataset`.

There's still k-folding missing but I figured that it would be better to have an intermediate check. I also added tests to ensure the  existence of the implementations of the required methods for the two new types and they also provide a look at how the two types can be used

### DIFF
--- a/linfa-clustering/benches/gaussian_mixture.rs
+++ b/linfa-clustering/benches/gaussian_mixture.rs
@@ -22,7 +22,8 @@ fn gaussian_mixture_bench(c: &mut Criterion) {
             let n_features = 3;
             let centroids =
                 Array2::random_using((n_clusters, n_features), Uniform::new(-30., 30.), &mut rng);
-            let dataset : DatasetBase<_,_> = (generate_blobs(cluster_size, &centroids, &mut rng),()).into();
+            let dataset: DatasetBase<_, _> =
+                (generate_blobs(cluster_size, &centroids, &mut rng), ()).into();
             bencher.iter(|| {
                 black_box(
                     GaussianMixtureModel::params(n_clusters)
--- a/linfa-clustering/benches/gaussian_mixture.rs
+++ b/linfa-clustering/benches/gaussian_mixture.rs
@@ -22,7 +22,7 @@ fn gaussian_mixture_bench(c: &mut Criterion) {
             let n_features = 3;
             let centroids =
                 Array2::random_using((n_clusters, n_features), Uniform::new(-30., 30.), &mut rng);
-            let dataset = DatasetBase::from(generate_blobs(cluster_size, &centroids, &mut rng));
+            let dataset : DatasetBase<_,_> = (generate_blobs(cluster_size, &centroids, &mut rng),()).into();
             bencher.iter(|| {
                 black_box(
                     GaussianMixtureModel::params(n_clusters)
--- a/linfa-clustering/benches/gaussian_mixture.rs
+++ b/linfa-clustering/benches/gaussian_mixture.rs
@@ -3,7 +3,7 @@ use criterion::{
     PlotConfiguration,
 };
 use linfa::traits::Fit;
-use linfa::Dataset;
+use linfa::DatasetBase;
 use linfa_clustering::{generate_blobs, GaussianMixtureModel};
 use ndarray::Array2;
 use ndarray_rand::rand::SeedableRng;
@@ -22,7 +22,7 @@ fn gaussian_mixture_bench(c: &mut Criterion) {
             let n_features = 3;
             let centroids =
                 Array2::random_using((n_clusters, n_features), Uniform::new(-30., 30.), &mut rng);
-            let dataset = Dataset::from(generate_blobs(cluster_size, &centroids, &mut rng));
+            let dataset = DatasetBase::from(generate_blobs(cluster_size, &centroids, &mut rng));
             bencher.iter(|| {
                 black_box(
                     GaussianMixtureModel::params(n_clusters)
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -66,8 +66,13 @@ impl<R: Records, S: Targets> DatasetBase<R, S> {
 
         self
     }
+}
 
-    pub fn map_targets<T, G: FnMut(&S::Elem) -> T>(self, fnc: G) -> DatasetBase<R, Vec<T>> {
+impl<F: Float, D: Data<Elem = F>, T: Targets> DatasetBase<ArrayBase<D, Ix2>, T> {
+    pub fn map_targets<S, G: FnMut(&T::Elem) -> S>(
+        self,
+        fnc: G,
+    ) -> DatasetBase<ArrayBase<D, Ix2>, Array1<S>> {
         let DatasetBase {
             records,
             targets,
@@ -75,7 +80,8 @@ impl<R: Records, S: Targets> DatasetBase<R, S> {
             ..
         } = self;
 
-        let new_targets = targets.as_slice().iter().map(fnc).collect::<Vec<T>>();
+        let new_targets = targets.as_slice().iter().map(fnc).collect::<Vec<S>>();
+        let new_targets = Array1::from_shape_vec(new_targets.len(), new_targets).unwrap();
 
         DatasetBase {
             records,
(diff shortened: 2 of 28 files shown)


src/dataset/mod.rs
```diff
@@ -96,3 +108,144 @@ where
...
+
+        let linspace: Array1<f64> = Array1::linspace(0.0, 0.8, 100);
+        let records = Array2::from_shape_vec((50, 2), linspace.to_vec()).unwrap();
+        let targets: Array1<f64> = Array1::linspace(0.0, 0.8, 50);
+        let dataset = Dataset::new(records, targets);
+
+        //Split with ratio view
+        let (train, val) = dataset.split_with_ratio_view(0.5);
+        assert_eq!(train.targets().len(), 25);
+        assert_eq!(val.targets().len(), 25);
```
Sauro98, line 146: `split_with_ratio_view` for type `Dataset` uses the already existing implementation of the method, but this returns a slice as the targets. Would it be better to re-implement it for `Dataset` so that it returns an `ArrayView1` as the targets?

src/dataset/impl_dataset.rs
```diff
@@ -247,13 +247,122 @@ impl<F: Float, D: Data<Elem = F>, I: Dimension> From<ArrayBase<D, I>>
...
+
+        let new_targets = targets.iter().map(fnc).collect::<Vec<T>>();
+        let new_targets = Array1::from_shape_vec(new_targets.len(), new_targets).unwrap();
+
+        DatasetBase {
+            records,
+            targets: new_targets,
+            weights,
+        }
+    }
```
Sauro98, line 278: I had to add it in order to change the return type of functions in the `datasets` sub-crate, since `map_targets` puts the new targets in a `Vec`

bytesnake:
awesome! I'm currently traveling and will be back end of december, but the changes are looking good :+1: for `map_targets`: just drop the Vec implementation, it is only confusing for the user

20.12.2020 10:59:02 Ivano Donadi <notifications@github.com>:

bytesnake:
yes!

20.12.2020 10:56:43 Ivano Donadi <notifications@github.com>:

Sauro98:
Hi, I've added k-folding for `Dataset` and `DatasetView` types, implemented the changes mentioned in the previous comments and I've added a shy attempt at documentation. There are some questions I would like to ask:
*  Is that style/content of documentation alright? If so, or even if you have some suggestions, I would gladly write some more for the `impl_dataset.rs` file
* I see that in general the methods in this file don't check for correctness of input parameters, and so I left the `k` parameter for k-folding unchecked. Is this alright? Or else, should methods instead return a `Result` catching any error with the inputs?

Of course, I'm here for any comments about the changes already implemented. Happy holidays ☃️

bytesnake:
> Hi, I've added k-folding for `Dataset` and `DatasetView` types, implemented the changes mentioned in the previous comments and I've added a shy attempt at documentation. There are some questions I would like to ask:
> 
>     * Is that style/content of documentation alright? If so, or even if you have some suggestions, I would gladly write some more for the `impl_dataset.rs` file
> 

They are looking good, I think that documentation should really be improved. You can take this blogpost as a guide how to document functions: https://deterministic.space/machine-readable-inline-markdown-code-cocumentation.html

>     * I see that in general the methods in this file don't check for correctness of input parameters, and so I left the `k` parameter for k-folding unchecked. Is this alright? Or else, should methods instead return a `Result` catching any error with the inputs?

yes we should, but only once we have proper error handling with `thiserror`

> 
> Of course, I'm here for any comments about the changes already implemented. Happy holidays snowman_with_snow

thanks :)

src/dataset/mod.rs
```diff
@@ -70,6 +70,18 @@ where
     weights: Vec<f32>,
 }
 
+/// Dataset
+///
+/// The most commonly used typed of dataset. It contains a number of records
+/// stored as an `Array2` and each record corresponds to a single target. Such
+/// targets are stored as an `Array1`.
+pub type Dataset<D, T> = DatasetBase<ArrayBase<OwnedRepr<D>, Ix2>, ArrayBase<OwnedRepr<T>, Ix1>>;
```
bytesnake, line 78: should we have a `type DatasetPr<D> = DatasetBase<ArrayBase<OwnedRepr<D>, Ix2>, ArrayBase<OwnedRepr<Pr>, Ix2>>` as well to multiclass probabilities :thinking:

bytesnake (commented):
thanks for 
 * switching to `DatasetBase`
 *  extending testing and documentation
 * introducing k-folding

bytesnake:
I merged the naive bayes PR, can you rebase?

bytesnake:
one idea which would improve readability:
add a field to `DatasetBase` called `features: Vec<String>` to give each feature a name. This could be used in printing correlation or decision tree etc.
But this would probably go into a follow up PR as these are already enough changes

linfa-clustering/benches/gaussian_mixture.rs
```diff
@@ -22,7 +22,7 @@ fn gaussian_mixture_bench(c: &mut Criterion) {
             let n_features = 3;
             let centroids =
                 Array2::random_using((n_clusters, n_features), Uniform::new(-30., 30.), &mut rng);
-            let dataset = Dataset::from(generate_blobs(cluster_size, &centroids, &mut rng));
+            let dataset = DatasetBase::from(generate_blobs(cluster_size, &centroids, &mut rng));
```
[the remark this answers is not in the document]
Sauro98: `Dataset<_,_>` at the moment does not allow for `()` targets and so `generate_blobs` can't be cast to a dataset, unless it is coupled with some kind of empty array

src/dataset/mod.rs
Sauro98: I think that would be a handy shortcut

Sauro98 (commented):
did not mean to start a review

src/dataset/impl_dataset.rs
```diff
@@ -247,13 +247,237 @@ impl<F: Float, D: Data<Elem = F>, I: Dimension> From<ArrayBase<D, I>>
...
+                Dataset::new(remaining_records, remaining_targets),
+                // validation
+                Dataset::new(folds_records[0].into_owned(), folds_targets[0].into_owned()),
+            ));
+
+            folds_records.rotate_left(1);
+            folds_targets.rotate_left(1);
+        }
+        res
+    }
```
Sauro98, line 482: How does this look? Is there a better way than shifting the vectors at each iteration to generate all permutations?

Sauro98: Maybe collecting into an `Array1` and then using `select` on the chunks indices?

linfa-clustering/benches/gaussian_mixture.rs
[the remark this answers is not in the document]
bytesnake: that's true

src/dataset/impl_dataset.rs
bytesnake: yes much better :+1: I did not anticipate that you cannot concatenate views in ndarray
perhaps we can do this without copying by:
 * changing to `fn fold(&'a mut self, k: usize) -> impl Iterator<Item = (Dataset<_, _>, Dataset<_, _>)>`
 * destructing the arrays with `as_slice` and swapping the validation set into place with https://doc.rust-lang.org/stable/std/primitive.slice.html#method.swap_with_slice
 *  return a view with `ArrayView::from_shape`

this would borrow the dataset for the k-folding and return the dataset in original order

bytesnake: perhaps we should name that `iter_fold` and add later on

src/dataset/impl_dataset.rs
```diff
@@ -43,47 +47,53 @@ impl<R: Records, S: Targets> Dataset<R, S> {
...
+        DatasetBase {
             records,
             targets: new_targets,
             weights,
         }
     }
 }
 
-impl<F: Float, T: Clone> Dataset<Array2<F>, Vec<T>> {
+impl<F: Float, T: Clone> DatasetBase<Array2<F>, Vec<T>> {
```
bytesnake, line 96: can't you drop this implementation?

src/dataset/impl_dataset.rs
Sauro98: Hi, I have a couple of doubts. Please bear with me as this is the first time I've tried to do something like this:

* The way I've understood it is: take a slice for the first fold and a slice for the rest of the array, then, at every step of the iterator, swap the content of the fold with the content of the remaining array at the position of the next fold. Is that what you meant?
* Wouldn't `swap_with_slice` panic if `k` is such that the last chunk does not have the same length as the others?

bytesnake: > Hi, I have a couple of doubts. Please bear with me as this is the first time I've tried to do something like this:
> 
>     * The way I've understood it is: take a slice for the first fold and a slice for the rest of the array, then, at every step of the iterator, swap the content of the fold with the content of the remaining array at the position of the next fold. Is that what you meant?
yes and of course swap the content back in the next step otherwise you would change order of chunks

> 
>     * Wouldn't `swap_with_slice` panic if `k` is such that the last chunk does not have the same length as the others?
I would propose to ignore the trail of the dataset and always include it in the training set

> 
>     * Doesn't `swap_with_slice` want a mutable slice?  Can I take a mutable slice from an `ArrayView`? Shouldn't it be only possible for an `ArrayViewMut`?

you should be able to create a mutable view of the underlying arrays with `pub fn view_mut(&mut self) -> ArrayViewMut<'_, A, D> `

bytesnake: are you in the middle of creating this or should we merge the PR and amend such a function in a follow-up?

Sauro98: > you should be able to create a mutable view of the underlying arrays with pub fn view_mut(&mut self) -> ArrayViewMut<'_, A, D>
So I can only implement it for `Dataset` and not `DatasetView` right? Because I can't take a mutable view of an `ArrayView`.

I'm currently trying to implement it but If you say so we can stop here for this PR and implement it another time.  I still have a commit to push with the changes you suggested. For `fold` i substitute the shifting of the `Vec` with just a swap of the first chunk with the one in the next fold.

Sauro98: I've tried my hand at it and I'm having some issues with lifetimes, I think it would be better to leave it for an eventual follow-up PR

bytesnake: :+1:

bytesnake (approved):

PR #75, opened 2021-01-06
Improve kernel tests and documentation

Sauro98:
Here is my attempt at testing and documenting the linfa-kernel crate. I tested almost all functions apart from `to_upper_traingle` and `weighted_sum` because for those I wouldn't know how to test them other than straight-up copying the contents of those methods. Still, it could be useful if one was to modify them in the future.

Some of the functions/methods used were new to me so I hope that I interpreted everything right for the documentation.

I also have a couple of doubts/suggestions:

* All members of `Kernel` are public. Wouldn't it be better not to expose them directly?
* `Kernel` exposes `new` allowing one to use a non-linear kernel while setting `is_linear` to true
* `Kernel` has its own `is_linear` flag, with the same function as `method.is_linear()`. Wouldn't it be better to remove this kind of redundancy, especially given the two points above?

Of course, I'm up to implement those changes if they are deemed useful.

### DIFF
--- a/linfa-kernel/src/lib.rs
+++ b/linfa-kernel/src/lib.rs
@@ -61,7 +61,6 @@ where
     )]
     pub method: KernelMethod<R::Elem>,
     pub dataset: R,
-    pub linear: bool,
 }
 
 impl<'a, F: Float> Kernel<ArrayView2<'a, F>> {
@@ -69,7 +68,6 @@ impl<'a, F: Float> Kernel<ArrayView2<'a, F>> {
         dataset: ArrayView2<'a, F>,
         method: KernelMethod<F>,
         kind: KernelType,
-        linear: bool,
     ) -> Kernel<ArrayView2<'a, F>> {
         let inner = match kind {
             KernelType::Dense => KernelInner::Dense(dense_from_fn(&dataset, &method)),
@@ -80,11 +78,10 @@ impl<'a, F: Float> Kernel<ArrayView2<'a, F>> {
             inner,
             method,
             dataset,
-            linear,
         }
     }
 
-    /// Performs the element-wise product between the kernel matrix
+    /// Performs the matrix product between the kernel matrix
     /// and the input
     ///
     /// ## Parameters
@@ -93,7 +90,7 @@ impl<'a, F: Float> Kernel<ArrayView2<'a, F>> {
     ///
     /// ## Returns
     ///
-    /// A new matrix containing the element-wise product between the kernel
+    /// A new matrix containing the matrix product between the kernel
     /// and `rhs`
     ///
     /// ## Panics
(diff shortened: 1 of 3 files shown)


linfa-kernel/src/lib.rs
```diff
@@ -82,13 +84,33 @@ impl<'a, F: Float> Kernel<ArrayView2<'a, F>> {
         }
     }
 
+    /// Performs the element-wise product between the kernel matrix
```
bytesnake, line 87: this performs a matrix vector and not the hadamard product

bytesnake (approved):
You could test the `to_upper_triangle` and `weighted_sum` with a gaussian kernel and a very large variance. The upper triangle should be a diagonal and the weighted sum should just contain equal entries.

> All members of Kernel are public. Wouldn't it be better not to expose them directly?

The members in `Kernel` should not be public, but was made so for the SVM implementation. 

> Kernel exposes new allowing one to use a non-linear kernel while setting is_linear to true

This is legacy code because `Kernel` used to accept a closure, but we are now just using a enum in order to serialize the structure. The `is_linear` could be removed altogether

> Kernel has its own is_linear flag, with the same function as method.is_linear(). Wouldn't it be better to remove this kind of redundancy, especially given the two points above?

yep :laughing:

linfa-kernel/src/lib.rs
Sauro98: so `mul` is just like `dot`? Isn't `rhs` a matrix too?  I think I got confused because for the linear inner product `mul` and then `sum` is used

Sauro98: Or does the resulting matrix have in each column the matrix vector product of the kernel matrix by the corresponding sample?

bytesnake: I quick checked what `mul` actually does and its behaviour is the dot product for sparse matrices (`sprs`) and the hadamard product for dense matrices (`ndarray`). I did not catch the error because all tests in `linfa-reduction` uses sparse problems :/ please change that to `dot` for the dense case

Sauro98: Whew! Will do 👍🏻

Sauro98:
> The upper triangle should be a diagonal 

But in `to_upper_triang` the diagonal is left out. Should the diagonal be included too or is that right as it is?

bytesnake:
> > The upper triangle should be a diagonal
> 
> But in `to_upper_triang` the diagonal is left out. Should the diagonal be included too or is that right as it is?

no `to_upper_triangle` is really just the upper triangle and should therefore be zero for all entries

Sauro98:
* Removed `is_linear` attribute from `Kernel` struct;
* Fixed all occurrences of `Kernel::new()` accordingly;
* Changed `mul` to `dot` for the `dot` method on dense kernels;
* Changed `dot` documentation accordingly;
* Added tests for `to_upper_triang` and `weighted_sum`

bytesnake (approved):

PR #76, opened 2021-01-17
Elastic Net

bytesnake:
First implementation of elastic net regression, which encompasses Lasso and Ridge regression. Uses coordinate descent optimisation algorithm.

This contains PR #46 adapted to new datasets


paulkoerbitz:
Thanks for completing this! Great work 👏🏾

PR #77, opened 2021-01-27
Iter fold

Sauro98:
This PR introduces the iter_fold method for k_folding. The main points are:

### Linfa/src
* Modified `Fit` trait to remove training set lifetime
* Implemented iter_folding without cloning

### Linfa-kernel
* Introduced new trait `Inner` to represent an inner matrix. Implementors of this trait are: `Array2`, `ArrayView2`, `CsMat`, `CsMatView`
* enum `KernelInner` now takes two `Inner`  as parameters, one for the dense implementation and one for the sparse implementation
* Renamed `Kernel` to `KernelBase`, which now takes the two Inner types for the `KernelInner` enum
* Added  type `Kernel<'a, F>`  which is a kernel that keeps the data in an `ArrayView2` and has an owned inner matrix (the most common case)
* Added type `KernelOwned<F>` for which the data is owned too
* Added type `KernelView<'a, F>` which has a view on both its data and inner matrix
* Implemented functions `view()` and `to_owned()` to switch between the various types

This was done to be able to have both a kernel that borrows data with a lifetime and one that owns it with just one struct

### Linfa-svm

* Modified implementation so that now `Svm`  owns the kernel data (via a `KernelOwned` attribute)
* Adjusted to new `Fit` definition and kernel types
* Added tests for `iter_fold` both for regression and classification to show that it works

### Other sub crates

* Adjuster for new `Fit` definition and kernel types

This doesn't address the points on #70 about the `Dynamic` Kernel type and Svm that only keeps its support vectors because I wanted to have a solid and approved basis before doing any additional work.

### DIFF
--- a/linfa-kernel/Cargo.toml
+++ b/linfa-kernel/Cargo.toml
@@ -25,7 +25,7 @@ features = ["std", "derive"]
 
 [dependencies]
 ndarray = "0.13"
-sprs = { version = "0.9", default-features = false }
+sprs = { version = "0.9.3", default-features = false }
 hnsw = "0.6"
 space = "0.10"
 
--- a/linfa-kernel/src/lib.rs
+++ b/linfa-kernel/src/lib.rs
@@ -1,12 +1,14 @@
 //! Kernel methods
 //!
+pub mod inner;
 mod sparse;
 
+pub use inner::{Inner, KernelInner};
 use ndarray::prelude::*;
 use ndarray::Data;
 #[cfg(feature = "serde")]
 use serde_crate::{Deserialize, Serialize};
-use sprs::CsMat;
+use sprs::{CsMat, CsMatView};
 use std::ops::Mul;
 
 use linfa::{dataset::DatasetBase, dataset::Records, dataset::Targets, traits::Transformer, Float};
@@ -20,18 +22,6 @@ pub enum KernelType {
     Sparse(usize),
 }
 
-/// Storage for the kernel matrix
-#[cfg_attr(
-    feature = "serde",
-    derive(Serialize, Deserialize),
-    serde(crate = "serde_crate")
-)]
-#[derive(Debug)]
-pub enum KernelInner<F: Float> {
-    Dense(Array2<F>),
-    Sparse(CsMat<F>),
-}
-
 /// A generic kernel
 ///
 ///
@@ -40,18 +30,20 @@ pub enum KernelInner<F: Float> {
     derive(Serialize, Deserialize),
     serde(crate = "serde_crate")
 )]
-pub struct Kernel<R: Records>
+pub struct KernelBase<R: Records, K1: Inner, K2: Inner>
 where
     R::Elem: Float,
+    K1::Elem: Float,
+    K2::Elem: Float,
 {
     #[cfg_attr(
         feature = "serde",
         serde(bound(
-            serialize = "KernelInner<R::Elem>: Serialize",
-            deserialize = "KernelInner<R::Elem>: Deserialize<'de>"
+            serialize = "KernelInner<K1, K2>: Serialize",
+            deserialize = "KernelInner<K1, K2>: Deserialize<'de>"
         ))
     )]
-    pub inner: KernelInner<R::Elem>,
+    pub inner: KernelInner<K1, K2>,
     #[cfg_attr(
         feature = "serde",
         serde(bound(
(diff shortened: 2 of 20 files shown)


linfa-kernel/Cargo.toml
```diff
@@ -25,7 +25,7 @@ features = ["std", "derive"]
 
 [dependencies]
 ndarray = "0.13"
-sprs = { version = "0.9", default-features = false }
+sprs = { version = "0.9.3", default-features = false }
```
Sauro98, line 28: I changed this to be sure of being able to use the `diag` method on sparse matrices in order to make the interface uniform

linfa-kernel/src/lib.rs
```diff
@@ -40,18 +30,20 @@ pub enum KernelInner<F: Float> {
...
+    K1::Elem: Float,
+    K2::Elem: Float,
 {
     #[cfg_attr(
         feature = "serde",
         serde(bound(
-            serialize = "KernelInner<R::Elem>: Serialize",
-            deserialize = "KernelInner<R::Elem>: Deserialize<'de>"
+            serialize = "KernelInner<K1, K2>: Serialize",
+            deserialize = "KernelInner<K1, K2>: Deserialize<'de>"
```
Sauro98, line 43: I'm not familiar with `serde` at all so I don't know if this is the right way to do this

linfa-svm/src/lib.rs
```diff
@@ -278,3 +278,52 @@ impl<'a, A: Float, T> fmt::Display for Svm<'a, A, T> {
...
+                model
+                    .predict(&valid)
+                    .map_targets(|x| **x > 0.0)
+                    .confusion_matrix(&valid)
+                    .accuracy()
+            })
+            .sum::<f32>()
+            / 4_f32;
+        assert!(avg_acc >= 0.5)
+    }
```
Sauro98, line 311: This is how `iter_fold` looks for classification using `Svm`.  The ugly part is that it is unable to determine the type of the `Record` object used by the fittable model by itself, so I'm forced to write it down explicitly. 

The only cloning happening here is during the call to `fit` where the kernel is cloned inside the SVM

bytesnake:
thanks! I will look into it the next days :)

bytesnake (approved):
okay I added some comments on the PR. This is really awesome work :+1:
I think it would be better to separate `Kernel` from `Dataset` in order to resolve the ambiguities between `Kernel` and `OwnedKernel`. I added some comments how this can be realized.

After merging this, I will investigate how to simplify the API with [Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html)

linfa-svm/src/lib.rs
```diff
@@ -278,3 +278,52 @@ impl<'a, A: Float, T> fmt::Display for Svm<'a, A, T> {
...
+
+    #[test]
+    fn test_iter_folding_for_regression() {
+        let mut dataset: Dataset<f64, f64> = linfa_datasets::diabetes();
+        let params = Svm::params().c_eps(10., 0.01);
+
+        let _avg_acc = dataset
+            .iter_fold::<Kernel<_>, _, _, _>(4, &params, |svm_params, training_set| {
```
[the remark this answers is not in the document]
Sauro98: That's a good idea! I actually think that I don't even need to move since fit takes`&self`, I think I could just use `params` inside the closure

linfa-kernel/src/lib.rs
bytesnake: this is correct, `serde` needs the signatures to understand the type

linfa-kernel/src/lib.rs
```diff
@@ -219,43 +214,100 @@ impl<'a, F: Float> Kernel<ArrayView2<'a, F>> {
             .sum()
```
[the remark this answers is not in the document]
Sauro98: Yes, that would make sense, but then what would be the input to `fit` for SVMs? Both the kernel and the dataset or either the dataset and a `KernelParams` struct?

linfa-svm/src/lib.rs
```diff
@@ -292,11 +292,11 @@ mod tests {
...
         let avg_acc = dataset
-            .iter_fold::<Kernel<_>, _, _, _>(4, &params, |svm_params, training_set| {
+            .iter_fold(4, |training_set| {
                 let train_kernel = Kernel::params()
                     .method(KernelMethod::Gaussian(80.0))
                     .transform(&training_set);
-                svm_params.fit(&train_kernel)
+                params.fit(&train_kernel)
             })
             .map(|(model, valid)| {
```
Sauro98: Is this what you meant?

bytesnake: yes that is really readable :+1:

linfa-kernel/src/lib.rs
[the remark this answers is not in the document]
bytesnake: mh no that is not so easy :/ then I would simplify the type definition later, in the end all necessary support vectors should be copied into the model

bytesnake:
you still have to remove the commented parameters, but for my part we can merge this

Sauro98:
my bad, I forgot to remove them 👍🏻

PR #78, opened 2021-02-02
Add feature naming and improve documentation

bytesnake:
Names are a great way to improve readability of feature analysis. For this purpose I also added Pearson's Correlation Coefficients (with p-value by re-sampling)

 * adds field `feature_names` to `DatasetBase`
 * adds `PearsonCorrelation`
 * augments datasets in `linfa-dataset` with feature names

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,6 +60,8 @@ ndarray-rand = "0.12"
 rand_isaac = "0.2"
 approx = "0.3"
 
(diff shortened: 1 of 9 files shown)


Sauro98 (approved):
Hi, I've looked through this and it all seems very good to me 👍🏻. Good work on adding Pearson's coefficients and p-values in addition to the names, their pretty-printing is a nice touch too! I won't add any other comments on the code because I didn't find anything to point out actually

bytesnake:
okay let's got

PR #80, opened 2021-02-06
Kernel API simplifications and SVMs improvements

Sauro98:
## linfa-kernel

As discussed on #70, I've changed the Kernel structure so that it doesn't hold the dataset in itself anymore. This allowed a simplification of the various kernel types, and now there are simply:
* `Kernel<F> = KernelBase<Array2<F>, CsMat<F>>`
* `KernelView<'a,F> = KernelBase<ArrayView2<'a,F>, CsMatView<'a,F>>` 

Since now the kernel does not have the dataset anymore, the weighted sum function has been moved to the SVM crate, since it was only used there. 

## linfa-svm

* `SvmParams` now has a `KernelParams` property that can be set with `with_kernel_params` for a generic `KernelParams` struct or otherwise with one of `with_linear`, `with_polynomial` or `with_gaussian`. The default value for the kernel method is set to be linear so that if one does not set the kernel_params property it will behave like a standard SVM
* `Fit` for SVMs now takes the input dataset instead of the transformed kernel and it applies the transformation inside itself.
* SVMs now store only the input vectors that have an alpha weight different than zero and predictions are made accordingly
* SVMs now do not store the kernel inner matrix after training anymore since they only need the kernel method to compute distances between support vectors and new samples

### DIFF
--- a/linfa-svm/src/regression.rs
+++ b/linfa-svm/src/regression.rs
@@ -1,6 +1,6 @@
 //! Support Vector Regression
 use linfa::{dataset::DatasetBase, traits::Fit, traits::Predict, traits::Transformer};
-use linfa_kernel::KernelOwned;
+use linfa_kernel::Kernel;
 use ndarray::{Array1, Array2, ArrayBase, ArrayView1, ArrayView2, Data, Ix2};
 use std::ops::Mul;
 
@@ -22,7 +22,8 @@ use super::{Float, Svm, SvmParams};
 /// * `p` - epsilon value for all targets
 pub fn fit_epsilon<F: Float>(
     params: SolverParams<F>,
-    kernel: KernelOwned<F>,
+    dataset: ArrayView2<F>,
+    kernel: Kernel<F>,
     target: &[F],
     c: F,
     p: F,
@@ -43,6 +44,7 @@ pub fn fit_epsilon<F: Float>(
         vec![F::zero(); 2 * target.len()],
         linear_term,
         targets.to_vec(),
+        dataset,
         kernel,
         vec![c; 2 * target.len()],
         params,
@@ -73,7 +75,8 @@ pub fn fit_epsilon<F: Float>(
 /// * `nu` - nu value for all targets
 pub fn fit_nu<F: Float>(
     params: SolverParams<F>,
-    kernel: KernelOwned<F>,
+    dataset: ArrayView2<F>,
+    kernel: Kernel<F>,
     target: &[F],
     c: F,
     nu: F,
@@ -100,6 +103,7 @@ pub fn fit_nu<F: Float>(
         alpha,
         linear_term,
         targets.to_vec(),
+        dataset,
         kernel,
         vec![c; 2 * target.len()],
         params,
@@ -124,10 +128,11 @@ impl<'a, F: Float> Fit<'a, Array2<F>, &Array1<F>> for SvmParams<F, F> {
     type Object = Svm<F, F>;
 
     fn fit(&self, dataset: &DatasetBase<Array2<F>, &Array1<F>>) -> Self::Object {
-        let kernel = self.kernel.transform(dataset.records().clone());
+        let kernel = self.kernel.transform(dataset.records());
         match (self.c, self.nu) {
             (Some((c, eps)), _) => fit_epsilon(
                 self.solver_params.clone(),
+                dataset.records().view(),
                 kernel,
                 dataset.targets().as_slice().unwrap(),
                 c,
@@ -135,6 +140,7 @@ impl<'a, F: Float> Fit<'a, Array2<F>, &Array1<F>> for SvmParams<F, F> {
             ),
             (None, Some((nu, eps))) => fit_nu(
                 self.solver_params.clone(),
+                dataset.records().view(),
                 kernel,
                 dataset.targets().as_slice().unwrap(),
                 nu,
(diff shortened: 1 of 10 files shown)


linfa-svm/src/regression.rs
```diff
@@ -43,19 +44,14 @@ pub fn fit_epsilon<F: Float>(
...
         false,
     );
 
-    let mut res = solver.solve();
-
-    for i in 0..target.len() {
-        let tmp = res.alpha[i + target.len()];
-        res.alpha[i] -= tmp;
-    }
-    res.alpha.truncate(target.len());
```
Sauro98, line 58: This had to be moved inside the solver because the precomputed vector for the linear kernel case had the wrong weights in it. I've now added a regression test to check the correctness of predictions with the precomputed vector too.

linfa-svm/src/regression.rs
```diff
@@ -223,7 +227,7 @@ impl<D: Data<Elem = f64>> Predict<ArrayBase<D, Ix2>, Vec<f64>> for Svm<f64, f64>
             .map(|data| {
                 let val = match self.linear_decision {
                     Some(ref x) => x.mul(&data).sum() - self.rho,
-                    None => self.kernel.weighted_sum(&self.alpha, data.view()) - self.rho,
+                    None => self.weighted_sum(data.view()) - self.rho,
```
Sauro98, line 230: I think we could use an enum here like the one used for the kernel inner matrix so that it either holds the precomputed weights in the linear case or either the support vectors. In that way, we could avoid storing the support vectors in the linear case. I can't think of what could be an appropriate name for such a structure tough

linfa-svm/src/solver_smo.rs
```diff
@@ -841,14 +863,27 @@ impl<'a, F: Float, K: 'a + Permutable<F>> SolverState<'a, F, K> {
...
         };
 
+        // Select only vectors that have non-zero influence on the outcome
+        // of future predictions
+        let support_vectors = self.dataset.select(
+            Axis(0),
+            &alpha
+                .iter()
+                .enumerate()
+                .filter(|(_, a)| !a.is_zero())
```
bytesnake, line 873: can you use a threshold here, because sometimes the solution can be very small but non-zero. I used `1e-5` previously but perhaps a multiple of the `F::eps` would be more sensible

linfa-svm/src/regression.rs
bytesnake: but how does moving this into the solver changes the solution? I missed a test here :/

bytesnake (approved):
awesome work:
 * the whole API around kernel matrices is much simpler now :+1:
 * the special case of support vectors are separated from the kernel matrix now
 * much better documentation

I have one nitpicks:
 * do you think that `gaussian_kernel` instead of `with_gaussian` would make the API more readable, because `with_*` normally means that some object is passed

linfa-svm/src/solver_smo.rs
```diff
@@ -825,14 +830,31 @@ impl<'a, F: Float, K: 'a + Permutable<F>> SolverState<'a, F, K> {
...
+
         // if the kernel is linear, then we can pre-calculate the dot product
         let linear_decision = if self.kernel.inner().is_linear() {
-            let mut tmp = Array1::zeros(self.kernel.inner().dataset.len_of(Axis(1)));
-            for (i, elm) in self.kernel.inner().dataset.outer_iter().enumerate() {
+            let mut tmp = Array1::zeros(self.dataset.len_of(Axis(1)));
+
+            for (i, elm) in self.dataset.outer_iter().enumerate() {
                 tmp.scaled_add(self.target(i) * alpha[i], &elm);
             }
```
Sauro98, line 860: Here, when pre-calculating the dop product in the linear case only the first `n` values of alpha are used but in the case of regression there are `2n` alpha values, so you need to do `alpha[i] - alpha[n+i]` before you pre-calculate the dot product, otherwise the coefficients will be wrong

linfa-svm/src/regression.rs
```diff
@@ -290,17 +292,33 @@ pub mod tests {
...
+
+        // Test the precomputed dot product in the linear kernel case
+        let model = Svm::params().nu_eps(2., 0.01).with_linear().fit(&dataset);
+
+        println!("{}", model);
+
+        let predicted = Array1::from(model.predict(dataset.records()));
+        assert!(predicted.mean_squared_error(&dataset.targets().view()) < 1e-2);
+    }
 }
```
Sauro98, line 324: This is the test I meant. Without computing the correct values of alpha for regression the mean squared error is around 5000

linfa-svm/src/regression.rs
Sauro98: Yes, that sounds like a good name for it

linfa-svm/src/solver_smo.rs
bytesnake: ah I see

bytesnake: can you add a comment for the code block that this is a special case of eps regression, because the it's difficult to understand on first glance

linfa-svm/src/solver_smo.rs
Sauro98: I set it to `100 * F::eps` so to get ~1e-5 for f32 and ~2e-14 for f64, just FYI

Sauro98:
* Renamed `with_gaussian` to `gaussian_kernel`, and the same for the polynomial and linear kernels
* Left the naming `with_kernel_params` for the general method since it takes the `KernelParams` structure in input
* Comparison to zero for alpha now check an absolute value smaller than `100 * F::eps()`
* Added enum `SeparatingHyperplane` to differentiate between the linear case and the support vectors case
* Extended comments on alpha calculations inside the solver
* Removed unnecessary lifetimes in linfa-hierarchical

PR #86, opened 2021-02-13
linfa trees documentation and examples

Sauro98:
Expanding linfa trees documentation and adding some examples. Extending Tikz functionality.

### `linfa-trees`

* Extended documentation of public APIs, added examples  to the main structures
* Extended documentation and comments for internal functions to facilitate future changes
* Fixed left/right mismatch for the tree branches names
* Added the feature name to the internal nodes  of the tree
* Modified Tikz so that when the `legend` flag is set a legend with  the feature names is produced
* Modified Tikz so that if the `complete` flag is not set then it doesn't import packages and it doesn't add the beginning and end of the document 
* Modified Tikz to reflect the actual branch naming conventions used
* switched from rand Isaac to SmallRng to conform to the other sub crates

### `linfa`

* modified some dataset methods like `shuffle` and `split` so that they preserve feature names and, in the case of `split`, the weights too

### DIFF
--- a/linfa-trees/README.md
+++ b/linfa-trees/README.md
@@ -14,7 +14,7 @@ Decision Trees (DTs) are a non-parametric supervised learning method used for cl
 
 ## Examples
 
-There is an example in the `examples/` directory on how to use decision trees. To run, use:
+There is an example in the `examples/` directory showing how to use decision trees. To run, use:
 
 ```bash
 $ cargo run --release --example decision_tree --features linfa/intel-mkl-system 
--- a/linfa-trees/README.md
+++ b/linfa-trees/README.md
@@ -17,10 +17,11 @@ Decision Trees (DTs) are a non-parametric supervised learning method used for cl
 There is an example in the `examples/` directory on how to use decision trees. To run, use:
 
 ```bash
-$ cargo run --release --example decision_tree --features linfa/intel-mkl-system
+$ cargo run --release --example decision_tree --features linfa/intel-mkl-system 
 ```
 
-This generates the following tree:
+This generates the following tree: 
+
 <p align="center">
    <img src="./iris-decisiontree.svg">
 </p>
--- a/linfa-trees/README.md
+++ b/linfa-trees/README.md
@@ -14,7 +14,7 @@ Decision Trees (DTs) are a non-parametric supervised learning method used for cl
 
 ## Examples
 
-There is an example in the `examples/` directory how to use decision trees. To run, use:
+There is an example in the `examples/` directory on how to use decision trees. To run, use:
 
 ```bash
 $ cargo run --release --example decision_tree --features linfa/intel-mkl-system
--- a/linfa-trees/src/decision_trees/algorithm.rs
+++ b/linfa-trees/src/decision_trees/algorithm.rs
@@ -10,7 +10,7 @@ use super::hyperparameters::{DecisionTreeParams, SplitQuality};
 use super::NodeIter;
 use super::Tikz;
 use linfa::{
-    dataset::{Labels, Records},
+    dataset::{Labels, Records, Targets},
     traits::*,
     DatasetBase, Float, Label,
 };
@@ -472,7 +472,7 @@ impl<F: Float, L: Label + std::fmt::Debug> TreeNode<F, L> {
 /// // Fit the tree
 /// let tree = DecisionTree::params().fit(&dataset);
 /// // Get accuracy on training set
-/// let accuracy = tree.predict(dataset.records()).confusion_matrix(&dataset).accuracy();
+/// let accuracy = tree.predict(&dataset).confusion_matrix(&dataset).accuracy();
 ///
 /// assert!(accuracy > 0.9);
 ///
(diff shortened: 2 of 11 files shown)


Sauro98:
I played around with the svg to make the tree example visible in the readme. It is not pretty by **any** means but it is visible 🤣.

linfa-trees/README.md
```diff
@@ -14,13 +14,14 @@ Decision Trees (DTs) are a non-parametric supervised learning method used for cl
 
 ## Examples
 
-There is an example in the `examples/` directory how to use decision trees. To run, use:
+There is an example in the `examples/` directory on how to use decision trees. To run, use:
```
bytesnake, line 17: perhaps `showing how to use`?

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -648,20 +797,23 @@ mod tests {
...
+
+        for (imp, truth) in model.feature_importance().iter().zip(&ground_truth) {
+            assert_abs_diff_eq!(imp, truth);
+        }
 
         // check for perfect accuracy
         let cm = model.predict(dataset.records()).confusion_matrix(&dataset);
-        assert!(cm.accuracy() == 1.0);
+        // can't be more than one, but >= avoids clippy complaining about float exact comparison
+        assert!(cm.accuracy() >= 1.0);
```
bytesnake, line 810: we should use `approx` most of the time, this can be expressed as `assert_abs_diff_eq!(.., 1.0, epsilon = 1e-15)` as well

linfa-trees/src/decision_trees/algorithm.rs
```diff
@@ -361,22 +489,36 @@ pub struct DecisionTree<F: Float, L: Label> {
...
     /// Make predictions for each row of a matrix of features `x`.
-    fn predict(&self, x: ArrayBase<D, Ix2>) -> Vec<L> {
-        x.genrows()
-            .into_iter()
-            .map(|row| make_prediction(&row, &self.root_node))
-            .collect()
+    fn predict(&self, x: ArrayBase<D, Ix2>) -> Array1<L> {
+        x.map_axis(Axis(1), |observation| {
+            make_prediction(&observation, &self.root_node)
+        })
```
bytesnake, line 513: another implementation which came to my mind: take a `&Dataset<R, T>` and returns an `Array1<L>`, then you don't have to call `.predict(dataset.records())`, but just `.predict(&dataset)`

linfa-trees/src/decision_trees/hyperparameters.rs
```diff
@@ -7,20 +7,50 @@ use std::marker::PhantomData;
 #[cfg(feature = "serde")]
 use serde_crate::{Deserialize, Serialize};
 
-/// The possible impurity measures for training.
+/// The metric used to decide the feature on which to split a node
```
bytesnake, line 10: perhaps `The metric used to determine the feature by which a node is splitted`

src/dataset/impl_dataset.rs
```diff
@@ -194,8 +194,17 @@ impl<F: Float, T: Targets, D: Data<Elem = F>> DatasetBase<ArrayBase<D, Ix2>, T>
...
-        let dataset1 = DatasetBase::new(first, first_targets);
-        let dataset2 = DatasetBase::new(second, second_targets);
+        let (first_weights, second_weights) = if self.weights.len() == self.observations() {
+            (self.weights[..n].to_vec(), self.weights[n..].to_vec())
+        } else {
+            (Vec::new(), Vec::new())
+        };
+        let dataset1 = DatasetBase::new(first, first_targets)
+            .with_weights(first_weights)
+            .with_feature_names(self.feature_names());
```
bytesnake, line 204: oh well :sweat_smile:

bytesnake (approved):
Awesome :+1: added two suggestions on wording and perhaps an additional `Predict` implementation

linfa-trees/src/decision_trees/algorithm.rs
bytesnake: do you know the default precision without specifying `epsilon`?

bytesnake:
> I played around with the svg to make the tree example visible in the readme. It is not pretty by **any** means but it is visible rofl.

I should use the dark mode more often

linfa-trees/src/decision_trees/algorithm.rs
Sauro98: Yes, that would really work well

linfa-trees/src/decision_trees/algorithm.rs
Sauro98: I found this in the approx repo:
```rust

pub trait AbsDiffEq<Rhs = Self>: PartialEq<Rhs>
where
    Rhs: ?Sized,
{
    ....
    /// The default tolerance to use when testing values that are close together.
    ///
    /// This is used when no `epsilon` value is supplied to the [`abs_diff_eq!`], [`relative_eq!`],
    /// or [`ulps_eq!`] macros.
    fn default_epsilon() -> Self::Epsilon;
    ....
}

...

macro_rules! impl_signed_abs_diff_eq {
    ($T:ident, $default_epsilon:expr) => {
        impl AbsDiffEq for $T {
            ....
            fn default_epsilon() -> $T {
                $default_epsilon
            }
            ....
        }
    };
}

....

impl_signed_abs_diff_eq!(f32, f32::EPSILON);
impl_signed_abs_diff_eq!(f64, f64::EPSILON);
```
So it uses the epsilon value of the floating point type

PR #87, opened 2021-02-15
Add website to infrastructure

bytesnake:
This PR adds an entry website, where we can pitch the idea about linfa and have some announcements/news section.

 * [x] add static page based on the Zola generator to `docs/website`: [rendered](https://bytesnake.github.io/linfa/)
 * [x] move all implementations into `algorithms/` because the root became too messy
 * [x] adds workflows which automatically builds the latest rust docs and website and publishes them to `gh-pages`
 * [x] complete announcement in news section

sorry for the large number of file changes, you can search for diffs in a subpath with `git diff master docs/website/` :sweat_smile:


docs/website/config.toml
```diff
@@ -0,0 +1,17 @@
+# The URL the site will be built for
+base_url = "https://bytesnake.github.io/linfa/"
```
quietlychris, line 2: Do we want this to be hosted at bytesnake.github.io, or create a rust-ml.github.io page?

quietlychris (approved):
Besides just wondering if we want to host directly from the rust-ml group, everything else looks good! I've clicked around the current site, and didn't run into any bugs. If you're satisfied, this looks good to go from me :)

README.md
```diff
@@ -4,6 +4,7 @@
 
 [![crates.io](https://img.shields.io/crates/v/linfa.svg)](https://crates.io/crates/linfa)
 [![Documentation](https://docs.rs/linfa/badge.svg)](https://docs.rs/linfa)
+[![DocumentationLatest](https://img.shields.io/badge/docs-latest-blue)](https://rust-ml.github.com/linfa/rustdocs/linfa/)
```
Sauro98, line 7: This gives me a 404 error

algorithms/linfa-elasticnet/examples/elasticnet.rs
```diff
@@ -7,7 +7,7 @@ fn main() -> Result<()> {
 
     // train pure LASSO model with 0.1 penalty
     let model = ElasticNet::params()
-        .penalty(0.1)
+        .penalty(0.3)
```
Sauro98, line 10: Should the comment be updated too?

docs/website/content/snippets/elasticnet.md
```diff
@@ -0,0 +1,19 @@
...
+let model = ElasticNet::params()
+    .penalty(0.1)
+    .l1_ratio(1.0)
+    .fit(&train)?;
+
+println!("z score: {:?}", model.z_score());
+
+// validate
+let y_est = model.predict(&valid);
+println!("predicted variance: {}", valid.r2(&y_est));
```
Sauro98, line 18: Should this be `y_est.r2(&valid)`? The current r2 implementation assumes that the function argument is the ground truth and that `self` is the predicted values array

Sauro98 (approved):
Fantastic work on the website💪🏻 I believe that cleaning up the main folder by moving all the algorithms to their subfolder was a very good idea too and it looks much nicer now.

I only have two suggestions:
* Should we add the link to the website in the readme for linfa?
* In the `About` page of the website (in the mobile version) the table obviously does not fit, I was wondering if maybe it would be better to make only the table side-scrollable and not the whole page? This is a very picky comment so feel free to ignore it 😅

docs/website/config.toml
bytesnake: no I forgot to update the link, this will be only valid after merging the PR

README.md
bytesnake: yeah the rustdocs are built after the PR because we don't want to publish every branch/PR into `gh-pages` branch

algorithms/linfa-elasticnet/examples/elasticnet.rs
bytesnake: should be `0.1` here, I was playing around with the example

docs/website/content/snippets/elasticnet.md
bytesnake: the `r2` score is symmetric, but perhaps we are running into implementation issues, because `valid` uses multiple targets? Should definitely be `y_est.r2(&valid)?` though because this may fail

bytesnake:
> Fantastic work on the website💪🏻 I believe that cleaning up the main folder by moving all the algorithms to their subfolder was a very good idea too and it looks much nicer now.
> 
> I only have two suggestions:
> 
>     * Should we add the link to the website in the readme for linfa?
> 
>     * In the `About` page of the website (in the mobile version) the table obviously does not fit, I was wondering if maybe it would be better to make only the table side-scrollable and not the whole page? This is a very picky comment so feel free to ignore it sweat_smile


yes and yes, will figure something out for the table later this day :)

bytesnake:
> Besides just wondering if we want to host directly from the rust-ml group, everything else looks good! I've clicked around the current site, and didn't run into any bugs. If you're satisfied, this looks good to go from me :)

I would publish the website at `rust-ml.github.io/linfa/`, because it's also under the `rust-ml` organization

docs/website/content/snippets/elasticnet.md
Sauro98: You are right about the r2, sorry 😅 
It should have the `?` now because with the recent changes now it returns a `Result`, but the only difference between `y_est.r2(&valid)?` and `valid.r2(&y_est)?` I believe would be that one is `r2` and the other is `[r2]`

bytesnake:
okay let's merge

bytesnake:
damn the links in README are not updated :sweat_smile: I will push another commit when the website is online

bytesnake:
fixed in latest commit and everything seems fine :+1:

Sauro98:
Everything seems to be working fine 🚀

relf:
Links to algorithms in [About](https://rust-ml.github.io/linfa/about/) page are broken.

bytesnake:
yep fixed in f79fe9e053d7387a11a0682590e2c7cb56640444

relf:
Nice website! Bravo! :clap: :clap: :clap:

PR #88, opened 2021-02-16
Introduce multi targets and probability matrices

bytesnake:
This PR introduces multi-target datasets and makes it easier to work with the `Dataset` API. 

## Changes to datasets
 * rename `Targets` to `AsTargets` and introduce `AsTargetsMut` for mutable borrowing and `FromTargetArray` for counted labels construction
 * default to multi-target datasets in `pub type Dataset..`
 * re-name `observations` to `nsamples` and add `ntargets`
 * rename `frequencies_with_mask` to `label_frequencies_with_mask`
 * add `sample_iter`, `target_iter`, `feature_iter` iterators
 * rename `axis_chunks_iter` to `sample_chunks`
 * rename `bootstrap` to `bootstrap_samples` and add `bootstrap` (simultaneously sample features and samples) and `bootstrap_features`
 * add a `try_single_target -> Result<..>` which may return an `Error::MultipleTargets` for use in ROC/SVM/etc.
 * add `FromTarget` in order to accept `Array1` in `Dataset::new` and use `Targets == Array2<()>` for empty targets
 * add `PredictRef` which is used by a "proxy" trait `Predict` to derive all possible input/output combinations

## Other changes
 * add tests to `linfa-datasets`
 * add error types with `thiserror` where possible

## Updates for sub-crates
 * [x] linfa-reduction
 * [x] linfa-elasticnet
 * [x] linfa-bayes
 * [x] linfa-clustering
 * [x] linfa-hierarchical
 * [x] linfa-kernel
 * [x] linfa-linear
 * [x] linfa-ica
 * [x] linfa-logistic
 * [x] linfa-svm
 * [x] linfa-trees

## Other changes
 * [x] explain mechanism and difference between `Predict` and `PredictRef` in `CONTRIBUTION.md`
 * [x] improve implementation of `src/metrics_regression.rs` to accept `DatasetBase` variants
 * [x] add wrapper `MultiTargetModel` which can merge multiple single-target models 
 * [ ] (improve story around `DatasetPr`, probably in a later PR)

PR #89, opened 2021-02-26
Add physical exercise dataset with multiple targets

relf:
Dataset references:
* [scikit-learn](https://scikit-learn.org/stable/datasets/toy_dataset.html#linnerrud-dataset)
* [Tenenhaus, M. (1998). La regression PLS: theorie et pratique. Paris: Editions Technic.](https://books.google.fr/books?hl=fr&lr=&id=OesjK2KZhsAC&oi=fnd&pg=PA43&dq=Tenenhaus,+M.+(1998).+La+regression+PLS:+theorie+et+pratique.+Paris:+Editions+Technic.&ots=ExNsw80zCK&sig=BQkSlPc36v9GMy6oqhq0cHd5O9o#v=onepage&q=Linnerud&f=false)

Handling enabled in linfa with #88


datasets/src/lib.rs
```diff
@@ -121,6 +127,33 @@ pub fn winequality() -> Dataset<f64, usize> {
         .with_feature_names(feature_names)
 }
 
+#[cfg(feature = "linnerud")]
+/// Read in the physical exercise dataset from dataset path.
+///
+/// Linnerud dataset contains 20 samples collected from 20 middle-aged men in a fitness club.
```
bytesnake, line 133: this is the smallest dataset I ever encountered :laughing:

bytesnake (commented):
looking good to me, but according to Github your archive is empty. Can you re-order the struct in `src/iter.rs` so that clippy is happy, I don't know why this wasn't caught before

datasets/src/lib.rs
relf: > but according to Github your archive is empty

I made the gz archives on Windows :grin:. I am going to use a Unix box...
> Can you re-order the struct in src/iter.rs

Ok

relf:
I have integrated the linnerud dataset without thinking too much about its usage (till now! :laughing: ) and I have used a conversion of the integer targets as usize (shamelessly copying from wine quality stuff). 

But actually, I want to use the dataset as floats... Should we add some conversion helpers to `Dataset`? Or should I just remove the `as usize` conversion?

bytesnake:
you probably don't want usize targets here, except when you want to do ordinal classification

bytesnake (approved):

PR #93, opened 2021-03-08
Preprocessing

Sauro98:
Scaling. WIP

- [x] linear scaling (normal, min max, max abs)
- [x] norm scaling (l1, l2, max abs)
- [x] whitening
    - [x] PCA
    - [x] ZCA
    - [x] Cholesky
- [x] count vectorization
    - [x] base
    - [x] n-grams
    - [x] min-max df
    - [x] unicode normalization
    - [x]  fit string array
    - [x] fit vocabulary
    - [x] fit files (vec of paths) with choice of encoding
    - [x] custom split with regex
    - [x] transform array of strings
    - [x] transform files (vec of paths)
    - [x] allow sparse matrix transformation
    - [x] optional stopwords
 
- [x] tf idf normalization
    - [x] smooth, non smooth, textbook tf-idf functions
    - [x] wrap over all count vectorization methods

- [x] benchmarks with `iai`
    - [x] download `20news-bydate`  dataset in set up
    - [x] delete dataset in tear down?
    - [x] basic benchmarks 
    - [x] integration into CI
    - [x] deeper benchmarks
- [x] check feature names/weights retainment (do not use `with records` which states it doesn't in the docs)
- [x] update readme/docs with the list of all the preprocessing algorithms
- [x] examples
    - [x] count vectorization
    - [x] tf-idf
    - [x] scaling
    - [x] whitening


algorithms/linfa-preprocessing/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+            ScalingMethod::MaxAbs => FittedScaler::max_abs(x.records()),
+            ScalingMethod::Norm => FittedScaler::norm(),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct FittedScaler<F: Float> {
+    means: Option<Array1<F>>,
+    std_devs: Option<Array1<F>>,
```
jettify, line 67: I think this approach might be simpler:
https://github.com/onnx/sklearn-onnx/blob/a8267e7ba946d8b0596951060e5dca39fec47439/skl2onnx/operator_converters/scaler_op.py#L32-L58

With `offset` and `scale` one can do most linear scalers.

Sauro98: I see, thanks for the input! much better than `mins`/`maxes`/`means` etc...
Right now this PR is in a very early stage, I'll write more compliance tests and then go back to simplify the whole API.
If you're interested I can ping you again when this is in a more refined state 👍🏻

Sauro98:
Can I get an initial review on this? Below are some sparse thoughts on the PR/possible additions/improvements
* There are for sure some performance improvements possible in the count vectorizer but I want to write some benchmarks before diving into that
* I see that sklearn allows for a normalization method to be defined for the count vectorizer, but here I think it would be less confusing to just use the count vectorizer as is and then applying the NormScaler to it (might be worthy of an example)
* sklearn uses the `20news-bydate` dataset to test/bench the vectorizer but that is a 15 MB compressed dataset, I think it might be too big to just include it in the dataset subcrate? Related to this, would It be useful to add a `fit_files` function to the vectorizer to pass in filenames instead of documents?
* I don't know how I feel about the TFIDF struct being basically just a wrapper around count vectorizer that just returns a different struct when fitting

bytesnake:
I'm starting a review, but may take a day until I'm through :smile:

Sauro98:
Thank you, no rush 👍🏻

algorithms/linfa-preprocessing/src/linear_scaling.rs
```diff
@@ -0,0 +1,560 @@
...
+    fn transform(&self, x: Array2<F>) -> Array2<F> {
+        if x.is_empty() {
+            return x;
+        }
+        let mut x = x;
+        Zip::from(x.gencolumns_mut())
+            .and(self.offsets())
+            .and(self.scales())
+            .apply(|mut col, &offset, &scale| {
+                col.mapv_inplace(|el| (el - offset) * scale);
```
bytesnake, line 263: I think there is an error in only-variance-scaling mode here: You've set offset=0 for this case, but is should be: `(el - offset) * scale + offset` to retain the original offset?

algorithms/linfa-preprocessing/src/tf_idf_vectorization.rs
```diff
@@ -0,0 +1,204 @@
...
+            TfIdfMethod::Textbook => (n as f64 / (1. + df as f64)).ln(),
+        }
+    }
+}
+
+/// Simlar to [`CountVectorizer`](studct.CountVectorizer.html) but instead of
+/// just counting the term frequency of each vocabulary entry in each given document,
+/// it computes the termfrequecy times the inverse document frequency, thus giving more importance
+/// to entries that appear many times but only on some documents. The weight function can be adjusted
+/// by setting he appropriate [method](enum.TfIdfMethod.html). This struct provides the same string
```
bytesnake, line 36: typo `the`

bytesnake (commented):
> Can I get an initial review on this? Below are some sparse thoughts on the PR/possible additions/improvements
> 

looking good :+1:

>     * There are for sure some performance improvements possible in the count vectorizer but I want to write some benchmarks before diving into that
> 
>     * I see that sklearn allows for a normalization method to be defined for the count vectorizer, but here I think it would be less confusing to just use the count vectorizer as is and then applying the NormScaler to it (might be worthy of an example)
> 

this probably avoids duplicates

>     * sklearn uses the `20news-bydate` dataset to test/bench the vectorizer but that is a 15 MB compressed dataset, I think it might be too big to just include it in the dataset subcrate? Related to this, would It be useful to add a `fit_files` function to the vectorizer to pass in filenames instead of documents?
> 

we could perhaps add a benchmark which downloads this files, but I don't know how easy that is. Related to this: we could also add a `fit_reader(input: R) where R: Read` which can then directly be used with `R = File`

algorithms/linfa-preprocessing/src/count_vectorization.rs
```diff
@@ -0,0 +1,491 @@
...
+        let mut vocabulary: HashMap<String, (usize, usize)> = HashMap::new();
+        for string in x
+            .iter()
+            .map(|s| transform_string(s.to_string(), &self.properties))
+        {
+
+            let mut document_vocabulary: HashSet<String> = HashSet::new();
+            let words = regex.find_iter(&string).map(|mat| mat.as_str()).collect();
+            let queue = NGramQueue::new(words, self.properties.n_gram_range);
+            for ngram_items in queue {
```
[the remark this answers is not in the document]
Sauro98: quite shorter indeed 😄

algorithms/linfa-preprocessing/src/count_vectorization.rs
```diff
@@ -0,0 +1,491 @@
...
+                for item in ngram_items {
+                    // if item was already in the hashet it simply gets overwritten,
+                    // not a problem
+                    document_vocabulary.insert(item);
+                }
+            }
+            for word in document_vocabulary {
+                let len = vocabulary.len();
+                // If vocabulary item was already present then increase its document frequency
+                if let Some((_, freq)) = vocabulary.get_mut(&word) {
```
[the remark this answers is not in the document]
Sauro98: `document_vocabulary` is a hash-set but `vocabulary` is a hash-map. `document_vocabulary` is different for each document, so a vocabulary item may be already in `vocabulary` because it appeared in a previous document

algorithms/linfa-preprocessing/src/count_vectorization.rs
```diff
@@ -0,0 +1,491 @@
...
+                    document_frequencies[i] += 1;
+                }
+            }
+        }
+        (vectorized, document_frequencies)
+    }
+}
+
+fn validate_properties(properties: &VectorizerProperties) -> Result<Regex> {
+    let (n_gram_min, n_gram_max) = properties.n_gram_range;
```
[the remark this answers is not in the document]
Sauro98: In line 253 I call `Regex::new(&properties.split_regex)?` which I believe should do that?

algorithms/linfa-preprocessing/src/linear_scaling.rs
Sauro98: So like:
```rust
if self.method == ScalingMethod::Standard(false, true) {
    col.mapv_inplace(|el| (el - offset) * scale + offset);
} else {
    col.mapv_inplace(|el| (el - offset) * scale);
}
```
with `offset` being the mean and not zero so that, in that case, the features will be scaled but still around the same mean?

algorithms/linfa-preprocessing/README.md
```diff
@@ -0,0 +1,18 @@
...
+`linfa-preprocessing` is a crate in the [`linfa`](https://crates.io/crates/linfa) ecosystem, an effort to create a toolkit for classical Machine Learning implemented in pure Rust, akin to Python's `scikit-learn`.
+
+## Current state
+`linfa-preprocessing` provides a pure Rust implementation of:
+* Standard scaling 
+* Min-max scaling
+* Max Abs Scaling
+* Normalization (l1, l2 and max norm)
+* Count vectorization
+
```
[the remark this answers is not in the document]
Sauro98: I will look into it 👍🏻

algorithms/linfa-preprocessing/src/count_vectorization.rs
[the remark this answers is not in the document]
bytesnake: you're right, haven't seen the outer loop

algorithms/linfa-preprocessing/src/count_vectorization.rs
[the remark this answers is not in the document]
bytesnake: yep :+1:

algorithms/linfa-preprocessing/src/linear_scaling.rs
bytesnake: yes

algorithms/linfa-preprocessing/src/count_vectorization.rs
```diff
@@ -0,0 +1,491 @@
...
+    /// Constains all vocabulary entries, in the same order used by the `transform` method.
+    pub fn vocabulary(&self) -> &Vec<String> {
+        &self.vec_vocabulary
+    }
+
+    pub(crate) fn get_term_and_document_frequencies<T: ToString, D: Data<Elem = T>>(
+        &self,
+        x: &ArrayBase<D, Ix1>,
+    ) -> (Array2<usize>, Array1<usize>) {
+        let mut vectorized = Array2::zeros((x.len(), self.vocabulary.len()));
```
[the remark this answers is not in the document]
Sauro98: Yes, that makes sense, as soon as I have some benchmarks ready I will look into that. Perhaps it could be behind a flag? Or maybe we should just say: if you don't want it as a sparse matrix just call the related `sprs` method?

Sauro98:
Thanks for the review 😄 

> we could perhaps add a benchmark which downloads this files, but I don't know how easy that is. Related to this: we could also add a fit_reader(input: R) where R: Read which can then directly be used with R = File

I like this suggestion, I will play with it and get back to you when I have something which works

Sauro98:
The solution I found for downloading the `20news` dataset before doing the benchmarks was to copy the code of the `iai::main` macro in the benchmark file and add the call to the download function inside the main before initializing the benches. This feels very hacky but it seems to be working since the benchmarks show virtually no changes between benching with the files already downloaded and benching + downloading (~0.09% max diff)

Sauro98:
Also, I opted for `fit_files` instead of `fit_reader` for two reasons:

- The `20news` dataset has too many files inside to open all of them at the same time, making it impossible to have a reader for every file
- Since it expects files, it can directly address the possible encodings of the files (`20news` is encoded in ISO 8859-1 and not utf8)

Sauro98:
@bytesnake whitening with PCA would just mean setting the params to `embedding_size = n_features` and `apply_whitening = true` right?

bytesnake:
> @bytesnake whitening with PCA would just mean setting the params to `embedding_size = n_features` and `apply_whitening = true` right?

yes, though it would make sense to use complete SVD because I think the lobpcg may show worse convergence behaviour for large embedding_size, n_features ratios. You can apply whitening similar to https://github.com/rust-ml/linfa/blob/master/algorithms/linfa-reduction/src/pca.rs#L80 and avoid dependency to `linfa-reduction` as well

Sauro98:
Sorry for the long list in the main message. I've been using it as my checklist for this pr 😅 

I think this is finally ready for review. `iai` seems to be still in a very early stage so it has its limitations:
* I have to run the vectorizer benches with criterion otherwise it somehow messes up with `iai` and the results of the other benches end up having +/- infinity complexity increase with unchanged code (i believe it may be caused by the dataset download which is async? criterion, however, does not give any problem of this kind)
* it's not yet able to run specific benches based on the name. What I ended up doing was putting an "iai" prefix on all `iai` benches so that criterion can exclude itself when calling `cargo bench iai`. This last command is the only thing I'm doing rn on the benchmarks' continuous integration

algorithms/linfa-preprocessing/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+    #[error(transparent)]
+    IoError(#[from] std::io::Error),
+    #[error("Encoding error {0}")]
+    EncodingError(std::borrow::Cow<'static, str>),
+    #[error(transparent)]
+    LinalgError(#[from] ndarray_linalg::error::LinalgError),
+    #[error(transparent)]
+    NdarrayLinalgEmptyError(#[from] ndarray_stats::errors::EmptyInput),
+    #[error("Eigenvectors were not computed")]
+    NoneEignenvectors,
```
bytesnake, line 36: typo `NoneEigenvector`. Does this error occurs when the decomposition fails?

algorithms/linfa-preprocessing/src/linear_scaling.rs
```diff
@@ -0,0 +1,571 @@
...
+        }
+        let mut x = x;
+        Zip::from(x.gencolumns_mut())
+            .and(self.offsets())
+            .and(self.scales())
+            .apply(|mut col, &offset, &scale| {
+                if let ScalingMethod::Standard(false, true) = self.method {
+                    col.mapv_inplace(|el| (el - offset) * scale + offset);
+                } else {
+                    col.mapv_inplace(|el| (el - offset) * scale);
```
bytesnake, line 262: what happens for `(false, false)`?

algorithms/linfa-preprocessing/src/count_vectorization.rs
```diff
@@ -0,0 +1,806 @@
...
+                normalize: true,
+                document_frequency: (0., 1.),
+                stopwords: None,
+            },
+        }
+    }
+}
+
+impl CountVectorizer {
+    pub fn convert_to_lowercase(mut self, convert_to_lowercase: bool) -> Self {
```
bytesnake, line 65: can you document these functions as well

algorithms/linfa-preprocessing/src/whitening.rs
```diff
@@ -0,0 +1,304 @@
...
+        let sigma = x.records() - &mean;
+        match self.method {
+            WhiteningMethod::Pca => {
+                let (_, s, v_t) = sigma.svd(false, true)?;
+                if v_t.is_none() {
+                    return Err(Error::NoneEignenvectors);
+                }
+                let mut v_t = v_t.unwrap();
+                let s = s.mapv(|x| F::from(x).unwrap().max(F::from(1e-8).unwrap()));
+                let cov_scale = (F::from(x.nsamples()).unwrap() - F::one()).sqrt();
```
bytesnake, line 78: `F::from(x.nsamples() - 1).unwrap()`

bytesnake (approved):

bytesnake:
> Sorry for the long list in the main message. I've been using it as my checklist for this pr sweat_smile
> 

the first time I saw these 35 items in the issues I just thought, wut :smile: but the PR is just large.

> I think this is finally ready for review. `iai` seems to be still in a very early stage so it has its limitations:
> 

Yes I know, it is also still unclear whether it will be integrated into criterion at some point

>     * I have to run the vectorizer benches with criterion otherwise it somehow messes up with `iai` and the results of the other benches end up having +/- infinity complexity increase with unchanged code (i believe it may be caused by the dataset download which is async? criterion, however, does not give any problem of this kind)
> 

I have added a comment for the async part, perhaps it would be better to use `curl` there. This has two advantages (1) not have to use blocking async (2) reduce dependency tree

>     * it's not yet able to run specific benches based on the name. What I ended up doing was putting an "iai" prefix on all `iai` benches so that criterion can exclude itself when calling `cargo bench iai`. This last command is the only thing I'm doing rn on the benchmarks' continuous integration

sounds good

I think the PR is almost finished, I added one comment for the `async` and some missing docs for setter functions in `CountVectorizer` and `TfIdfNormalization`. The scope of this PR is huge, awesome work :+1:

algorithms/linfa-preprocessing/src/error.rs
Sauro98: In theory it should never happen. The svd function takes two booleans to know whether to build `U` and/or `V_T` in output, so these two matrices are given as `Option`. That error would only appear if someone set the needed flag to false, getting `None` for the matrix needed. So this kind of error can only appear to a developer modifying the whitening procedure. 

When decomposition fails I just propagate the `ndarray-linalg` error

algorithms/linfa-preprocessing/src/linear_scaling.rs
Sauro98: Good catch 👍🏻 
Since the stds are set to 1 when `with_scaling` is set to false, I can just change the first condition to `(false, _)`. I added a test case for this

algorithms/linfa-preprocessing/examples/count_vectorization.rs
```diff
@@ -0,0 +1,166 @@
...
+use linfa_bayes::GaussianNbParams;
+use linfa_preprocessing::count_vectorization::CountVectorizer;
+use ndarray::Array1;
+use std::collections::HashSet;
+use std::path::Path;
+use tar::Archive;
+
+async fn download_20news_bydate() {
+    let target = "http://qwone.com/~jason/20Newsgroups/20news-bydate.tar.gz";
+    let response = reqwest::get(target).await.unwrap();
```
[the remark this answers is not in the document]
Sauro98: It *is* huge 😅, the switch to curl seems to be working, I'll try to switch the benches back to `iai`

algorithms/linfa-preprocessing/src/error.rs
bytesnake: I would remove the error type and just unwrap the result with a comment explaining why this is safe

Sauro98:
Btw the `curl` crate worked out well, now without `reqwest`, `tokio` and `criterion` it has ~100 dependencies less

PR #94, opened 2021-03-13
[WIP] Add dataframes with polars

bytesnake:
This is WIP


datasets/src/lib.rs
```diff
@@ -58,6 +67,20 @@ fn array_from_buf(buf: &[u8]) -> Array2<f64> {
...
+#[cfg(feature = "income")]
+fn dataframe_from_buf(buf: &[u8]) -> DataFrame {
+    let mut file = GzDecoder::new(buf);
+    let mut buf: Vec<u8> = Vec::new();
+    file.read_to_end(&mut buf).unwrap();
+    let buf = Cursor::new(buf);
+
+    CsvReader::new(buf)
+        .infer_schema(None)
+        .has_header(false)
```
bernardo-sb, line 79: Shouldn't this option be give as a function argument `has_header: bool`?

DataPsycho: Hi, Is not it better to have direct arrow support than providing support for polars. So that any dataframe library based on arrow2 can use the preprocessing, and training methods? I am new to Rust and lifa project my question could be stupid. Thanks in advance for any clarity.

Eveeifyeve: > Hi, Is not it better to have direct arrow support than providing support for polars. So that any dataframe library based on arrow2 can use the preprocessing, and training methods? I am new to Rust and lifa project my question could be stupid. Thanks in advance for any clarity.

But there is so many unique implementations of arrow2 I feel the api should be developed so you can byo arrow2 implementation, this may already an issue, if not I thought I would mention it and maybe I could come up with a poc pr.

Issue #99, opened 2021-03-15
DBSCAN does not count border points


YuhanLiin:
The DBSCAN implementation in linfa-clustering only includes core points with minPts neighbours. Border points (points with less than minPts neighbours) connected to core points are counted as noise, which is different behaviour from how DBSCAN is typically described. Is this behaviour intentional or a bug?

Sauro98:
Hi, nice find, I don't believe that they should be counted as noise. In theory, border points could belong to more than one cluster but, since we can only return a single cluster membership value for each point, I believe that we should arbitrarily choose a single cluster for each border point like the approximate dbscan implementation does. Would you like to open a PR fixing this? This should be fixable with like three lines of code I think 👍🏻

YuhanLiin:
Might also be an issue with appx_dbscan as well, since those tests are failing after my fix.

bytesnake:
resolved in #100

PR #100, opened 2021-03-16
Include border points in DBSCAN + perf fixes

YuhanLiin:
Fixed border point inclusion in dbscan and appx_dbscan and added tests to verify the behaviour (addresses #99). Also fixed a bug where dbscan search queues contain duplicates values, causing significant perf issues.

### DIFF
--- a/algorithms/linfa-clustering/src/dbscan/algorithm.rs
+++ b/algorithms/linfa-clustering/src/dbscan/algorithm.rs
@@ -102,8 +102,9 @@ impl<F: Float, D: Data<Elem = F>> Transformer<&ArrayBase<D, Ix2>, Array1<Option<
                     self.tolerance(),
                     &cluster_memberships,
                 );
+                // Make the candidate a part of the cluster even if it's not a core point
+                cluster_memberships[candidate.0] = Some(current_cluster_id);
                 if neighbor_count >= self.minimum_points() {
-                    cluster_memberships[candidate.0] = Some(current_cluster_id);
                     search_queue.append(&mut neighbors);
                 }
             }
--- a/algorithms/linfa-clustering/src/dbscan/algorithm.rs
+++ b/algorithms/linfa-clustering/src/dbscan/algorithm.rs
@@ -1,5 +1,5 @@
 use crate::dbscan::hyperparameters::{DbscanHyperParams, DbscanHyperParamsBuilder};
-use ndarray::{Array1, ArrayBase, ArrayView, Axis, Data, Ix1, Ix2};
+use ndarray::{Array1, ArrayBase, Axis, Data, Ix2};
 use ndarray_stats::DeviationExt;
 
 use linfa::traits::Transformer;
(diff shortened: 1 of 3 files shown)


Sauro98 (approved):
Good work 👍🏻 

* Corrects border points bugs in dbscan and appx_dbscan
* Improves dbscan performance on the benches significantly (at least on my machine) while still passing all previous tests
* Adds tests for border points

For me this is good to go, I don't know whether @bytesnake would like to review it too.

algorithms/linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -79,32 +79,43 @@ impl<F: Float, D: Data<Elem = F>> Transformer<&ArrayBase<D, Ix2>, Array1<Option<
...
                 continue;
             }
             let (neighbor_count, mut search_queue) =
-                find_neighbors(&obs, observations, self.tolerance(), &cluster_memberships);
+                find_neighbors(i, observations, self.tolerance(), &cluster_memberships);
             if neighbor_count < self.minimum_points() {
                 continue;
             }
+            // Tracks whether a value is in the search queue to prevent duplicates
+            let mut search_found = vec![false; observations.nrows()];
```
bytesnake, line 94: doesn't this add a memory complexity of O(n^2) to the algorithm by allocating a bitmask not only for the neighborhood but all points. I'm not really familiar with the algorithm though

bytesnake (commented):
I

algorithms/linfa-clustering/src/dbscan/algorithm.rs
YuhanLiin: The `search_found` bitmask is instantiated once for each observation, but only one instance exists at a time because it gets dropped at the end of each iteration. The size of `search_found` is equal to the number of observations/points, which is what n is equal to.

bytesnake: yes sure but you still do N^2 allocations there, you could just re-use the same vec in every pass by moving it out of the loop

bytesnake:
awesome! We can also use a `VecDeque` as FIFO here such that `let candidate_idx = search_queue.remove(0);` is not so memory intensive. I simplified the tree search a bit and came up with this:
```rust
    fn transform(&self, observations: &ArrayBase<D, Ix2>) -> Array1<Option<usize>> {
        let mut cluster_memberships = Array1::from_elem(observations.nrows(), None);
        let mut current_cluster_id = 0;

        // tracks whether a value is in the search queue to prevent duplicates
        let mut search_inqueue = vec![false; observations.nrows()];
        let mut search_queue = VecDeque::with_capacity(observations.nrows());

        'outer: for i in 0..observations.nrows() {
            if cluster_memberships[i].is_some() {
                continue;
            }

            // add observation to queue
            search_queue.push_back(i);

            'inner: while !search_queue.is_empty() {
                let candidate_idx = search_queue.pop_front().unwrap();
                search_inqueue[candidate_idx] = false;

                let (neighbor_count, neighbors) = find_neighbors(
                    candidate_idx,
                    observations,
                    self.tolerance(),
                    &cluster_memberships,
                );

                // if our anchor's neighbourhood is too small, then just skip the pass of the
                // outer loop, this prevents that `current_cluster_id` is incremented and the
                // membership set
                if neighbor_count < self.minimum_points() && i == candidate_idx {
                    continue 'outer;
                }

                // make the candidate a part of the cluster even if it's not a core point
                cluster_memberships[candidate_idx] = Some(current_cluster_id);

                // only add neighbourhood to queue if its sufficient large
                if neighbor_count < self.minimum_points() {
                    continue 'inner;
                }

                for n in neighbors.into_iter() {
                    if !search_inqueue[n] {
                        search_queue.push_back(n);
                        search_inqueue[n] = true;
                    }
                }
            }

            current_cluster_id += 1;
        }
        cluster_memberships
    }
```

YuhanLiin:
Nice catch on the `.remove(0)`. I've change the queue to use VecDeque. The code structure you proposed does the same steps as what's in the PR right now. I personally prefer the current structure, since it's closer to the DBSCAN algorithm described on Wikipedia.

bytesnake:
good point :crab:

PR #101, opened 2021-03-18
Add Barnes-Hut implementation of t-SNE

bytesnake:
This PR adds the t-SNE algorithm by wrapping the [bhtsne](https://github.com/frjnn/bhtsne) crate.


algorithms/linfa-tsne/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+It wraps the [bhtsne](https://github.com/frjnn/bhtsne) crate, all kudos to them.
+
+## Examples
+
+There is an usage example in the `examples/` directory. To run, use:
+
+```bash
+$ cargo run --example tsne --features linfa/intel-mkl-system
+```
```
Sauro98, line 24: I see that some crates include this feature and some not, probably depending on who wrote the command. I don't know whether it is something worth of a note (like: see the list of openblas backends in [link to linfa main page]) or if it is something so small that that we can just leave it like it is

algorithms/linfa-tsne/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+    /// # Defaults to:
+    ///  * `theta`: 0.5
+    ///  * `perplexity`: 1.0
+    ///  * `max_iter`: 2000
+    pub fn embedding_size_with_rng(embedding_size: usize, rng: R) -> TSne<F, R> {
+        TSne {
+            embedding_size,
+            rng,
+            threshold: F::from(0.5).unwrap(),
+            perplexity: F::from(5.0).unwrap(),
```
Sauro98, line 72: Perplexity is said to be 1.0 in the comments but is 5.0 in the code. Also applies to the comment for `embedding_size`

Sauro98: Also, I don't know if it would be a good thing to specify that `theta` here refers to the `threshold` param (which is the name the user needs to use to change its value), mainly because I'm not really familiar with the algorithm so I don't know if they are used interchangeably.

Sauro98 (approved):
Nice work 👍🏻 

I only added some comments on typos and default values in the documentation; for the rest, everything looks good to me 👍🏻

algorithms/linfa-tsne/README.md
bytesnake: the example whiten the data prior to the embedding and requires therefore a LAPACK backend. I added a small note to look at the README front page with a link, but this happens at multiple places.

algorithms/linfa-tsne/src/lib.rs
bytesnake: the parameter is called theta in the paper, but I have called it `approx_threshold` here because it's the ratio of cell size to distance at which the approximation is used.

PR #105, opened 2021-03-19
Use m_k-means to avoid empty cluster problems

YuhanLiin:
Modify the centroid update step so the old centroid is essentially treated as an extra point in the cluster. Avoids most cases of empty clusters and prevents NaNs from appearing in the centroids.


bytesnake (approved):
thanks!

PR #106, opened 2021-03-20
Use more clusters in Kmeans benchmarks

YuhanLiin:
Increased number of clusters in Kmeans benchmarks. This will help gain insight into the performance of K-means with higher number of clusters, which will be helpful for evaluating future changes to Kmeans.


bytesnake (approved):

bytesnake:
Do you have the PR in draft mode intentionally?

YuhanLiin:
Yes. I'm planning to change the benchmarks to use Normal distribution instead of Uniform

YuhanLiin:
Just realized that the benchmarks actually use Normal distributions to generate blobs, so this is ready to merge as-is

Issue #109, opened 2021-03-22
Update to ndarray 0.14


relf:
Just to mention I've started to look at using ndarray 0.14 (and hopefully I intend to make a PR in the end)

For now, I am blocked by an issue with `sprs` which is used by `linfa-kernel`: https://github.com/vbarrielle/sprs/issues/278

bytesnake:
thanks for tackling this, last time I looked `ndarray-stats` still used the old `ndarray` version

bytesnake:
but it isn't anymore, version 0.4.0 is already published on `crates.io`!

relf:
ah yes! :facepalm: what did I check?

quietlychris:
@relf As of two days ago (March 25) it actually looks like `ndarray` has bumped its version up from `0.14` to `0.15`. I'm not sure how many breaking chances are actually between the two, but I think if you're planning on updating the version up from `0.13` anyway, it might make sense to jump all the way up (assuming that there's no major issues in the `0.15.0`).

relf:
Yep, I haven't dive in but it seems there is more changes from `0.14` to  `0.15` than from  `0.13` to `0.14`. 

That said, I think `linfa` should have at least a minor release for each minor release of `ndarray` (here we could have `linfa 0.4` with `ndarray 0.14`) because upgrading `ndarray` is not that easy due to the dependencies.

I do not know how cargo handles several versions of the same crate but I experienced compilation problems when there are various versions of `ndarray` drawn by various dependencies. And finally I waited that all dependencies upgraded to `ndarray 0.14` before making this PR.

Unfortunately, the `sprs` issue is blocking right now, and I do not know what is the best course of actions. I picked the revision vbarrielle/sprs@761d4f0f877de846b41b6e410822ce521ed008e8 where `sprs` upgraded to `ndarray 0.14` to make this PR compile. Should we wait for the issue resolution (which would be the same with `ndarray 0.15` anyway) or could we publish a `linfa 0.4` based on this?

relf:
fixed by #110

PR #110, opened 2021-03-23
Update ndarray 0.14

relf:
This PR update linfa to ndarray 0.14 version with the following crates updates as well
* approx = 0.4
* argmin = 0.4
* ndarray-csv = 0.5
* ndarray-linalg = 0.13
* ndarray-npy = 0.7
* ndarray-rand = 0.13
* ndarray-stats = 0.4
* num-traits = 0.2
* rand = 0.8
* rand-isaac = 0.3
* sprs = git master rev=761d4f0

The PR is still WIP as there is an issue with sprs 0.10 (see #109 )

### DIFF
--- a/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
+++ b/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
@@ -466,6 +466,7 @@ mod tests {
     use super::*;
     use crate::generate_blobs;
     use approx::{abs_diff_eq, assert_abs_diff_eq};
+    use lax::error::Error;
     use ndarray::{array, concatenate, ArrayView1, ArrayView2, Axis};
     use ndarray_linalg::error::LinalgError;
     use ndarray_linalg::error::Result as LAResult;
@@ -571,7 +572,8 @@ mod tests {
         assert!(
             match gmm.expect_err("should generate an error with reg_covar being nul") {
                 GmmError::LinalgError(e) => match e {
-                    LinalgError::Lapack(_) => true,
+                    LinalgError::Lapack(Error::LapackComputationalFailure { return_code: 2 }) =>
+                        true,
                     _ => panic!("should be a lapack error 2"),
                 },
                 _ => panic!("should be a linear algebra error"),
@@ -599,7 +601,8 @@ mod tests {
         assert!(
             match gmm.expect_err("should generate an error with reg_covar being nul") {
                 GmmError::LinalgError(e) => match e {
-                    LinalgError::Lapack(_) => true,
+                    LinalgError::Lapack(Error::LapackComputationalFailure { return_code: 1 }) =>
+                        true,
                     _ => panic!("should be a lapack error 1"),
                 },
                 _ => panic!("should be a linear algebra error"),
--- a/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
+++ b/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
@@ -296,7 +296,7 @@ impl<F: Float + Lapack + Scalar> GaussianMixtureModel<F> {
         let n_samples = observations.nrows();
         let (weights, means, covariances) = Self::estimate_gaussian_parameters(
             &observations,
-            &log_resp.mapv(|v| v.exp()),
+            &log_resp.mapv(|v| Scalar::exp(v)),
             &self.covar_type,
             reg_covar,
         )?;
@@ -325,9 +325,9 @@ impl<F: Float + Lapack + Scalar> GaussianMixtureModel<F> {
     ) -> (Array1<F>, Array2<F>) {
         let weighted_log_prob = self.estimate_weighted_log_prob(&observations);
         let log_prob_norm = weighted_log_prob
-            .mapv(|v| v.exp())
+            .mapv(|v| Scalar::exp(v))
             .sum_axis(Axis(1))
-            .mapv(|v| v.ln());
+            .mapv(|v| Scalar::ln(v));
         let log_resp = weighted_log_prob - log_prob_norm.to_owned().insert_axis(Axis(1));
         (log_prob_norm, log_resp)
     }
@@ -384,12 +384,12 @@ impl<F: Float + Lapack + Scalar> GaussianMixtureModel<F> {
             .unwrap()
             .slice(s![.., ..; n_features+1])
             .to_owned()
-            .mapv(|v| v.ln());
+            .mapv(|v| Scalar::ln(v));
         log_diags.sum_axis(Axis(1))
     }
 
     fn estimate_log_weights(&self) -> Array1<F> {
-        self.weights().mapv(|v| v.ln())
+        self.weights().mapv(|v| Scalar::ln(v))
     }
 }
 
@@ -420,7 +420,7 @@ impl<'a, F: Float + Lapack + Scalar, R: Rng + SeedableRng + Clone, D: Data<Elem 
                 lower_bound =
                     GaussianMixtureModel::<F>::compute_lower_bound(&log_resp, log_prob_norm);
                 let change = lower_bound - prev_lower_bound;
-                if change.abs() < self.tolerance() {
+                if ndarray_rand::rand_distr::num_traits::Float::abs(change) < self.tolerance() {
                     converged_iter = Some(n_iter);
                     break;
                 }
@@ -456,7 +456,7 @@ impl<F: Float + Lapack + Scalar, D: Data<Elem = F>> PredictRef<ArrayBase<D, Ix2>
     fn predict_ref<'a>(&'a self, observations: &ArrayBase<D, Ix2>) -> Array1<usize> {
         let (_, log_resp) = self.estimate_log_prob_resp(&observations);
         log_resp
-            .mapv(|v| v.exp())
+            .mapv(|v| Scalar::exp(v))
             .map_axis(Axis(1), |row| row.argmax().unwrap())
     }
 }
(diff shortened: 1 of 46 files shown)


algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -293,7 +293,7 @@ impl<F: Float + Lapack + Scalar> GaussianMixtureModel<F> {
         let n_samples = observations.nrows();
         let (weights, means, covariances) = Self::estimate_gaussian_parameters(
             &observations,
-            &log_resp.mapv(|v| v.exp()),
+            &log_resp.mapv(|v| Scalar::exp(v)),
```
bytesnake, line 296: can you import the trait and use the original order? There are several places where this becomes really long

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -596,7 +596,7 @@ mod tests {
         assert!(
             match gmm.expect_err("should generate an error with reg_covar being nul") {
                 GmmError::LinalgError(e) => match e {
-                    LinalgError::Lapack { return_code: 1 } => true,
+                    LinalgError::Lapack(_) => true,
```
bytesnake, line 599: you have to check of the `return_code` here

algorithms/linfa-clustering/src/k_means/algorithm.rs
```diff
@@ -365,7 +365,7 @@ fn get_random_centroids<F: Float, D: Data<Elem = F>>(
 mod tests {
     use super::*;
     use approx::assert_abs_diff_eq;
-    use ndarray::{array, stack, Array, Array1, Array2, Axis};
+    use ndarray::{array, concatenate, Array, Array1, Array2, Axis};
```
bytesnake, line 368: `concatenate` is really long, I don't know why they didn't go for the abbreviation `concat`

algorithms/linfa-kernel/Cargo.toml
```diff
@@ -24,8 +24,9 @@ default-features = false
 features = ["std", "derive"]
 
 [dependencies]
-ndarray = "0.13"
-sprs = { version = "0.9.3", default-features = false }
+ndarray = "0.14"
+num-traits = "0.2"
+sprs = { git="https://github.com/vbarrielle/sprs.git", rev="761d4f0", default-features = false }
```
bytesnake, line 29: didn't know that you can specify the commit hash in your dependency tree :) to what revision does this point, I assume the `linfa-kernel` would not compile with the latest `sprs` release?

algorithms/linfa-linear/Cargo.toml
```diff
@@ -17,16 +17,16 @@ keywords = ["machine-learning", "linfa", "ai", "ml", "linear"]
 categories = ["algorithms", "mathematics", "science"]
 
 [dependencies]
-ndarray = {version = "0.13", features = ["blas", "approx"]}
-ndarray-linalg = "0.12"
-ndarray-stats = "0.3"
+ndarray = {version = "0.14", features = ["blas", "approx"]}
```
bytesnake, line 20: if you're correcting the spacing, do here as well. I actually don't know why rustfmt doesn't do that automatically

algorithms/linfa-reduction/src/pca.rs
```diff
@@ -293,7 +293,7 @@ mod tests {
             let mp_law = ((b - x) * (x - a)).sqrt() / std::f64::consts::PI / x;
             let empirical = count as f64 / 500. / ((2.8 - 0.1) / 28.);
 
-            assert_abs_diff_eq!(mp_law, empirical, epsilon = 0.05);
+            assert_abs_diff_eq!(mp_law, empirical, epsilon = 0.06);
```
bytesnake, line 296: this fails for you? The test should be deterministic

bytesnake (commented):
Thanks for the cleanup and dependency update. This won't work with the latest `sprs` release because of https://github.com/vbarrielle/sprs/issues/278 correct? The `0.14` -> `0.15` jump seems to be quite big (I just skimmed the release notes, there are probably some breakages for the parallel stuff in kmeans) and should be in a separate crate

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
relf: I am not sure to understand... What do you mean by the original order? Here if I use `v.exp()` I get the following compilation error:
```bash
   Compiling linfa-clustering v0.3.1 (D:\rlafage\workspace\linfa\algorithms\linfa-clustering)
error[E0034]: multiple applicable items in scope
    --> algorithms\linfa-clustering\src\gaussian_mixture\algorithm.rs:296:34
     |
296  |             &log_resp.mapv(|v| v.exp()),
     |                                  ^^^ multiple `exp` found
     |
note: candidate #1 is defined in the trait `ndarray_rand::rand_distr::num_traits::Float`
    --> D:\rlafage\.cargo\registry\src\github.com-1ecc6299db9ec823\num-traits-0.2.14\src\float.rs:1336:5
     |
1336 |     fn exp(self) -> Self;
     |     ^^^^^^^^^^^^^^^^^^^^^
note: candidate #2 is defined in the trait `Scalar`
    --> D:\rlafage\.cargo\registry\src\github.com-1ecc6299db9ec823\cauchy-0.3.0\src\lib.rs:112:5
     |
112  |     fn exp(self) -> Self;
     |     ^^^^^^^^^^^^^^^^^^^^^
help: disambiguate the associated function for candidate #1
     |
296  |             &log_resp.mapv(|v| ndarray_rand::rand_distr::num_traits::Float::exp(v)),
     |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: disambiguate the associated function for candidate #2
     |
296  |             &log_resp.mapv(|v| Scalar::exp(v)),
     |                                ^^^^^^^^^^^^^^
```

algorithms/linfa-kernel/Cargo.toml
relf: Yes, `linfa-kernel` can not compile with `sprs 0.10`, my question was: could we release linfa 0.4 using such spec or is it too bold? Sure I would prefer a `sprs 0.10.1` fixing the issue but atm it blocks the ndarray upgrade.

algorithms/linfa-reduction/src/pca.rs
relf: Yes I had to adjust the epsilon, I supposed some subtle computation changes in the dependencies? I will check what I get currently.

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
relf: You're right, I took a shortcut, Lapack error has changed, I will check

relf:
> This won't work with the latest sprs release because of vbarrielle/sprs#278 correct?

Exactly. I open the `sprs` issue after I get a compilation error with `linfa-kernel`.

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
relf: done

algorithms/linfa-linear/Cargo.toml
relf: done

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
bytesnake: ah I see, these are conflicting because the first exponential is real only and the second is complex. I was confused by the name `Scalar` representing the complex as well as real numbers. Any idea why this ambiguity appears now?

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
bytesnake: you can also click on `Resolve conversation`

algorithms/linfa-kernel/Cargo.toml
bytesnake: no that is not possible, because `crates.io` only allows self referencing and we would therefore not be able to publish the crate when using a foreign git source. We can't go forward without `sprs` fixing and publishing their crate.

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
relf: After a bit of digging, I found that `ndarray-linalg` undergoes the same phenomenon : https://github.com/rust-ndarray/ndarray-linalg/commit/5157b22b0d3c911da7ad25f93037cd24dcc476f1 which led me to compare cauchy 0.2 vs 0.3 https://github.com/rust-math/cauchy/compare/0.2.2...0.3.0 and it seems that operations like `exp()` were changed to take `self` (instead of `&self`) which could explain the ambiguity is popping up now.

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
relf: Yeah, I always tend to feel that the conversation should be resolved by the reviewer.

relf:
Hmmm... I do not understand what is going on. I am not familiar with the various CI checks. :eyes: It seems that build fails with rustc 1.42 which would be our current lowest rustc supported version, @bytesnake correct?

mulimoen:
Running `cargo tree --all -i ndarray` shows you are pulling in `ndarray:0.15.1` through
```
ndarray v0.15.1
└── ndarray-rand v0.14.0
    └── argmin v0.4.2
        ├── linfa-linear v0.3.1 (/home/magnus/Documents/projects/linfa/algorithms/linfa-linear)
        └── linfa-logistic v0.3.1 (/home/magnus/Documents/projects/linfa/algorithms/linfa-logistic)
```

relf:
@mulimoen thanks for the heads up. Indeed in the meantime, `argmin` bumped from 0.4.1 to 0.4.2 pulling new `ndarray_rand` version and then new `ndarray 0.15`! At the same time  `argmin 0.4.2` fixes the issue https://github.com/argmin-rs/argmin/issues/104 which broke the `linfa` CI system! I am doomed! :sweat_smile:

mulimoen:
@relf Ouch, that is quite unfortunate, you were almost there with the `sprs` bump

bytesnake (approved):
okay let's bump and I will try to clean up a bit in a follow up :+1:

PR #115, opened 2021-04-03
Least Angle Regression

stokhos:
This is a dummy PR.

**TODO**


bytesnake:
you can find the original paper here: https://web.stanford.edu/~hastie/Papers/LARS/LeastAngle_2002.pdf there is also a section in the [ELSII](https://web.stanford.edu/~hastie/Papers/ESLII.pdf) book on section 3.4.4 (p. 92) about LAR

stokhos:
I have finished reading the paper. 

I did some search on LARS, looks like there are other variants. Like Robust LARS, and LARS with LASSO modification.

For now, I'll just focus on LARS. And then I will try to implement others.

PR #118, opened 2021-04-06
Add additional bounds to Float trait

YuhanLiin:
Adding these bounds allow trait bounds on `Float` in dependent crates to be simplified. These simplifications were made in `linfa-clustering`.

### DIFF
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         toolchain:
           - 1.42.0
-          - nightly-2021-03-26
+          - nightly-2021-03-24
           - beta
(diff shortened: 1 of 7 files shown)


bytesnake:
I will look into the issue with argmin first, before mergin this

bytesnake:
can confirm that this is a regression introduced by the latest nightly compiler version, can you pin the toolchain to `nightly-2021-03-26` in `.github/workflows/checking.yml` for the time being?

bytesnake:
you need to do the same for `.github/workflows/codequality.yml`

bytesnake:
sorry I'm a bit busy right now working through my backlog :sweat_smile:

PR #121, opened 2021-04-20
NdArray bump follow up

bytesnake:
This is the followup to the PR #110. It tries to simplify our story of [cauchy::Scalar](https://docs.rs/cauchy/0.4.0/src/cauchy/lib.rs.html#51-130) and [num_trait::Float](https://docs.rs/num-traits/0.2.14/src/num_traits/float.rs.html#905-1809) by moving the trait bound of cauchy behind an associated type and avoiding thus the ambiguity when mixing both bounds.

It adds a function [Float::map_cauchy_bound](https://github.com/rust-ml/linfa/pull/121/files#diff-52d42126f7ad584e11c25ac3974c4cce4e290ff3b8a900e46bb87bca6829f2c3R72) and pulls in the trait bound here for an example in `linfa-clustering`: https://github.com/rust-ml/linfa/pull/121/files#diff-cc618ed2646ccdd41e994796498d4b9ed256dc0d99c1976302eba1c496caaf64R256. 

I also added a `F::cast` function to shorten number casting and avoid [num_traits::FromPrimitive](https://docs.rs/num-traits/0.2.14/num_traits/cast/trait.FromPrimitive.html) :sweat_smile: 

@relf

### DIFF
--- a/src/dataset/mod.rs
+++ b/src/dataset/mod.rs
@@ -29,9 +29,7 @@ mod impl_targets;
 mod iter;
 pub mod multi_target_model;
 
-#[cfg(feature = "ndarray-linalg")]
 mod lapack_bounds;
-#[cfg(feature = "ndarray-linalg")]
 pub use lapack_bounds::*;
 
 /// Floating point numbers
@@ -63,6 +61,8 @@ pub trait Float:
 {
     #[cfg(feature = "ndarray-linalg")]
     type Lapack: Float + Scalar + Lapack;
+    #[cfg(not(feature = "ndarray-linalg"))]
+    type Lapack: Float;
 
     fn cast<T: NumCast>(x: T) -> Self {
         NumCast::from(x).unwrap()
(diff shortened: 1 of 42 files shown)


src/dataset/mod.rs
```diff
@@ -43,10 +50,113 @@ pub trait Float:
...
     + for<'a> DivAssign<&'a Self>
     + SampleUniform
+    + ScalarOperand
 {
+    type F2: Scalar + Lapack + Float;
+
+    fn self_to_float(self) -> Self::F2;
+    fn float_to_self(x: Self::F2) -> Self;
+
+    fn ndarray_to_float<I: Dimension>(x: ArrayBase<OwnedRepr<Self>, I>) -> ArrayBase<OwnedRepr<Self::F2>, I>;
```
bytesnake, line 60: oh boy

src/dataset/mod.rs
```diff
@@ -43,10 +50,113 @@ pub trait Float:
     + for<'a> SubAssign<&'a Self>
     + for<'a> DivAssign<&'a Self>
     + SampleUniform
+    + ScalarOperand
```
relf, line 53: Maybe we could add `num_traits::MulAdd<Output = F>` (see https://github.com/vbarrielle/sprs/issues/278#issuecomment-821520964) to ease the upgrade to `sprs 0.10`?

bytesnake: yes! Haven't seen the post

PR #122, opened 2021-04-21
Fit trait modification and cross validation proposal

Sauro98:
## Changes in the `Fit` trait

from this: 
```rust 
pub trait Fit<'a, R: Records, T> {
    type Object: 'a;

    fn fit(&self, dataset: &DatasetBase<R, T>) -> Self::Object;
}
```
to this:
```rust
pub trait Fit<R: Records, T, E: std::error::Error + std::convert::From<linfa::Error>> {
    type Object;

    fn fit(&self, dataset: &DatasetBase<R, T>) -> Result<Self::Object, E>;
}
```
by:
- removing `'a` lifetime (left from previous svm implementation, not actually used by any algorithm anymore)
- forcing every implementation to return a result with an error struct (every implementation except PCA already returned an error, some implementations returned a String as the error but the transition keeps the same error messages)

#### Edit 1:
Added conversion from linfa error bound on fit error type. Every sub-crate should be able to handle the errors caused by using the base crate

## Cross validation POC

Cross-validation can be defined by exploiting the new `Fit` definition. This is what it looks like for regression:

```rust
    use linfa::prelude::*;
    use linfa_elasticnet::{ElasticNet, Result};

    // load Diabetes dataset (mutable to allow fast k-folding)
    let mut dataset = linfa_datasets::diabetes();

    // prameters to compare
    let ratios = vec![0.1, 0.2, 0.5, 0.7, 1.0];

    // create a model for each parameter
    let models = ratios
        .iter()
        .map(|ratio| ElasticNet::params().penalty(0.3).l1_ratio(*ratio))
        .collect::<Vec<_>>();

    // get the mean r2 validation score across all folds for each model
    let r2_values =
        dataset.cross_validate(5, &models, |prediction, truth| prediction.r2(&truth))?;

    for (ratio, r2) in ratios.iter().zip(r2_values.iter()) {
        println!("L1 ratio: {}, r2 score: {}", ratio, r2);
    }
```

And this is what it looks like for classification:

```rust
    use linfa::prelude::*;
    use linfa_logistic::error::Result;
    use linfa_logistic::LogisticRegression;

    // Load dataset. Mutability is needed for fast cross validation
    let mut dataset =
        linfa_datasets::winequality().map_targets(|x| if *x > 6 { "good" } else { "bad" });

    // define a sequence of models to compare. In this case the
    // models will differ by the amount of l2 regularization
    let alphas = vec![0.1, 1., 10.];
    let models: Vec<_> = alphas
        .iter()
        .map(|alpha| {
            LogisticRegression::default()
                .alpha(*alpha)
                .max_iterations(150)
        })
        .collect();

    // use cross validation to compute the validation accuracy of each model. The
    // accuracy of each model will be averaged across the folds, 5 in this case
    let accuracies = dataset.cross_validate(5, &models, |prediction, truth| {
        Ok(prediction.confusion_matrix(truth)?.accuracy())
    })?;

    // display the accuracy of the models along with their regularization coefficient
    for (alpha, accuracy) in alphas.iter().zip(accuracies.iter()) {
        println!("Alpha: {}, accuracy: {} ", alpha, accuracy);
    }
```

## Possible follow up

Redefine the `Transformer` trait to return a `Result`, like the t-sne implementation does, in order to avoid panicking as much as possible. It would be better to wait for #121 in order to avoid dealing with the `unwrap()` calls when performing float conversions.

## Notes
- ~~Currently there are two versions of cross-validation: one for single target datasets and one for multi-target datasets. The main reasons for this are:~~
    - ~~`Array1`s of evaluation values can be constructed with `collect`, `Array2` cannot and must be populated row by row (could be avoided by stacking, or maybe there is a dedicated ndarray method I am not aware of)~~ 
    - ~~Regression metrics behave differently when they are applied `Array1-Array2` or `Array2-Array1`, making writing evaluation closures possibly more difficult than it should.~~

~~More than likely there is a solution to both problems, but I got stuck on it for too long and so I would consider it out of scope for this PR, unless someone has an easy solution to suggest. Forcing the user to give a single return value in the multi-target case (like a mean across the targets) could make the problem easier but the evaluation for each target would be lost in the process.~~

- Got this error in elasticnet just once:
```rust
     Running target/release/deps/linfa_elasticnet-d4c6cf99e56de5dd

running 11 tests
test algorithm::tests::coordinate_descent_lowers_objective ... ok
test algorithm::tests::elastic_net_2d_toy_example_works ... ok
test algorithm::tests::elastic_net_diabetes_1_works_like_sklearn ... ok
test algorithm::tests::diabetes_z_score ... ok
test algorithm::tests::elastic_net_penalty_works ... ok
test algorithm::tests::elastic_net_toy_example_works ... ok
test algorithm::tests::lasso_toy_example_works ... ok
test algorithm::tests::lasso_zero_works ... ok
error: test failed, to rerun pass '-p linfa-elasticnet --lib'

Caused by:
  process didn't exit successfully: `/home/ivano/Scrivania/github_projects/linfa/target/release/deps/linfa_elasticnet-d4c6cf99e56de5dd` (signal: 11, SIGSEGV: invalid memory reference)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ features = ["cblas"]
 default-features = false
 
 [dependencies.openblas-src]
-version = "0.9.0"
+version = "0.10.4"
 optional = true
 default-features = false
 features = ["cblas"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,6 @@ features = ["cblas"]
 
 [dev-dependencies]
 ndarray-rand = "0.13"
-
 linfa-datasets = { path = "datasets", features = ["winequality", "iris", "diabetes"] }
 
 [workspace]
--- a/algorithms/linfa-clustering/src/appx_dbscan/hyperparameters.rs
+++ b/algorithms/linfa-clustering/src/appx_dbscan/hyperparameters.rs
@@ -99,7 +99,7 @@ impl<F: Float> AppxDbscanHyperParams<F> {
     }
 
     fn build(tolerance: F, min_points: usize, slack: F) -> Self {
-        if tolerance <= F::cast(0.) {
+        if tolerance <= F::zero() {
             panic!("`tolerance` must be greater than 0!");
         }
         // There is always at least one neighbor to a point (itself)
@@ -107,13 +107,13 @@ impl<F: Float> AppxDbscanHyperParams<F> {
             panic!("`min_points` must be greater than 1!");
         }
 
-        if slack <= F::cast(0.) {
+        if slack <= F::zero() {
             panic!("`slack` must be greater than 0!");
         }
         Self {
-            tolerance: tolerance,
-            min_points: min_points,
-            slack: slack,
+            tolerance,
+            min_points,
+            slack,
             appx_tolerance: tolerance * (F::one() + slack),
         }
     }
--- a/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
+++ b/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
@@ -215,10 +215,10 @@ impl<F: Float> GaussianMixtureModel<F> {
         reg_covar: F,
     ) -> Result<(Array1<F>, Array2<F>, Array3<F>)> {
         let nk = resp.sum_axis(Axis(0));
-        if nk.min().unwrap() < &(F::cast(10.) * F::epsilon()) {
+        if nk.min()? < &(F::cast(10.) * F::epsilon()) {
             return Err(GmmError::EmptyCluster(format!(
                 "Cluster #{} has no more point. Consider decreasing number of clusters or change initialization.",
-                nk.argmin().unwrap() + 1
+                nk.argmin()? + 1
             )));
         }
 
@@ -400,12 +400,12 @@ impl<F: Float> GaussianMixtureModel<F> {
     }
 }
 
-impl<'a, F: Float, R: Rng + SeedableRng + Clone, D: Data<Elem = F>, T> Fit<'a, ArrayBase<D, Ix2>, T>
-    for GmmHyperParams<F, R>
+impl<F: Float, R: Rng + SeedableRng + Clone, D: Data<Elem = F>, T>
+    Fit<ArrayBase<D, Ix2>, T, GmmError> for GmmHyperParams<F, R>
 {
-    type Object = Result<GaussianMixtureModel<F>>;
+    type Object = GaussianMixtureModel<F>;
 
-    fn fit(&self, dataset: &DatasetBase<ArrayBase<D, Ix2>, T>) -> Self::Object {
+    fn fit(&self, dataset: &DatasetBase<ArrayBase<D, Ix2>, T>) -> Result<Self::Object> {
         self.validate()?;
         let observations = dataset.records().view();
         let mut gmm = GaussianMixtureModel::<F>::new(self, dataset, self.rng())?;
@@ -488,7 +488,7 @@ mod tests {
     }
     impl MultivariateNormal {
         pub fn new(mean: &ArrayView1<f64>, covariance: &ArrayView2<f64>) -> LAResult<Self> {
-            let lower = covariance.cholesky(UPLO::Lower).unwrap();
+            let lower = covariance.cholesky(UPLO::Lower)?;
             Ok(MultivariateNormal {
                 mean: mean.to_owned(),
                 covariance: covariance.to_owned(),
--- a/algorithms/linfa-logistic/src/lib.rs
+++ b/algorithms/linfa-logistic/src/lib.rs
@@ -14,7 +14,11 @@
 //! ```bash
 //! $ cargo run --example winequality
 //! ```
+//!
 
+pub mod error;
+
+use crate::error::{Error, Result};
 use argmin::prelude::*;
 use argmin::solver::linesearch::MoreThuenteLineSearch;
 use argmin::solver::quasinewton::lbfgs::LBFGS;
@@ -138,11 +142,7 @@ impl<F: Float> LogisticRegression<F> {
     /// i.e. any values are `Inf` or `NaN`, `y` doesn't have as many items as
     /// `x` has rows, or if other parameters (gradient_tolerance, alpha) have
     /// been set to inalid values.
-    fn fit<A, T, C>(
-        &self,
-        x: &ArrayBase<A, Ix2>,
-        y: T,
-    ) -> Result<FittedLogisticRegression<F, C>, String>
+    fn fit<A, T, C>(&self, x: &ArrayBase<A, Ix2>, y: T) -> Result<FittedLogisticRegression<F, C>>
     where
         A: Data<Elem = F>,
         T: AsTargets<Elem = C>,
@@ -159,45 +159,38 @@ impl<F: Float> LogisticRegression<F> {
 
     /// Ensure that `x` and `y` have the right shape and that all data and
     /// configuration parameters are finite.
-    fn validate_data<A, B>(
-        &self,
-        x: &ArrayBase<A, Ix2>,
-        y: &ArrayBase<B, Ix1>,
-    ) -> Result<(), String>
+    fn validate_data<A, B>(&self, x: &ArrayBase<A, Ix2>, y: &ArrayBase<B, Ix1>) -> Result<()>
     where
         A: Data<Elem = F>,
         B: Data<Elem = F>,
     {
         if x.shape()[0] != y.len() {
-            return Err(
-                "Incompatible shapes of data, expected `x` and `y` to have same number of rows"
-                    .to_string(),
-            );
+            return Err(Error::MismatchedShapes(x.shape()[0], y.len()));
         }
         if x.iter().any(|x| !x.is_finite())
             || y.iter().any(|y| !y.is_finite())
             || !self.alpha.is_finite()
         {
-            return Err("Values must be finite and not `Inf`, `-Inf` or `NaN`".to_string());
+            return Err(Error::InvalidValues);
         }
         if !self.gradient_tolerance.is_finite() || self.gradient_tolerance <= F::zero() {
-            return Err("gradient_tolerance must be a positive, finite number".to_string());
+            return Err(Error::InvalidGradientTolerance);
         }
         self.validate_init_params(x)?;
         Ok(())
     }
 
-    fn validate_init_params<A>(&self, x: &ArrayBase<A, Ix2>) -> Result<(), String>
+    fn validate_init_params<A>(&self, x: &ArrayBase<A, Ix2>) -> Result<()>
     where
         A: Data<Elem = F>,
     {
         if let Some((params, intercept)) = self.initial_params.as_ref() {
             let (_, n_features) = x.dim();
             if n_features != params.dim() {
-                return Err("Size of initial parameter guess must be the same as the number of columns in the feature matrix `x`".to_string());
+                return Err(Error::InvalidInitialParametersGuessSize);
             }
             if params.iter().any(|p| !p.is_finite()) || !intercept.is_finite() {
-                return Err("Initial parameter guess must be finite".to_string());
+                return Err(Error::InvalidInitialParametersGuess);
             }
         }
         Ok(())
@@ -254,14 +247,14 @@ impl<F: Float> LogisticRegression<F> {
         problem: LogisticRegressionProblem<'a, F, A>,
         solver: LBFGSType<F>,
         init_params: Array1<F>,
-    ) -> Result<ArgminResult<LogisticRegressionProblem<'a, F, A>>, String>
+    ) -> Result<ArgminResult<LogisticRegressionProblem<'a, F, A>>>
     where
         A: Data<Elem = F>,
     {
         Executor::new(problem, solver, ArgminParam(init_params))
             .max_iters(self.max_iterations)
             .run()
-            .map_err(|err| format!("Error running solver: {}", err))
+            .map_err(|err| err.into())
     }
 
     /// Take an ArgminResult and return a FittedLogisticRegression.
@@ -269,7 +262,7 @@ impl<F: Float> LogisticRegression<F> {
         &self,
         labels: ClassLabels<F, C>,
         result: &ArgminResult<LogisticRegressionProblem<F, A>>,
-    ) -> Result<FittedLogisticRegression<F, C>, String>
+    ) -> Result<FittedLogisticRegression<F, C>>
     where
         A: Data<Elem = F>,
         C: PartialOrd + Clone,
@@ -285,9 +278,9 @@ impl<F: Float> LogisticRegression<F> {
 }
 
 impl<'a, C: 'a + PartialOrd + Clone, F: Float, D: Data<Elem = F>, T: AsTargets<Elem = C>>
-    Fit<'a, ArrayBase<D, Ix2>, T> for LogisticRegression<F>
+    Fit<ArrayBase<D, Ix2>, T, Error> for LogisticRegression<F>
 {
-    type Object = Result<FittedLogisticRegression<F, C>, String>;
+    type Object = FittedLogisticRegression<F, C>;
 
     /// Given a 2-dimensional feature matrix array `x` with shape
     /// (n_samples, n_features) and an array of target classes to predict,
@@ -305,10 +298,7 @@ impl<'a, C: 'a + PartialOrd + Clone, F: Float, D: Data<Elem = F>, T: AsTargets<E
     /// i.e. any values are `Inf` or `NaN`, `y` doesn't have as many items as
     /// `x` has rows, or if other parameters (gradient_tolerance, alpha) have
     /// been set to inalid values.
-    fn fit(
-        &self,
-        dataset: &DatasetBase<ArrayBase<D, Ix2>, T>,
-    ) -> Result<FittedLogisticRegression<F, C>, String> {
+    fn fit(&self, dataset: &DatasetBase<ArrayBase<D, Ix2>, T>) -> Result<Self::Object> {
         self.fit(dataset.records(), dataset.targets())
     }
 }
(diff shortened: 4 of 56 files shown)


Sauro98:
@YuhanLiin I got this test error on kmeans just once and I haven't been able to reproduce it:
```rust
---- k_means::init::tests::test_sample_subsequent_candidates stdout ----
thread 'k_means::init::tests::test_sample_subsequent_candidates' panicked at 'assertion failed: `(left == right)`
  left: `3`,
 right: `2`', algorithms/linfa-clustering/src/k_means/init.rs:259:9
```

It seems like the test takes a fixed seed in input so it should be deterministic, right? Could there be some randomness in the rng seeds for each thread? 
Maybe it was just my machine acting up 😅, and also the error doesn't show up on the CI system, but I still tagged you in case you have some insight on why it might have happened

Cargo.toml
```diff
@@ -59,7 +59,7 @@ features = ["cblas"]
 [dev-dependencies]
 ndarray-rand = "0.13"
 approx = { version = "0.4", default-features = false, features = ["std"] }
-
+thiserror = "1"
```
bytesnake, line 62: `thiserror` is already in your dependencies

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -393,12 +393,12 @@ impl<F: Float + Lapack + Scalar> GaussianMixtureModel<F> {
     }
 }
 
-impl<'a, F: Float + Lapack + Scalar, R: Rng + SeedableRng + Clone, D: Data<Elem = F>, T>
-    Fit<'a, ArrayBase<D, Ix2>, T> for GmmHyperParams<F, R>
+impl<F: Float + Lapack + Scalar, R: Rng + SeedableRng + Clone, D: Data<Elem = F>, T>
+    Fit<ArrayBase<D, Ix2>, T, GmmError> for GmmHyperParams<F, R>
 {
-    type Object = Result<GaussianMixtureModel<F>>;
+    type Object = GaussianMixtureModel<F>;
```
bytesnake, line 399: this makes the associated type easier to read :+1:

algorithms/linfa-logistic/src/lib.rs
```diff
@@ -159,45 +159,38 @@ impl<F: Float> LogisticRegression<F> {
...
     where
         A: Data<Elem = F>,
         B: Data<Elem = F>,
     {
         if x.shape()[0] != y.len() {
-            return Err(
-                "Incompatible shapes of data, expected `x` and `y` to have same number of rows"
-                    .to_string(),
-            );
+            return Err(Error::MismatchedShapes);
```
bytesnake, line 168: you can also pass the number of rows and print that in your error, like "Expected same shape, got <..> ≠ <..>"

algorithms/linfa-tsne/examples/tsne.rs
```diff
@@ -5,7 +5,7 @@ use std::{io::Write, process::Command};
 
 fn main() -> Result<()> {
     let ds = linfa_datasets::iris();
-    let ds = Pca::params(3).whiten(true).fit(&ds).transform(ds);
+    let ds = Pca::params(3).whiten(true).fit(&ds).unwrap().transform(ds);
```
bytesnake, line 8: ~~unwrap~~ -> ?

src/dataset/impl_dataset.rs
```diff
@@ -724,50 +740,119 @@ where
...
+        parameters: &[M],
+        eval: C,
+    ) -> Array1<F>
+    where
+        DT: Data<Elem = E>,
+        M: for<'c> Fit<ArrayView2<'c, F>, ArrayView2<'c, E>, ER, Object = O>,
+        O: for<'d> PredictRef<ArrayView2<'a, F>, ArrayBase<DT, Ix2>>,
+        C: Fn(&ArrayBase<DT, Ix2>, &ArrayView2<E>) -> F,
+    {
+        let mut accuracies = Array1::from_elem(parameters.len(), F::zero());
```
bytesnake, line 792: not necessarily `accuracies`, could be called `metrics` or `eval_results`

src/dataset/impl_dataset.rs
```diff
@@ -724,50 +740,119 @@ where
...
+        for acc in self
+            .iter_fold(k, |train| {
+                parameters
+                    .iter()
+                    .map(|p| p.fit(&train).unwrap())
+                    .collect::<Vec<_>>()
```
bytesnake, line 798: you should be able to bail out when one model fails with:
```rust
.map(|p| p.fit(&train)?)
.collect::<Result<Vec<_>, crate::error::Error>>()
```

src/dataset/impl_dataset.rs
```diff
@@ -724,50 +740,119 @@ where
...
 
         objs.into_iter().zip(self.sample_chunks(fold_size))
-        //
     }
-}
 
-fn assist_swap_array2<F>(slice: &mut [F], index: usize, fold_size: usize, features: usize) {
-    if index == 0 {
-        return;
+    pub fn cross_validate_mt<O, DT, ER: std::error::Error, M, C>(
```
bytesnake, line 780: what does `_mt` means here?

bytesnake (commented):
I really like the proposal, because
1. it makes the associated type easier to read
2. makes error handling more explicit
3. allows to perform cross validation in an easier way

~~what do you think about renaming the functions to `cross_validate_sum` and `cross_validate_mean`?~~

src/dataset/impl_dataset.rs
bytesnake: > Currently two methods, one for single targets and one for multi-targets. Can they be merged?

well I should read the PR message more carefully :smile:

src/dataset/impl_dataset.rs
```diff
@@ -724,50 +740,119 @@ where
...
+                        eval(&predicted, valid.targets())
+                    })
+                    .collect::<Array1<F>>()
+            })
+        {
+            for i in 0..accuracies.len() {
+                accuracies[i] += acc[i];
+            }
+        }
+        accuracies
```
bytesnake, line 814: this should also be divided by the number of folds

src/dataset/impl_dataset.rs
```diff
@@ -724,50 +740,119 @@ where
...
+
+    pub fn cross_validate<O, DT, ER: std::error::Error, M, C>(
+        &'a mut self,
+        k: usize,
+        parameters: &[M],
+        eval: C,
+    ) -> Array1<F>
+    where
+        DT: Data<Elem = E>,
+        M: for<'c> Fit<ArrayView2<'c, F>, ArrayView2<'c, E>, ER, Object = O>,
```
bytesnake, line 825: can't you make this generic with 
```
M: for<'c> Fit<ArrayView2<'c, F>,T, ER, Object = O>,
T: AsTargets<Elem = E>
```
for multi and single targets?

Cargo.toml
Sauro98: whoops 😄

algorithms/linfa-logistic/src/lib.rs
Sauro98: 👍🏻

algorithms/linfa-tsne/examples/tsne.rs
Sauro98: While we are at it, I think it would be good to run the examples in the CI too, in the last PR I caught a similar error on a decision tree example I believe and these kinds of errors go unnoticed during testing. I can try to add that in this PR if it makes sense

src/dataset/impl_dataset.rs
Sauro98: Yes but this cross-validation proposal is still very early stage. I have in mind to return a struct with the set of fit results in order to be able to see, for example, the number of fits that returned an error and maybe on which parameters, and then add a method (maybe called `eval_with`) which takes a closure and only runs on successful fits. This is still very "cloudy" so I am not paying too much attention to the naming right now, but I agree on `eval_results` since one could compare them by loss instead of accuracy

src/dataset/impl_dataset.rs
Sauro98: Whoops only did so for the single target version. Maybe I should use a macro to generate both so that errors like this one do not happen again

src/dataset/impl_dataset.rs
Sauro98: The problem is the prediction. Single target algorithms do not have an implementation of `PredictRef` returning an Array2 and so things can get a little tricky when giving arguments to the closure. I'll see if I can come up with something, but it would mean adding a match on the number of dimensions of the input targets, which could be two because it is an actual multi-target model or just because an axis was added. Then if it it the latter case the user would get a multi-target evaluation (like `r2`) and would have to call `[0]` on it which is not very pretty

algorithms/linfa-tsne/examples/tsne.rs
bytesnake: yeah you could pass `--examples` to the build workspace

src/dataset/impl_dataset.rs
bytesnake: sounds like a good idea :+1:

src/dataset/impl_dataset.rs
bytesnake: have you tried to add a `I: Dimension` bound and then make `O` and `C` generic over that dimensionality? Something like:
```rust
O: for<'d> PredictRef<ArrayView2<'a, F>, ArrayBase<DT, I>>,
C: Fn(&ArrayBase<DT, I>, &ArrayBase<DT, I>) -> F,
I: Dimension
```

Sauro98: I guess that my main point is that I can't distinguish between multi-target datasets and single-target datasets if not by checking the dimension of the second axis at runtime, and I wanted to avoid forcing the user to average over 1 value, but the only way to do so is to call `try_single_target` in a dedicated function. Or at least this is what i concluded earlier, tomorrow I'll play with it and see if I can figure out something.

YuhanLiin:
> @YuhanLiin I got this test error on kmeans just once and I haven't been able to reproduce it:
> 
> ```rust
> ---- k_means::init::tests::test_sample_subsequent_candidates stdout ----
> thread 'k_means::init::tests::test_sample_subsequent_candidates' panicked at 'assertion failed: `(left == right)`
>   left: `3`,
>  right: `2`', algorithms/linfa-clustering/src/k_means/init.rs:259:9
> ```
> 
> It seems like the test takes a fixed seed in input so it should be deterministic, right? Could there be some randomness in the rng seeds for each thread?
> Maybe it was just my machine acting up sweat_smile, and also the error doesn't show up on the CI system, but I still tagged you in case you have some insight on why it might have happened

That RNG is seeded once per Rayon thread (not once per Rayon task), so if Rayon spawns a different number of threads on your machine then you may get different results. I'm going to look into a more deterministic way to run this test.

algorithms/linfa-tsne/examples/tsne.rs
Sauro98: I would need to add an error variant for PCA error to the t-sne error struct in order to do this, do you think it is reasonable or should I leave `unwrap`?

src/dataset/impl_dataset.rs
Sauro98: Basically, you can have just one of the following two nice properties:

* Having just one function `cross_validate` which handles both single and multi-dimensional cases
* Having the ability to call `try_single_target` on "fake" multi-dimensional datasets

Knowing that all functions in the `dataset` crate return an `Array2` as targets by default.

Basically, I can parameterize by dimension as you mention but then I cannot call `try_single_target` (inferring that the dataset should have single targets from the predict method signature), so the user would have to do something like this:

```rust
let mut dataset = linfa_datasets::diabetes();

let models = vec![whatever];

let evaluations = dataset.cross_validate(5, &models, |prediction, truth| truth.r2(prediction).unwrap()[0]);
```

because regression metrics automatically compute multi-target metrics if the caller is an array2. This could get even more confusing since `prediction.r2(truth)` is single-target if the prediction is too.

Also, I would like the `eval` closure to return `F` in single-dimensional cases and `Array1<F>` in multi-dimensional cases, so that one can check the fold-averaged evaluation for every target separately and then perform averaging or whatever else they want, without having to average across targets in every fold, and I don't know how I can do that with generics.

Sauro98: Also, maybe the eval closure should return a result so that `r2?` and `confusion_matrix()?.accuracy()` can be used

Sauro98:
I updated the PR description with the changes brought and some considerations on the single-target/multi-target issue

bytesnake:
I will review later this day :)

algorithms/linfa-clustering/src/appx_dbscan/hyperparameters.rs
```diff
@@ -99,21 +99,21 @@ impl<F: Float> AppxDbscanHyperParams<F> {
     }
 
     fn build(tolerance: F, min_points: usize, slack: F) -> Self {
-        if tolerance <= F::cast(0.) {
+        if tolerance <= F::zero() {
```
bytesnake, line 102: there is also `tolerance.is_negative()`

algorithms/linfa-pls/src/pls_generic.rs
```diff
@@ -259,7 +264,7 @@ impl<F: Float, D: Data<Elem = F>> Fit<'_, ArrayBase<D, Ix2>, ArrayBase<D, Ix2>>
                 Algorithm::Nipals => {
                     // Replace columns that are all close to zero with zeros
                     for mut yj in yk.gencolumns_mut() {
-                        if *(yj.mapv(|y| num_traits::float::Float::abs(y)).max().unwrap())
+                        if *(yj.mapv(|y| num_traits::float::Float::abs(y)).max()?)
```
bytesnake, line 267: does `y.abs()` now work?

src/metrics_classification.rs
```diff
@@ -267,6 +267,16 @@ where
     T: AsTargets<Elem = L> + Labels<Elem = L>,
 {
     fn confusion_matrix(&self, ground_truth: ArrayBase<S, Ix1>) -> Result<ConfusionMatrix<L>> {
+        self.confusion_matrix(&ground_truth)
+    }
+}
+
+impl<L: Label, S, T> ToConfusionMatrix<L, &ArrayBase<S, Ix1>> for T
```
bytesnake, line 274: I'm curious how this does not run into conflicts with the trait definition above? I always thought that in the instance of two `for T` with `T` bounded the compiler runs into issues.

bytesnake: I also just saw that all tests are commented out here, I'm not sure where this happened :thinking:

src/dataset/impl_dataset.rs
```diff
@@ -724,50 +740,213 @@ where
...
+    ///
+    /// // mutability needed for fast cross validation
+    /// let mut dataset = linfa_datasets::diabetes();
+    ///
+    /// let models = vec![model1, model2, ... ];
+    ///
+    /// let r2_scores = dataset.cross_validate_mt(5,&models, |prediction, truth| prediction.r2(truth))?;
+    ///
+    /// ```
+    pub fn cross_validate_mt<O, DT, ER, M, FACC, C>(
```
bytesnake, line 818: I still have to play around with the bounds but don't have enough time right now :/

bytesnake (approved):

src/dataset/impl_dataset.rs
Sauro98: I pinged the ndarray maintainer on discord and maybe there's a way to get around the Array1/Array2 construction problem by collecting to vec and then using stack. I'll try and let you know how it goes

src/metrics_classification.rs
Sauro98: I'm not sure tbh. I tried to generalize by using `std::borrow::Borrow` but it wasn't straightforward, so I just split the two implementations. Maybe I should try using `AsRef`? 

> I also just saw that all tests are commented out here, I'm not sure where this happened 🤔

I think I always saw them commented, I'll try to run them and see how it goes

algorithms/linfa-clustering/src/appx_dbscan/hyperparameters.rs
Sauro98: From the documentation it seems like `is_negative` is the `< 0` operator, not the `<= 0` operator

algorithms/linfa-pls/src/pls_generic.rs
Sauro98:
```
multiple applicable items in scope

multiple `abs` foundrustc(E0034)
```

algorithms/linfa-clustering/src/appx_dbscan/hyperparameters.rs
bytesnake: yeah I should do them when I'm fully awake :flushed:

Sauro98:
Ok, I re-enabled the tests that were commented out, the tests just needed a little tweak to work with the non-determinism of the hashset used to construct the confusion matrices. There is still one test that is not passing, for what I believe are good reasons. Basically, it tries to compute the confusion matrix between two sets by first calling `with_labels` on the ground truth set to exclude one label. The problem now is that the two sets have different lengths and so the confusion matrix comes out messy, as it can be expected. Is this a feature that would be useful, and so that I should try to fix, or is it better to just remove that single test? 
The code in the test was really outdated so maybe it reflected a property of some previous implementation. The good thing is that it allowed me to catch a weird bug in the `with_labels` method.

I'm available for a quick chat in zulip/zoom if that's needed

bytesnake:
added a PR for this PR here https://github.com/Sauro98/linfa/pull/3 

 * uses `WithLapack` for `linfa-pls` (renaming of function still missing)
 * adds build script for archlinux

algorithms/linfa-pls/src/pls_generic.rs
bytesnake: done in https://github.com/Sauro98/linfa/pull/3/commits/e9bdd75e66a6cd1a73cc52e132e96c2e64116228

src/dataset/impl_dataset.rs
bytesnake: I played around with the trait bounds and came up with this: https://github.com/Sauro98/linfa/pull/3/commits/ee5e368c33e5589b59fb91c72f25ca9cee9e4b3c

bytesnake: it only works if the evaluation metric is the same for all targets though :thinking: perhaps we could rename the other to `cross_validate_multi` and use it for per-target metric function calls.

Sauro98: Well, I like the idea of having a generic `cross_validate` which only allows you to get a summary over all the targets in case of multi target datasets, and then another one with a longer name that allows the user to keep targets separated.
This way one needs to use the less simple name only if they actually care about keeping the targets separated.

Also, by forcing the closure to take only array1s and then iterating over the prediction rows, you are bypassing the `MultiTargetRegression` trait ensuring no conflict when calling `prediction.metric(truth)` or `truth.metric(prediction)`

Sauro98:
I'm writing tests for the `with_label` method and I am wondering how the function should behave. Right now it works well for the single target case, but not for the multi-label case. What I mean is that:
```rust
let records = array![[0.,0.],[0.,0.]];
let targets = array![[0,7],[1,8]];
let dataset = DatasetBase((records, targets)).with_labels(&[&[0,1],&[8]]);
```
returns a dataset with both samples inside, while I would expect the first sample to be removed since its second target does not figure in the allowed ones.

Should it behave like I am expecting or should the sample be kept, since its first target is allowed? In the latter case, we can simplify the method by accepting just a slice instead of a slice of slices. Otherwise ,I'll try to modify the method to check that each target is allowed.

bytesnake:
I would expect that this is an `any`: if any target is contained in one of the labels, then the sample is retained. I think that `with_labels` only accepting a simple slice is enough for our usecase now. You can create a case where you want to retain a label in one target and not the other and then the modificated `with_labels` would behaviour differently. But nobody is using that right now. In the future we could add a `with_labels_any` and `with_labels_all` which accepts a double level slice.

Sauro98:
It keeps throwing a segmentation fault when running tests with tarpaulin on linfa-pls

bytesnake:
> It keeps throwing a segmentation fault when running tests with tarpaulin on linfa-pls

mhh I will try to debug this

bytesnake:
this is caused in line https://github.com/rust-ml/linfa/pull/122/files#diff-d93fb3eb4fda95138ffbeb69520f77ce15fc2b6698b5db7eadd8b3c6aa4231d5R97

bytesnake:
taking the eigendecomposition in the sign flippy function by value makes clippy happy. Can you add that? Haven't opened another PR for this
```
--- a/algorithms/linfa-pls/src/pls_svd.rs
+++ b/algorithms/linfa-pls/src/pls_svd.rs
@@ -74,7 +74,7 @@ impl<F: Float, D: Data<Elem = F>> Fit<ArrayBase<D, Ix2>, ArrayBase<D, Ix2>, PlsE
         // safe unwraps because both parameters are set to true in above call
         let u = u.unwrap().slice_move(s![.., ..self.n_components]);
         let vt = vt.unwrap().slice_move(s![..self.n_components, ..]);
-        let (u, vt) = utils::svd_flip(&u, &vt);
+        let (u, vt) = utils::svd_flip(u, vt);
         let v = vt.reversed_axes();
 
         let x_weights = u;
diff --git a/algorithms/linfa-pls/src/utils.rs b/algorithms/linfa-pls/src/utils.rs
index fc85677..1b780e0 100644
--- a/algorithms/linfa-pls/src/utils.rs
+++ b/algorithms/linfa-pls/src/utils.rs
@@ -89,8 +89,8 @@ pub fn svd_flip_1d<F: Float>(
 }
 
 pub fn svd_flip<F: Float>(
-    u: &ArrayBase<impl Data<Elem = F>, Ix2>,
-    v: &ArrayBase<impl Data<Elem = F>, Ix2>,
+    u: ArrayBase<impl Data<Elem = F>, Ix2>,
+    v: ArrayBase<impl Data<Elem = F>, Ix2>,
 ) -> (Array2<F>, Array2<F>) {
     // columns of u, rows of v
     let abs_u = u.mapv(|v| v.abs());
@@ -101,7 +101,7 @@ pub fn svd_flip<F: Float>(
         .and(&max_abs_val_indices)
         .and(&range)
         .apply(|s, &i, &j| *s = u[[i, j]].signum());
-    (u * &signs, v * &signs.insert_axis(Axis(1)))
+    (&u * &signs, &v * &signs.insert_axis(Axis(1)))
 }
```

Sauro98:
I will

Sauro98:
I have also updated the contribution guide to reflect the changes in the `Fit` trait, and I removed an outdated section about prediction traits, since they were discussed later in the documents in the updated way. 
I did not add anything about the new Lapack traits, since I didn't know whether you wanted to do it; otherwise, I'd be happy to try and write something down 👍🏻 .

bytesnake (approved):
do you also want to add an example snippet for the website? Otherwise ready to merge :+1:

Sauro98:
There's already a snippet called cross-validation, maybe I could rename that to k-folding and add another one for cross-validation? I would basically adapt the elasticnet_cv example

bytesnake:
sounds good :+1:

PR #123, opened 2021-04-21
Fix occasional subsequent_candidates test error

YuhanLiin:
Intended to fix the test failure observed in #122 by making one of the distance values 0 rather than just a low value. This guarantees that the output only consists of 2 elements.


bytesnake:
it would be great if there is a way to tell a test only to run in single threading mode

bytesnake (approved):

PR #125, opened 2021-04-25
Use least_squares solver from ndarray-linalg

### LINKED ISSUE #25: Improve Linear Regression: Use ndarray-linalg's lstsq, improve names and follow conventions
There are some improvements to the linear regression / least squares code which would be nice to implement.

- Use least squares from ndarray-linalg (lapack), [now that it is exposed](https://github.com/rust-ndarray/ndarray-linalg/pull/197)
- Address comments from original [PR](https://github.com/rust-ml/linfa/pull/20)
    - Fix names `fit_intercept`, `fit_intercept_with_normalisation` to be consistent with linfa-logistic
- Follow [C-GETTER](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter) for getter names

janmarthedal:
Use the least squares solver from ndarray-linalg instead of solving the normal equations explicitly.

Resolves #25


algorithms/linfa-linear/src/ols.rs
```diff
@@ -205,10 +204,13 @@ where
     B: Data<Elem = F>,
     C: Data<Elem = F>,
 {
-    let rhs = X.t().dot(y);
-    let linear_operator = X.t().dot(X);
-    linear_operator
-        .solve_into(rhs)
+    // It would be cleaner to use ndarray_linalg::LeastSquaresSvd::least_squares,
+    // but that is currently not possible since B = C is required
+    let X = X.to_owned();
```
bytesnake, line 209: can you accept a view representation as parameter and avoid the copy here?

bytesnake (approved):

algorithms/linfa-linear/src/ols.rs
janmarthedal: I don't see how, since the types of the data containers for `X` and `y` are different (`B` and `C` respectively). In an upcoming version of `ndarray-linalg` this restriction has been lifted so we don't have to make an explicit copy here (https://github.com/rust-ndarray/ndarray-linalg/pull/272).

bytesnake: the data is copied for the intercept branch anyways. I pushed a PR https://github.com/janmarthedal/linfa/pull/1 which adds the copy only for the branch without intercept computation

janmarthedal: Ah, I see, that makes sense. Thank you for helping out!

PR #129, opened 2021-05-09
Change confusion matrix to run on validation dataset

jeremyarde:
Ran into a typo that didn't allow me to run the multi-class svm example


bytesnake:
thanks!

Issue #119, opened 2021-04-15
Nearest Neighbour algorithms


YuhanLiin:
Nearest neighbour algorithms are an item in the roadmap. They're helpful in clustering algorithms, particularly DBSCAN. There are a number of existing implementations of advanced nearest neighbour algorithms in the Rust ecosystem, but they are not compatible with `ndarray` so they'd need to be modified to be included in Linfa.

Existing algorithms I found:
- Linear search: The simplest and least efficient method. O(N) time. Already exists in our DBSCAN implementation.
- K-D tree: O(logN) data structure. The [kdtree](https://docs.rs/kdtree/0.6.0/kdtree/index.html) crate looks pretty solid.
- BSP tree: O(logN) data structure. This [binary_space_partition](https://docs.rs/binary-space-partition/0.1.2/binary_space_partition/) crate is pretty basic so I'm not sure how useful it is to us. BSP tree is also the nearest neighbour implementation use by SmartCore.
- Ball tree: O(logN) as well. This [ball_tree](https://docs.rs/ball-tree/0.2.0/ball_tree/) crate implements the algorithm.
- R* tree: O(logN) algorithm. The [rstar](https://docs.rs/rstar/0.8.2/rstar/) crate is the most well-maintained of the nearest neighbour implementations with existing benchmarks.

We should pick at least one of the O(logN) algorithms to integrate into Linfa, along with linear search due to its simplicity. I'm not sure how many of these algorithms we really need since I don't know the detailed tradeoffs between them.

Sauro98:
I used the `rstar` crate in an initial implementation of approximate dbscan algorithm and it was nice to use. The only thing is that it wasn't possible to use it with an arbitrary number of features, so I had to use const generics. Maybe now that const generics are in stable rust it could be easier easier to use it inside linfa, it would certainly be interesting to try 👍🏻

YuhanLiin:
For it to be compatible with the rest of Linfa we'd have to make `rstar` work with `ndarray`, which uses runtime values as dimensions, so I don't think const generics helps there.

bytesnake:
this is a good analysis of the state of space partition algorithms atm :+1: do you think that approximate nearest neighbor would also fit into that? I have used the [hnsw](https://github.com/rust-cv/hnsw) crate for sparse kernel approximation here https://github.com/rust-ml/linfa/blob/master/algorithms/linfa-kernel/src/sparse.rs#L24 and it worked well

YuhanLiin:
[...]

PR #134, opened 2021-05-19
Dbscan nearest neighbour integration

YuhanLiin:
Allow DBSCAN to use nearest neighbour algorithms for range query. The only major breaking change to DBSCAN is the removal of the `Transformer` impl for `DbscanHyperparamBuilder`. The impl didn't really make sense on a builder struct and didn't play well with NearestNeighbour structs. The `Clone` and `PartialEq` derives on the hyperparameters were also removed since they didn't work with NearestNeighbour and are not needed to run the algorithm. Breaking changes are also required for `linfa-nn` outlined [here](https://github.com/rust-ml/linfa/issues/119#issuecomment-841839071).


YuhanLiin:
Turns out serde derives don't work really well with trait objects, and `Deserialize` doesn't work at all even with macros like `typetag`. Might want to use an enum instead of trait object to represent `NearestNeighbour` types.

YuhanLiin:
Changed NearestNeighbour trait objects to an enum and everything works fine. Will run benchmarks soon

YuhanLiin:
Compared the benchmarks and the differences were insane.
Master benchmarks:
```
dbscan/dbscan/10        time:   [28.040 us 28.088 us 28.137 us]                              
Found 7 outliers among 100 measurements (7.00%)
  3 (3.00%) high mild
  4 (4.00%) high severe
dbscan/dbscan/100       time:   [2.6009 ms 2.6080 ms 2.6169 ms]                               
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
Benchmarking dbscan/dbscan/1000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 27.4s, or reduce sample count to 10.
dbscan/dbscan/1000      time:   [269.90 ms 270.38 ms 271.02 ms]                               
Found 6 outliers among 100 measurements (6.00%)
  3 (3.00%) high mild
  3 (3.00%) high severe
```
Note that I didn't include the results for `dbscan/10000` since that was taking way too long.

Benchmark with KD Tree:
```
dbscan/dbscan/10        time:   [26.876 us 27.045 us 27.240 us]                              
                        change: [-4.5658% -2.9422% -1.5433%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high severe
dbscan/dbscan/100       time:   [653.07 us 660.53 us 668.45 us]                              
                        change: [-74.642% -74.385% -74.103%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
dbscan/dbscan/1000      time:   [14.886 ms 14.977 ms 15.082 ms]                               
                        change: [-94.497% -94.461% -94.422%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
Benchmarking dbscan/dbscan/10000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 48.7s, or reduce sample count to 10.
dbscan/dbscan/10000     time:   [465.09 ms 466.72 ms 468.52 ms]                                
Found 7 outliers among 100 measurements (7.00%)
  5 (5.00%) high mild
  2 (2.00%) high severe
```
We might want to make `balltree` the default instead of `kdtree` because `kdtree` panics on non-contiguous data points, but even then the speed difference is still remarkable.

bytesnake:
will review shortly

algorithms/linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -71,27 +74,60 @@ use linfa::{DatasetBase, Float};
...
+    /// * `nn_impl = KdTree`
+    pub fn params<F: Float>(
+        min_points: usize,
+    ) -> DbscanHyperParamsBuilder<F, L2Dist, CommonNearestNeighbour> {
+        Self::params_with_custom_fields(min_points, L2Dist, CommonNearestNeighbour::KdTree)
+    }
+
+    /// Configures the hyperparameters with the minimum number of points, a custom distance metric,
+    /// and a custom nearest neighbour algorithm
+    pub fn params_with_custom_fields<F: Float, D: Distance<F>, N: NearestNeighbour>(
```
bytesnake, line 91: the postfix `custom_fields` sounds a bit too vague, why no `params_with_metric_nn` or just `params_with`

algorithms/linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -71,27 +74,60 @@ use linfa::{DatasetBase, Float};
...
+            dist_fn,
+            nn_impl,
+        }
     }
 }
 
-impl<F: Float, D: Data<Elem = F>> Transformer<&ArrayBase<D, Ix2>, Array1<Option<usize>>>
-    for DbscanHyperParams<F>
+impl<F: Float, D: Data<Elem = F>, DF: Distance<F>, N: NearestNeighbour>
+    Transformer<&ArrayBase<D, Ix2>, Array1<Option<usize>>> for DbscanHyperParams<F, DF, N>
```
bytesnake, line 106: this should actually be a prediction here, because it does not transform the representation of a dataset

algorithms/linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -142,62 +202,27 @@ impl<F: Float, D: Data<Elem = F>, T>
     }
 }
 
-impl<F: Float, D: Data<Elem = F>> Transformer<&ArrayBase<D, Ix2>, Array1<Option<usize>>>
-    for DbscanHyperParamsBuilder<F>
-{
-    fn transform(&self, observations: &ArrayBase<D, Ix2>) -> Array1<Option<usize>> {
+impl<F: Float, DF: 'static + Distance<F>, N: NearestNeighbour> DbscanHyperParamsBuilder<F, DF, N> {
```
bytesnake, line 205: distinguishing between params and a param builder does not make so much sense here 'cause build does not fail atm, right?

algorithms/linfa-clustering/src/dbscan/algorithm.rs
```diff
@@ -142,62 +202,27 @@ impl<F: Float, D: Data<Elem = F>, T>
...
+    pub fn transform<D: Data<Elem = F>>(
```
bytesnake, line 206: both of these transforms are actually prediction, you only have to implement `predict_ref` here. `Transform` would take the variables and transform them into integer associations

algorithms/linfa-clustering/src/dbscan/hyperparameters.rs
```diff
@@ -7,54 +8,70 @@ use serde_crate::{Deserialize, Serialize};
...
-impl<F: Float> DbscanHyperParamsBuilder<F> {
+impl<F: Float, D: Distance<F>, N: NearestNeighbour> DbscanHyperParamsBuilder<F, D, N> {
     /// Distance between points for them to be considered neighbours.
     pub fn tolerance(mut self, tolerance: F) -> Self {
         self.tolerance = tolerance;
         self
     }
 
+    /// Nearest neighbour algorithm used for range queries
+    pub fn nn_impl(mut self, nn_impl: N) -> Self {
```
bytesnake, line 42: I'd not use the `_impl` here because it's a reserved word and may confuse reader, perhaps `nn_backend` or `nn_algo` or just plain `nn`?

algorithms/linfa-clustering/src/dbscan/hyperparameters.rs
```diff
@@ -7,54 +8,70 @@ use serde_crate::{Deserialize, Serialize};
...
+        self.nn_impl = nn_impl;
+        self
+    }
+
+    /// Distance metric used in the DBSCAN calculation
+    pub fn dist_fn(mut self, dist_fn: D) -> Self {
```
bytesnake, line 48: for `fn` it's a bit more common, but why not `dist_metric`?

bytesnake (approved):
added some comments on the nomenclature, awesome to see those performance gain

algorithms/linfa-clustering/src/dbscan/algorithm.rs
YuhanLiin: `build` does panic if the builder contains invalid parameters, but we can move those checks elsewhere during initialization. Might make sense to get rid of the builder altogether and have the hyperparams act as the builder.

algorithms/linfa-clustering/src/dbscan/algorithm.rs
YuhanLiin: What's the distinction between `predict`, `transform`, and `fit` in this context?

algorithms/linfa-clustering/src/dbscan/hyperparameters.rs
YuhanLiin: I'm going to use `nn_algo` instead.

algorithms/linfa-clustering/src/dbscan/hyperparameters.rs
YuhanLiin: `dist_fn` is already used everywhere in `linfa-nn` so I want to keep it consistent. If `dist_fn` is confusing I can change it to `dist_metric` across both crates.

YuhanLiin:
How should I handle the errors thrown by the NN operations?

algorithms/linfa-clustering/src/dbscan/algorithm.rs
bytesnake: * `fit` takes a training dataset and returns a model, for NN this would be all points because they are interpolated
* `predict` takes a reference to data and predict their clusters (or noise)
* `transform` takes a dataset and transform the variables into the embedding (either integers or sparse one-hot matrix)

algorithms/linfa-clustering/src/dbscan/algorithm.rs
bytesnake: it's easier without the build function call, perhaps just move the build call into prediction

algorithms/linfa-clustering/src/dbscan/hyperparameters.rs
bytesnake: yeah `dist_fn` is obvious enough

bytesnake:
> How should I handle the errors thrown by the NN operations?

DBSCAN should have its own error variant which can wrap the NN errors

bytesnake:
thanks!

Issue #135, opened 2021-05-20
Enable features in CI


YuhanLiin:
Right now CI doesn't enable any Cargo features. I suggest that we run a separate `cargo check` on the algorithm crates with all non-conflicting features enabled, to avoid bugs like #128. At the very least we should check the crates with `serde` enabled.

bytesnake:
yeah I tried to add `serde` support to the action but there is no way to enable a feature for all sub-crates at the moment

YuhanLiin:
Can we add a `serde` feature to all crates/sub-crates and then have a new CI step that runs `cargo check` with the `--features serde` flag?

bytesnake:
I don't think that works, you'd have to specify the target crate and wildcards like `--features */serde` are not implemented

YuhanLiin:
Can't we put `--features serde` in the args? If the CI command is running for every crate then there should be no need to specify target crate as long as every crate has the `serde` feature, which we may not want.

YuhanLiin:
Apparently you can specify features for specific crates when you call "cargo check" on a workspace. See https://doc.rust-lang.org/cargo/commands/cargo-check.html#feature-selection for more details.

bytesnake:
I would propose adding a `linfa/serde` feature flag and use it in downstream crates to solve the base error issue

PR #136, opened 2021-06-11
Enable serde features for all applicable crates in new CI checking step

YuhanLiin:
Addresses #135

### DIFF
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -33,7 +33,8 @@ jobs:
           args: --workspace
 
       - name: Run cargo check (with serde)
+        if: ${{ matrix.toolchain != '1.42.0' }} # The following args don't work on older versions of Cargo
         uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --workspace --features \"linfa-clustering/serde linfa-ica/serde linfa-kernel/serde linfa-reduction/serde linfa-svm/serde linfa-elasticnet/serde linfa-pls/serde linfa-trees/serde\"
+          args: --workspace --features "linfa-clustering/serde linfa-ica/serde linfa-kernel/serde linfa-reduction/serde linfa-svm/serde linfa-elasticnet/serde linfa-pls/serde linfa-trees/serde"
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -37,4 +37,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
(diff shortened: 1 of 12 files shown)


YuhanLiin:
Apparently some of the code doesn't compile with `serde` unless I implement Serde traits on `linfa::Error`. Don't know if I should actually do that.

bytesnake:
> Apparently some of the code doesn't compile with `serde` unless I implement Serde traits on `linfa::Error`. Don't know if I should actually do that.
> The issue is that `ElasticNet` stores a `Result`, so Serde impls are required on the error type to be derived on `ElasticNet`. I'm not sure why `ElasticNet` stores a `Result` in the first place.

then let's just add a serde feature flag for the base crate and just use it in downstream crates, I didn't foresee that .. `linfa-pls` is also complaining that imports are missing

YuhanLiin:
This doesn't work because one of the variants of `linfa::Error` contains an external type (`ShapeError`) that doesn't implement serde traits.

YuhanLiin:
@bytesnake For MSRV testing using Rust version 1.42.0 does the version of Cargo also need to be set at 1.42.0 (which it is right now) or can we use the latest Cargo? I'm running into CI issues that I believe are being caused by outdate Cargo.

bytesnake:
> This doesn't work because one of the variants of `linfa::Error` contains an external type (`ShapeError`) that doesn't implement serde traits.

then let's skip this variant for the time being

> @bytesnake For MSRV testing using Rust version 1.42.0 does the version of Cargo also need to be set at 1.42.0 (which it is right now) or can we use the latest Cargo? I'm running into CI issues that I believe are being caused by outdate Cargo.

you mean selecting features for members of workspace is not supported for 1.42.0? Perhaps you could drop the separate serde features and just depend on `linfa/serde` in the subcrates

YuhanLiin:
I don't believe it's possible to conditionally compile code using features in dependency crates. Something like `cfg("linfa/serde")` wouldn't work in the algorithm crates. My workaround for 1.42.0 is to just disable serde checking for that version entirely. If it works on 1.42.0 without serde and also on stable with serde then it probably works with serde on 1.42.0. Still much better than before in terms of serde checking.

bytesnake:
> I don't believe it's possible to conditionally compile code using features in dependency crates. Something like `cfg("linfa/serde")` wouldn't work in the algorithm crates. My workaround for 1.42.0 is to just disable serde checking for that version entirely. If it works on 1.42.0 without serde and also on stable with serde then it probably works with serde on 1.42.0. Still much better than before in terms of serde checking.

indeed I searched for a solution but there is none.Then let's keep a separate feature flag for each subcrate. Can you merge the PR now?

bytesnake (approved):

PR #137, opened 2021-06-22
Parametrize KMeans by distance metric

YuhanLiin:
Instead of calling `sq_l2_dist`, have KMeans use the distance function it was instantiated with when computing distances for the cluster assignment step. Also removed `KMeansHyperParamBuilder`.


algorithms/linfa-clustering/src/k_means/algorithm.rs
```diff
@@ -614,6 +605,33 @@ mod tests {
         y
     }
 
+    macro_rules! calc_inertia {
```
bytesnake, line 608: any reason why this is a macro?

algorithms/linfa-clustering/src/k_means/hyperparameters.rs
```diff
@@ -123,81 +51,130 @@ impl<F: Float, R: Rng + Clone> KMeansHyperParams<F, R> {
...
-    /// * `max_n_iterations = 300`.
+    /// * `tolerance = 1e-4`
+    /// * `max_n_iterations = 300`
+    /// * `n_runs = 10`
+    /// * `init = KMeansPlusPlus`
 
-    pub fn new_with_rng(n_clusters: usize, rng: R) -> KMeansHyperParamsBuilder<F, R> {
-        KMeansHyperParamsBuilder {
+    pub fn new(n_clusters: usize, rng: R, dist_fn: D) -> Self {
+        if n_clusters == 0 {
```
bytesnake, line 60: In previous implementations I returned an error in the fitting function. I think it's good practice to avoid panics in most parts of the code

algorithms/linfa-nn/src/distance.rs
```diff
@@ -1,23 +1,21 @@
 use linfa::Float;
-use ndarray::Zip;
+use ndarray::{ArrayView, Dimension, Zip};
 use ndarray_stats::DeviationExt;
 
-use crate::Point;
-
 /// A distance function that can be used in spatial algorithms such as nearest neighbour.
-pub trait Distance<F: Float>: Clone {
+pub trait Distance<F: Float>: Clone + Send + Sync {
```
bytesnake, line 6: it should also be possible to move the `D: Dimension` trait bound here to the trait definition and avoid declaring it twice , for each function once

bytesnake (approved):
just done a quick review in the morning :+1:

algorithms/linfa-clustering/src/k_means/algorithm.rs
YuhanLiin: Input types are long so I didn't want to write them all out just for a test helper

algorithms/linfa-nn/src/distance.rs
YuhanLiin: I need to use the same instance of `Distance` to compute distances on both `ArrayView1` and `ArrayView2`, so the methods themselves have to be generic with respect to the dimensionality.

algorithms/linfa-clustering/src/k_means/hyperparameters.rs
YuhanLiin: In the previous version invalid hyperparams triggered panics, so I did the same for `n_clusters`. Should I throw errors for every invalid hyperparam instead of panicking?

YuhanLiin: If we really wanted to prevent illegal hyperparams without panicking we can use something like [`tightness`](https://crates.io/crates/tightness) to express constraints on individual hyperparams at the type level.

algorithms/linfa-clustering/src/k_means/algorithm.rs
bytesnake: sounds reasonable

algorithms/linfa-clustering/src/k_means/hyperparameters.rs
bytesnake: I just added a short description and collected all errors into a single variant

algorithms/linfa-nn/src/distance.rs
bytesnake: but the only flexibility you gain is that `distance` and `rdistance` can be called with different sized ndarrays, is that the intention?

YuhanLiin: Yeah, I want the same instance of `Distance` to be able to call `distance(ArrayView1, ArrayView1)` and `distance(ArrayView2, ArrayView2)`. This is necessary in `KMeans`, where I use one instance of `Distance` to compute distances on 1D and 2D arrays. Note that `distance(ArrayView1, ArrayView2)` is not possible and the method panics if the lengths of the arrays don't match.

YuhanLiin:
Ran the benchmarks and there were no regressions from this change

algorithms/linfa-clustering/src/k_means/hyperparameters.rs
```diff
@@ -120,23 +62,73 @@ impl<F: Float, R: Rng + Clone> KMeansHyperParams<F, R> {
...
+
+    /// Change the value of `init`
+    pub fn init_method(mut self, init: KMeansInit<F>) -> Self {
+        self.init = init;
+        self
+    }
+
+    /// Return an instance of `KMeansHyperParams` after
+    /// having performed validation checks on all the specified hyperparameters.
+    pub fn build(self) -> Result<KMeansHyperParams<F, R, D>, KMeansParamsError> {
```
bytesnake, line 108: any reason why not call this `validate` and perform the check once fitting starts?

bytesnake (approved):

algorithms/linfa-clustering/src/k_means/hyperparameters.rs
YuhanLiin: If I do that then I have to call `validate` in both `fit` and `fit_with` instead of just one place. Also calling `validate` in every iteration of `fit_with` is inefficient.

bytesnake: no for me lazy validation seems the right approach, because
 * the validation cost is negligible compared to the training (and `fit_with` can be called with update hyperparameters)
 * the use of `build` prior to training/prediction looks cluttered to me

PR #141, opened 2021-06-24
Fix new Clippy errors

YuhanLiin:


bytesnake (approved):
can clippy do this in an automated way?

YuhanLiin:
`cargo clippy --fix` can do that but it's unstable. I just used Rust-Analyzer's quick fix which auto-fixed most of the errors.

PR #142, opened 2021-06-26
Introduce unchecked hyperparams to DBSCAN, Appx DBSCAN, and KMeans

YuhanLiin:
<s>Similar to what I did for KMeans, I reintroduced hyperparam builders to DBSCAN and approximate DBSCAN. This allows the verification of hyperparams to be done via returning errors rather than panicking. I believe using a builder is more idiomatic then doing a `verify()` check since the builder forces the verification to be done exactly once before the hyperparams can be used to run any algorithms.</s>
Introduce a new trait called `UncheckedHyperParams` to represent the verification of hyperparameters. Integrate this trait for DBSCAN, appx DBSCAN, and KMeans.

### DIFF
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -8,12 +8,16 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event.pull_request.draft == false
     strategy:
+      fail-fast: false
       matrix:
         toolchain:
-          - 1.42.0
-          - nightly-2021-03-24
-          - beta
+          - 1.51.0
           - stable
+        experimental: [false]
+        include:
+          - toolchain: nightly
+            experimental: true
+    continue-on-error: ${{ matrix.experimental }}
 
     steps:
       - name: Checkout sources
@@ -33,7 +37,7 @@ jobs:
           args: --workspace
 
       - name: Run cargo check (with serde)
-        if: ${{ matrix.toolchain != '1.42.0' }} # The following args don't work on older versions of Cargo
+        if: ${{ matrix.toolchain != '1.51.0' }} # The following args don't work on older versions of Cargo
         uses: actions-rs/cargo@v1
         with:
           command: check
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -11,7 +11,6 @@ jobs:
       matrix:
         toolchain:
           - 1.42.0
-          - nightly-2021-03-24
           - beta
           - stable
 
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         toolchain:
-          - 1.49.0
+          - 1.42.0
           - nightly-2021-03-24
           - beta
           - stable
@@ -33,6 +33,7 @@ jobs:
           args: --workspace
 
       - name: Run cargo check (with serde)
+        if: ${{ matrix.toolchain != '1.42.0' }} # The following args don't work on older versions of Cargo
         uses: actions-rs/cargo@v1
         with:
           command: check
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         toolchain:
-          - 1.42.0
+          - 1.49.0
           - nightly-2021-03-24
           - beta
           - stable
@@ -33,7 +33,6 @@ jobs:
           args: --workspace
 
       - name: Run cargo check (with serde)
-        if: ${{ matrix.toolchain != '1.42.0' }} # The following args don't work on older versions of Cargo
         uses: actions-rs/cargo@v1
         with:
           command: check
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -11,6 +11,7 @@ jobs:
       matrix:
         toolchain:
           - 1.42.0
+          - nightly-2021-03-24
           - beta
           - stable
 
--- a/algorithms/linfa-clustering/benches/dbscan.rs
+++ b/algorithms/linfa-clustering/benches/dbscan.rs
@@ -33,8 +33,7 @@ fn dbscan_bench(c: &mut Criterion) {
                     black_box(
                         Dbscan::params(min_points)
                             .tolerance(tolerance)
-                            .check()
-                            .unwrap()
+                            .check_unwrap()
                             .transform(&dataset),
                     )
                 });
--- a/algorithms/linfa-clustering/benches/dbscan.rs
+++ b/algorithms/linfa-clustering/benches/dbscan.rs
@@ -28,13 +28,12 @@ fn dbscan_bench(c: &mut Criterion) {
                 let centroids =
                     Array2::random_using((min_points, n_features), Uniform::new(-30., 30.), rng);
                 let dataset = generate_blobs(cluster_size, &centroids, rng);
-                bencher.iter(|| {
-                    black_box(
-                        Dbscan::params(min_points)
-                            .tolerance(tolerance)
-                            .predict(&dataset),
-                    )
-                });
+                let model = Dbscan::params(min_points)
+                    .tolerance(tolerance)
+                    .build()
+                    .unwrap();
+
+                bencher.iter(|| black_box(model.predict(&dataset)));
             },
         );
     }
--- a/algorithms/linfa-clustering/benches/dbscan.rs
+++ b/algorithms/linfa-clustering/benches/dbscan.rs
@@ -2,6 +2,7 @@ use criterion::{
     black_box, criterion_group, criterion_main, AxisScale, BenchmarkId, Criterion,
     PlotConfiguration,
 };
+use linfa::prelude::UncheckedHyperParams;
 use linfa::traits::Predict;
 use linfa_clustering::{generate_blobs, Dbscan};
 use ndarray::Array2;
@@ -30,7 +31,7 @@ fn dbscan_bench(c: &mut Criterion) {
                 let dataset = generate_blobs(cluster_size, &centroids, rng);
                 let model = Dbscan::params(min_points)
                     .tolerance(tolerance)
-                    .build()
+                    .check()
                     .unwrap();
 
                 bencher.iter(|| black_box(model.predict(&dataset)));
--- a/algorithms/linfa-clustering/benches/dbscan.rs
+++ b/algorithms/linfa-clustering/benches/dbscan.rs
@@ -2,8 +2,7 @@ use criterion::{
     black_box, criterion_group, criterion_main, AxisScale, BenchmarkId, Criterion,
     PlotConfiguration,
 };
-use linfa::prelude::UncheckedHyperParams;
-use linfa::traits::Predict;
+use linfa::prelude::{Transformer, UncheckedHyperParams};
 use linfa_clustering::{generate_blobs, Dbscan};
 use ndarray::Array2;
 use ndarray_rand::rand::SeedableRng;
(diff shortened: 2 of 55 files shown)


bytesnake:
> Similar to what I did for KMeans, I reintroduced hyperparam builders to DBSCAN and approximate DBSCAN. This allows the verification of hyperparams to be done via returning errors rather than panicking. I believe using a builder is more idiomatic then doing a `verify()` check since the builder forces the verification to be done exactly once before the hyperparams can be used to run any algorithms.

but the performance regression caused by the parameter check is negligible to fitting and the ergonomic of lazy checking outweights it

bytesnake:
wait one idea how we can solve this: (can probably also be done with the `From` trait)
add 
```rust
trait ValidateParams {
    type Validated;

    fn validate<'a>(&'a self) -> Self::Validated;
}
```
and set `Self::Validated = Result<CheckedHyperParams, Error>`. `CheckedHyperParams` can either contain a reference to the actual parameters or, for a more complex preparation, contain an owned checked hyper parameter set.

YuhanLiin:
At an API level that would be the same as calling `build()` right? The only difference is that we don't introduce a whole new Builder type which complicates the implementation. Instead of wrapping the validated params in a newtype we could wrap the unvalidated params in a newtype. This would prevent breaking a lot of methods that are defined on the hyperparam struct.

bytesnake:
> At an API level that would be the same as calling `build()` right? The only difference is that we don't introduce a whole new Builder type which complicates the implementation. Instead of wrapping the validated params in a newtype we could wrap the unvalidated params in a newtype. This would prevent breaking a lot of methods that are defined on the hyperparam struct.

yeah this still calls `build` and simplifies it in two way:
 1. accept checked _or_ unchecked hyper parameter sets and if wished perform checking implicitly before fitting
 2. all of the current checked sets are not different unchecked ones and a simple wrapper is sufficient for them

YuhanLiin:
Should we allow unchecked params to be used for fitting? We can leverage the type system to only accept checked params while only introducing a newtype wrapper around the existing param types.

YuhanLiin:
I added a trait that represents unchecked hyperparams that can wrap and return references to checked hyperparams. I also added impls of `Fit`, `Transformer`, and `PredictRef` if those traits are also implemented for the checked hyperparams.

algorithms/linfa-clustering/benches/k_means.rs
```diff
@@ -97,7 +97,7 @@ fn k_means_incr_bench(c: &mut Criterion) {
                         .sample_chunks(200)
                         .cycle()
                         .try_fold(None, |current, batch| {
-                            let (model, converged) = clf.fit_with(current, &batch);
+                            let (model, converged) = clf.fit_with(current, &batch).unwrap();
```
bytesnake, line 100: one idea: you can add a variant `GmmError::NotConverged(model)` and branch it like:
```rust
let model = dataset.
   ....
   .try_fold(None, |current, batch| {
      clf.fit_with(current, &batch)
    });

match model {
    Some(model) | Err(GmmError::NotConverged(model)) => stats.add(model.transform(dataset.records()).mean().unwrap()),
    x => x.unwrap(),
}
```

src/hyperparams.rs
```diff
@@ -0,0 +1,79 @@
...
+use crate::{
+    prelude::Records,
+    traits::{Fit, IncrementalFit, PredictRef, Transformer},
+};
+
+/// A set of hyperparameters whose values have not been checked for validity. A reference to the
+/// checked hyperparameters can only be obtained after checking has completed. If any of the
+/// algorithm traits have been implemented on the checked hyperparameters, they will also be
+/// implemented on the unchecked hyperparameters with the checking step done automatically.
+pub trait UncheckedHyperParams {
```
bytesnake, line 12: :+1:

src/hyperparams.rs
```diff
@@ -0,0 +1,79 @@
...
+impl<R: Records, T, E, P: UncheckedHyperParams> Fit<R, T, E> for P
+where
+    P::Checked: Fit<R, T, E>,
+    E: Error + From<crate::error::Error> + From<P::Error>,
+{
+    type Object = <<P as UncheckedHyperParams>::Checked as Fit<R, T, E>>::Object;
+
+    fn fit(&self, dataset: &crate::DatasetBase<R, T>) -> Result<Self::Object, E> {
+        let checked = self.check_ref()?;
+        checked.fit(dataset)
```
bytesnake, line 45: can't you do `self.check_ref().map(|p| p.fit(dataset))` here as well?

bytesnake: same in `fit_with`

bytesnake (approved):
yes thank, I think this is the way to go with hyper-parameters. I would even propose changing the name to meta-parameters (e.g. `MetaGmmParams`) because
 * for it's immediately clear what meta means, but hyper is vague
 * it's shorter to write

src/hyperparams.rs
YuhanLiin: The error type of `check_ref()` isn't the same as the method error type but can be converted via `From`, so I need to use ? syntax.

YuhanLiin:
Not sure about renaming to meta-parameters because that'd be a massive renaming across the entire codebase and hyperparameters is a well-known term in ML (though many of our "hyperparameters" arent actually hyperparameters). We could just rename it to "parameters" and have that be an umbrella term.

relf:
FWIW, on the naming, I agree with @YuhanLiin. It reminds me I noticed inconsistencies among the subcrates using either `XXXHyperParams` or `XXXParams`. I would choose `XXXParams` as it is shorter and general.

YuhanLiin:
Integration of `UncheckedParams` into existing algorithms worked fine in all cases except when calling `predict()` on `Dataset`s, which caused some type inference problems due to the way blanket `Predict` impls work. An example of this is my change of the DBSCAN example.

YuhanLiin:
@bytesnake Is there a way to get rid of nightly from the CI. We don't need it to run Tarpaulin anymore and it causes occasional breakage.

src/hyperparams.rs
bytesnake: their variants are also returning errors, makes sense

bytesnake:
> Not sure about renaming to meta-parameters because that'd be a massive renaming across the entire codebase and hyperparameters is a well-known term in ML (though many of our "hyperparameters" arent actually hyperparameters). We could just rename it to "parameters" and have that be an umbrella term.

> FWIW, on the naming, I agree with @YuhanLiin. It reminds me I noticed inconsistencies among the subcrates using either `XXXHyperParams` or `XXXParams`. I would choose `XXXParams` as it is shorter and general.

using `XXXParams` for hyper-parameters and `XXX` for parameters is technically not correct, but 1. it's easy to understand 2. simpler to use, so I agree as well

.github/workflows/checking.yml
```diff
@@ -11,7 +11,6 @@ jobs:
       matrix:
         toolchain:
           - 1.42.0
```
bytesnake, line 13: can we at least check that the project compiles on the latest nightly or does that break as well?

algorithms/linfa-clustering/benches/dbscan.rs
```diff
@@ -28,13 +29,12 @@ fn dbscan_bench(c: &mut Criterion) {
...
-                bencher.iter(|| {
-                    black_box(
-                        Dbscan::params(min_points)
-                            .tolerance(tolerance)
-                            .predict(&dataset),
-                    )
-                });
+                let model = Dbscan::params(min_points)
+                    .tolerance(tolerance)
+                    .check()
```
bytesnake, line 34: perhaps you can also add a `check_unwrap` function which just panics

algorithms/linfa-clustering/benches/k_means.rs
```diff
@@ -91,18 +89,17 @@ fn k_means_incr_bench(c: &mut Criterion) {
...
                         .cycle()
                         .try_fold(None, |current, batch| {
-                            let (model, converged) = clf.fit_with(current, &batch);
-                            if converged {
+                            match clf.fit_with(current, &batch) {
                                 // Early stop condition for the kmeans loop
-                                Err(model)
-                            } else {
-                                Ok(Some(model))
+                                Ok(model) => Err(model),
```
bytesnake, line 100: a bit clunky but also possible :D

algorithms/linfa-clustering/examples/dbscan.rs
```diff
@@ -28,7 +28,9 @@ fn main() {
     );
 
     // Infer an optimal set of centroids based on the training data distribution
-    let cluster_memberships = Dbscan::params(min_points).tolerance(1.).predict(dataset);
+    let cluster_memberships: Result<_, _> =
```
bytesnake, line 31: mhh this is probably caused by the double blanket trait :/ perhaps two separate implementations for the algorithm traits is not ideal and we could circumenvent the problem doing the following:
```rust
pub trait ParameterSet {
    /// Error type resulting from failed hyperparameter checking
    type Error: Error;

    /// Checks the hyperparameters and returns a reference to the checked hyperparameters if
    /// successful
    fn check_ref(&self) -> Result<&Self, Self::Error>;

    /// Checks the hyperparameters and returns the checked hyperparameters if successful
    fn check(self) -> Result<Self, Self::Error>;
}


impl<F: Float> ParameterSet for AppxDbscanParams<F> {
    type Error = AppxDbscanParamsError;

    fn check_ref(&self) -> Result<&Self, Self::Error> {
        if self.is_checked {
            return Ok(&self);
        }

        if self.0.min_points <= 1 {
            Err(AppxDbscanParamsError::MinPoints)
        } else if self.0.tolerance <= F::zero() {
            Err(AppxDbscanParamsError::Tolerance)
        } else if self.0.slack <= F::zero() {
            Err(AppxDbscanParamsError::Slack)
        } else {
            Ok(&self)
        }
    }

    fn check(self) -> Result<Self::Checked, Self::Error> {
        self.check_ref()?;
        self.is_checked = true;
        Ok(self)
    }
}
```
and just constraint the type parameter for `Transform`, `Fit`, `Predict`, `FitIncremental`. This simplifies the trait definitions and general structure a lot by just imposing a single comparison performance regression.

bytesnake (commented):
another review. Synonyms to `check` would be `verify` or `validate`, but `check` is shortest so let's stick with it. I added an alternative to the double parameter set definition in the review comments

.github/workflows/checking.yml
YuhanLiin: I can try

algorithms/linfa-clustering/examples/dbscan.rs
YuhanLiin: I tried your suggested change and still got the same type inference error. It turns out the error isn't caused by the extra impls in `hyperparams.rs`, but the `Predict` impls I added in `impl_dataset.rs`. Those impls are there so that types with fallible `PredictRef` impls also have fallible `Predict` impls. However, this results in `Predict` impls with the same input type but two different output types (fallible and non-fallible), which confuses type inference. Removing those `Predict` impls means that fallible `PredictRef` impls no longer result in correct `Predict` impls (the output type is `Dataset<Result<...>>` instead of `Result<Dataset<...>>`). For this reason, I believe we should only implement `PredictRef` for infallible operations (this is already the case in master). As such, parameter types should not implement `PredictRef` at all, and opt for `Fit` and `Transformer` instead. The only implementers of `PredictRef` should be models, which do not need verification. This also clarifies the relationships between parameters, models, and Linfa traits: parameters `fit` models, parameters also `transform` into results, models `predict` results.

bytesnake: ah I see! I'm still thinking about the transform, predict distinction in our implementations. I'm of the opinion that predict is appropriate for most implementations, except where the dataset is actually transformed into a new embedding for PCA or PLS. Perhaps we just can return an error from `predict_ref` and unwrap it by default in `predict`?

YuhanLiin: From what I understand most user code call `predict` anyways, so unwrapping in `predict` just leads to more panics. If a `predict_ref` impl returns a Result then only 2 of the 4 blanket `predict` impls are applied, due to the trait bounds on the return type. Specifically, `predict` will not return `Dataset`s if `predict_ref` returns a result. If we're fine with that we could have some fallible `predict_ref` impls.

YuhanLiin:
Should I change DBSCAN and appx DBSCAN back to using predict or can this be merged?

bytesnake (approved):
yes let's merge this :) thank you so much for the improvements :+1:

bytesnake:
one idea: I think we can distinguish between checked and unchecked parameters with a const generic boolean

Issue #144, opened 2021-07-07
Fix the CI


YuhanLiin:
Right now the CI is broken for two reasons I know of. Firstly, the newest version of `thiserror`(1.0.26) breaks Clippy on older versions. This can be sidestepped by locking `thiserror` to an older version, but I don't see the value of running Clippy on older versions of Rust in the first place. We should only run code quality for stable Rust and maybe beta.
We are also pinning our version of nightly in CI. This was done because using the latest nightly caused occasional breakages in the past. Unfortunately our current nightly version is too old for some of our dependencies. We may be able to fix this by reverting back to using latest nightly, but I don't see the point in running nightly in CI since Linfa is used in stable Rust.

bytesnake:
Should we then run the complete pipeline one MVC, stable and beta and only run `cargo check` for the latest nightly version?

YuhanLiin:
IMO code quality and coverage checks should be run for stable only since those checks don't actually matter for the purposes of testing whether our code works on alternative toolchains. We don't care about whether our code passes Clippy on MSRV, we only care that it works. I also don't know why we test/check our code on nightly and beta, since we don't provide guarantees that it works on those platforms. I think we should run `cargo test` and `cargo check` on stable and MSRV and run the rest of the pipeline in just stable.

xd009642:
A standard for most library authors is to run on all three and just add something to nightly so it's allowed to fail and doesn't stop CI. It's always best to have some inkling that linfa's stopped working on nightly as other people working on ML and related areas might be using nightly for fancier const generics or other optimisations.

YuhanLiin:
You have a point about running nightly/beta checks and tests and allowing those runs to fail. I still don't see the point of running non-stable toolchains for code quality and coverage jobs, since those don't have any impact on whether Linfa "works" on a particular toolchain.

xd009642:
Oh yeah agree with that, although with tarpaulin in future nightly will provide more accurate coverage reports because of using some unstable flags :eyes:. So once that's done there'll be a benefit

bytesnake:
> IMO code quality and coverage checks should be run for stable only since those checks don't actually matter for the purposes of testing whether our code works on alternative toolchains. We don't care about whether our code passes Clippy on MSRV, we only care that it works. I also don't know why we test/check our code on nightly and beta, since we don't provide guarantees that it works on those platforms. I think we should run `cargo test` and `cargo check` on stable and MSRV and run the rest of the pipeline in just stable.

[...]

PR #145, opened 2021-07-12
Rework CI steps

YuhanLiin:
Eliminate non-stable toolchains from code quality/coverage steps and allow nightly steps to fail. Also add MSRV and nightly toolchains to the test step. Addresses #144.

### DIFF
--- a/.github/workflows/codequality.yml
+++ b/.github/workflows/codequality.yml
@@ -47,7 +47,10 @@ jobs:
 
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
-        with: stable
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
 
       - name: Get rustc version
         id: rustc-version
--- a/.github/workflows/codequality.yml
+++ b/.github/workflows/codequality.yml
@@ -8,13 +8,6 @@ jobs:
     name: codequality
     runs-on: ubuntu-latest
     if: github.event.pull_request.draft == false
-    strategy:
-      matrix:
-        toolchain:
-          - 1.42.0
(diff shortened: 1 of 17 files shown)


.github/workflows/codequality.yml
```diff
@@ -24,7 +17,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
```
bytesnake, line 19: would like to follow some best-practices (normally you're running your whole CI for all channels just in case), can we run clippy for MSVC and stable if a regression causes clippy to be insensible to bad code

.github/workflows/testing.yml
```diff
@@ -11,9 +11,16 @@ jobs:
       image: rustmath/mkl-rust:1.43.0
       options: --security-opt seccomp=unconfined
     strategy:
+      fail-fast: false
       matrix:
         toolchain:
+          - 1.49.0
           - stable
+        experimental: [false]
```
bytesnake, line 19: running the testing pipeline is the most involved one and perhaps we exclude nightly from this one as well. Regressions caused by testing of certain nightly versions are 1. hard to debug 2. not saying much about the actual test. Compilation errors, on the other hand, are much easier to understand and fix.

bytesnake (approved):
:+1:

YuhanLiin:
There are some CI steps that aren't being run and they're blocking the PR from being merged. Do I just merge it via command line?

bytesnake:
> There are some CI steps that aren't being run and they're blocking the PR from being merged. Do I just merge it via command line?

will just force merge and update the settings for the bouncer

PR #146, opened 2021-07-17
Improve docs of `linfa-hierarchical`

bytesnake:
The interface and documentation of `linfa-hierarchical` is not updated in a long time. This PR adds the example to the documentation and changes the implementation of `transform` to `predict`, which is more correct in this context.

Currently blocked on a cyclic dependency caused by implementing `linfa::Predict` for `linfa::PredictRef` with kernel data. A workaround is by adding a feature flag to `linfa-kernel` which removes all `linfa` related parts and including that into the dependency tree for `linfa`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,11 @@ optional = true
 default-features = false
 features = ["cblas"]
 
+[dependencies.linfa-kernel]
+version = "0.4.0"
+path = "algorithms/linfa-kernel"
+optional = true
+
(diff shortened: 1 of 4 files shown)


YuhanLiin:
Why do we need to reverse the dependency between `linfa` and `linfa-kernel`?

bytesnake:
to use the `Kernel` structure in the implementation of `Predict for T where T: PredictRef`

Issue #140, opened 2021-06-24
Plotting in Criterion benchmarks are extremely slow


YuhanLiin:
Many of our Criterion benchmarks (particularly the ones in `linfa-clustering`) call `plot_config`, which causes the runner to create plots at the end of the benchmarks. The issue is that the plot creation takes seemingly forever. I'm in favour of removing the plots since I don't really use them, but if someone objects to that then we can look into it.

bytesnake:
According to criterion's documentation the default is `gnuplot` and it falls back to the `plotter` crate if the binary is not existing. I would suspect that `gnuplot` is fast enough, therefore could the issue just be that we have to install it in the runner before executing the actions? https://bheisler.github.io/criterion.rs/book/user_guide/plots_and_graphs.html
Otherwise let's disable them by passing `--noplot`, i.e. `cargo bench -- --noplot`

YuhanLiin:
The criterion benchmarks don't run in CI (for good reason). The slowness is just on my side, so I'll try with gnuplot

YuhanLiin:
Everything works fine with gnuplot. It seems that there's something buggy going on with the `plotter` backend, as seen here: https://github.com/bheisler/criterion.rs/issues/471. Since we can't turn off `plotter` we should probably mention this in the README somewhere

YuhanLiin:
Best to keep this one open until the `plotters` backend actually gets fixed (or removed).

PR #149, opened 2021-07-22
Document plotters bug in README

YuhanLiin:
Addresses #140.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ This selects the `intel-mkl` system library as BLAS/LAPACK backend. On the other
 
 ## Note on Benchmarks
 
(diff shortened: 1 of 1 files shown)


bytesnake (approved):

README.md
```diff
@@ -66,6 +66,10 @@ cd linfa-reduction && cargo run --release --example pca --features linfa/intel-m
 ```
 This selects the `intel-mkl` system library as BLAS/LAPACK backend. On the other hand if you want to compile the library and link it with the generated artifacts, pass `intel-mkl-static`.
 
+## Note on Benchmarks
+
+When running the Criterion benchmarks make sure to have `gnuplot` installed on your system. If `gnuplot` is not installed, Criterion will fall back on the `plotters` backend when plotting results, which leads to the benchmark getting stuck due to [this bug](https://github.com/bheisler/criterion.rs/issues/471).
```
bytesnake, line 71: a bit easier to read:
If `gnuplot` is not installed, Criterion will produce plots with the `plotters` backend as a fallback. This leads to the benchmark getting stuck due to [this bug](https://github.com/bheisler/criterion.rs/issues/471).

bytesnake (approved):

Issue #130, opened 2021-05-10
Predicting in a loop makes loads of allocations


pYtoner:
I realise this is not a problem for most use cases and adding an option to do it otherwise would add complexity to the crate.

We could add a version of `Predit::predit` which would write the output to a buffer instead of allocating the output.

Something like `Predict::predict_into`. With a signature like `fn predict_into(&self, x: R, y: &mut T)`.

Would something like this make sense for `linfa` or is this too niche?

YuhanLiin:
A back-compatible way of adding this would be to add something like
```
pub trait Predict<R: Records, T> {
    ...
    fn predict_into(&self, x: R, out: &mut T) {
         *out = self.predict(x);
    }
}
```
which would be overridden for a more efficient implementation. This isn't really useful if `T` is a `Result`, however.

pYtoner:
Do we want to handle invalid target shapes like:
```rust
fn predict_inplace<'a>(&'a self, x: &'a R, y: &mut T) -> Result<(), InvalidTargetShape>
```
Or with a more generic result? Is it just a shape mismatch that we want to handle like this?

YuhanLiin:
We should have a `PredictError` enum that has `InvalidTargetShape` as a variant. The error type will only contain error variants that makes sense to panic on. For now we only care about shape mismatches.

PR #150, opened 2021-07-24
add PredictInto trait and impl for KMeans

pYtoner:
Work towards https://github.com/rust-ml/linfa/issues/130

It would be possible to make a default implementation for any type that implements `PredictRef`. But we do not have specialization yet.

Does only implementing this, for types that can do this more efficiently, make sense?

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -61,12 +61,16 @@ fn div_capped<F: Float>(num: F) {
 There are three different traits for predictions, `Predict` and `PredictInplace`. `PredictInplace` takes a reference to the records and writes them into targets. This should be implemented by a new algorithms, for example:
 ```rust
 impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array1<F>> for Svm<F, F> {
-    fn predict_inplace<'a>(&'a self, data: &ArrayBase<D; Ix2>, targets: &mut Array1<F>) {
+    fn predict_inplace(&self, data: &ArrayBase<D, Ix2>, targets: &mut Array1<F>) {
         assert_eq!(data.n_rows(), targets.len(), "The number of data points must match the number of output targets.");
 
         for (data, target) in data.outer_iter().zip(targets.iter_mut()) {
             *target = self.normal.dot(&data) - self.rho;
         }
+    }
+
+    fn default_target(&self, data: &ArrayBase<D, Ix2>) -> Array1<F> {
+        Array1::zeros(data.nrows())
     }
 }
 ```
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -58,7 +58,7 @@ fn div_capped<F: Float>(num: F) {
 
 ## Implement prediction traits
 
-There are three different traits for predictions, `Predict`, `PredictRef` and `PredictInplace`. `PredictInplace` takes a reference to the records and writes them into targets. This should be implemented by a new algorithms, for example:
+There are three different traits for predictions, `Predict` and `PredictInplace`. `PredictInplace` takes a reference to the records and writes them into targets. This should be implemented by a new algorithms, for example:
 ```rust
 impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array1<F>> for Svm<F, F> {
     fn predict_inplace<'a>(&'a self, data: &ArrayBase<D; Ix2>, targets: &mut Array1<F>) {
@@ -71,7 +71,7 @@ impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array1<F>> f
 }
 ```
 
-This implementation is then used by `Predict`, `PredictRef` to provide the following `records` and `targets` combinations:
+This implementation is then used by `Predict` to provide the following `records` and `targets` combinations:
 
  * `Dataset` -> `Dataset`
  * `&Dataset` -> `Array1`
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -58,7 +58,7 @@ fn div_capped<F: Float>(num: F) {
 
 ## Implement prediction traits
 
-There are three different traits for predictions, `Predict` and `PredictInplace`. `PredictInplace` takes a reference to the records and writes them into targets. This should be implemented by a new algorithms, for example:
+There are two different traits for predictions, `Predict` and `PredictInplace`. `PredictInplace` takes a reference to the records and writes the prediction into a target parameter. This should be implemented by a new algorithms, for example:
 ```rust
 impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array1<F>> for Svm<F, F> {
     fn predict_inplace(&self, data: &ArrayBase<D, Ix2>, targets: &mut Array1<F>) {
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -58,19 +58,20 @@ fn div_capped<F: Float>(num: F) {
 
 ## Implement prediction traits
 
-There are two different traits for predictions, `Predict` and `PredictRef`. `PredictRef` takes a reference to the records and produces a new set of targets. This should be implemented by a new algorithms, for example:
+There are three different traits for predictions, `Predict`, `PredictRef` and `PredictInto`. `PredictInto` takes a reference to the records and writes them into targets. This should be implemented by a new algorithms, for example:
 ```rust
-impl<F: Float, D: Data<Elem = F>> PredictRef<ArrayBase<D, Ix2>, Array1<F>> for Svm<F, F> {
-    fn predict_ref<'a>(&'a self, data: &ArrayBase<D; Ix2>) -> Array1<F> {
-        data.outer_iter().map(|data| {
-            self.normal.dot(&data) - self.rho
-        })
-        .collect()
+impl<F: Float, D: Data<Elem = F>> PredictInto<ArrayBase<D, Ix2>, Array1<F>> for Svm<F, F> {
+    fn predict_into<'a>(&'a self, data: &ArrayBase<D; Ix2>, targets: &mut Array1<F>) {
+        assert_eq!(data.n_rows(), targets.len(), "The number of data points must match the number of output targets.");
+
+        for (data, target) in data.outer_iter().zip(targets.iter_mut()) {
+            *target = self.normal.dot(&data) - self.rho;
+        }
     }
 }
 ```
 
-This implementation is then used by `Predict` to provide the following `records` and `targets` combinations:
+This implementation is then used by `Predict`, `PredictRef` to provide the following `records` and `targets` combinations:
 
  * `Dataset` -> `Dataset`
  * `&Dataset` -> `Array1`
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -58,10 +58,10 @@ fn div_capped<F: Float>(num: F) {
 
 ## Implement prediction traits
 
-There are three different traits for predictions, `Predict`, `PredictRef` and `PredictInto`. `PredictInto` takes a reference to the records and writes them into targets. This should be implemented by a new algorithms, for example:
+There are three different traits for predictions, `Predict`, `PredictRef` and `PredictInplace`. `PredictInplace` takes a reference to the records and writes them into targets. This should be implemented by a new algorithms, for example:
 ```rust
-impl<F: Float, D: Data<Elem = F>> PredictInto<ArrayBase<D, Ix2>, Array1<F>> for Svm<F, F> {
-    fn predict_into<'a>(&'a self, data: &ArrayBase<D; Ix2>, targets: &mut Array1<F>) {
+impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array1<F>> for Svm<F, F> {
+    fn predict_inplace<'a>(&'a self, data: &ArrayBase<D; Ix2>, targets: &mut Array1<F>) {
         assert_eq!(data.n_rows(), targets.len(), "The number of data points must match the number of output targets.");
 
         for (data, target) in data.outer_iter().zip(targets.iter_mut()) {
--- a/algorithms/linfa-clustering/src/k_means/algorithm.rs
+++ b/algorithms/linfa-clustering/src/k_means/algorithm.rs
@@ -412,7 +412,7 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> Transformer<&ArrayBase<DA, Ix
     }
 }
 
-impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictRef<ArrayBase<DA, Ix2>, Array1<usize>>
+impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInto<ArrayBase<DA, Ix2>, Array1<usize>>
     for KMeans<F, D>
 {
     /// Given an input matrix `observations`, with shape `(n_observations, n_features)`,
@@ -420,52 +420,33 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictRef<ArrayBase<DA, Ix2>
     ///
     /// You can retrieve the centroid associated to an index using the
     /// [`centroids` method](#method.centroids).
-    fn predict_ref(&self, observations: &ArrayBase<DA, Ix2>) -> Array1<usize> {
-        let mut memberships = Array1::zeros(observations.nrows());
-        update_cluster_memberships(
-            &self.dist_fn,
-            &self.centroids,
-            &observations.view(),
-            &mut memberships,
+    fn predict_into(&self, observations: &ArrayBase<DA, Ix2>, targets: &mut Array1<usize>) {
+        assert_eq!(
+            observations.nrows(),
+            targets.len(),
+            "The number of data points must match the number of output targets."
         );
-        memberships
-    }
-}
 
-impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictRef<ArrayBase<DA, Ix1>, usize>
-    for KMeans<F, D>
-{
-    /// Given one input observation, return the index of its closest cluster
-    ///
-    /// You can retrieve the centroid associated to an index using the
-    /// [`centroids` method](#method.centroids).
-    fn predict_ref(&self, observation: &ArrayBase<DA, Ix1>) -> usize {
-        closest_centroid(&self.dist_fn, &self.centroids, observation).0
-    }
-}
-
-impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInto<ArrayBase<DA, Ix2>, Array1<usize>>
-    for KMeans<F, D>
-{
-    /// Predict into a mutable reference of `targets`
-    ///
-    /// # Panics
-    /// `targets` must be sufficiently large to hold the targets `observations.nrows() <= targets.len()`
-    fn predict_into<'a>(
-        &'a self,
-        observations: &'a ArrayBase<DA, Ix2>,
-        targets: &mut Array1<usize>,
-    ) {
-        assert!(
-            observations.nrows() <= targets.len(),
-            "Tried to predict into a too small Array"
-        );
         update_cluster_memberships(
             &self.dist_fn,
             &self.centroids,
             &observations.view(),
             targets,
         );
+    }
+}
+
+impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInto<ArrayBase<DA, Ix1>, usize>
+    for KMeans<F, D>
+{
+    /// Given one input observation, return the index of its closest cluster
+    ///
+    /// You can retrieve the centroid associated to an index using the
+    /// [`centroids` method](#method.centroids).
+    fn predict_into(&self, observation: &ArrayBase<DA, Ix1>, target: &mut usize) {
+        assert_eq!(observation.len(), 1, "The number of data points must be 1.");
+
+        *target = closest_centroid(&self.dist_fn, &self.centroids, observation).0;
     }
 }
 
--- a/algorithms/linfa-clustering/src/k_means/algorithm.rs
+++ b/algorithms/linfa-clustering/src/k_means/algorithm.rs
@@ -451,9 +451,21 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInto<ArrayBase<DA, Ix2
     ///
     /// # Panics
     /// `targets` must be sufficiently large to hold the targets `observations.nrows() <= targets.len()`
-    fn predict_into<'a>(&'a self, observations: &'a ArrayBase<DA, Ix2>, targets: &mut Array1<usize>) {
-        assert!(observations.nrows() <= targets.len(), "Tried to predict into a too small Array");
-        update_cluster_memberships(&self.dist_fn, &self.centroids, &observations.view(), targets);
+    fn predict_into<'a>(
+        &'a self,
+        observations: &'a ArrayBase<DA, Ix2>,
+        targets: &mut Array1<usize>,
+    ) {
+        assert!(
+            observations.nrows() <= targets.len(),
+            "Tried to predict into a too small Array"
+        );
+        update_cluster_memberships(
+            &self.dist_fn,
+            &self.centroids,
+            &observations.view(),
+            targets,
+        );
     }
 }
 
--- a/algorithms/linfa-clustering/src/k_means/algorithm.rs
+++ b/algorithms/linfa-clustering/src/k_means/algorithm.rs
@@ -444,6 +444,10 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInplace<ArrayBase<DA, 
             memberships,
         );
     }
+
+    fn default_target(&self, x: &ArrayBase<DA, Ix2>) -> Array1<usize> {
+        Array1::zeros(x.nrows())
+    }
 }
 
 impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInplace<ArrayBase<DA, Ix1>, usize>
@@ -457,6 +461,10 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInplace<ArrayBase<DA, 
         assert_eq!(observation.len(), 1, "The number of data points must be 1.");
 
         *membership = closest_centroid(&self.dist_fn, &self.centroids, observation).0;
+    }
+
+    fn default_target(&self, _x: &ArrayBase<DA, Ix1>) -> usize {
+        0
     }
 }
 
--- a/algorithms/linfa-clustering/src/k_means/algorithm.rs
+++ b/algorithms/linfa-clustering/src/k_means/algorithm.rs
@@ -444,6 +444,19 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictRef<ArrayBase<DA, Ix1>
     }
 }
 
+impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInto<ArrayBase<DA, Ix2>, Array1<usize>>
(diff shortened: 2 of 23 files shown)


YuhanLiin:
While it is possible to have a default impl of `PredictInto` without specialization (that'd require `PredictRef` to be a supertrait), I think it makes more sense to have `PredictInto` only for algorithms where it's actually efficient. We can even call `predict_into` in `predict_ref`.

@bytesnake Should we have blanket impls of `PredictInto` on dataset types like the `Predict` impls in `impl_dataset.rs`?

YuhanLiin:
Actually now that I think about it you can define `predict_ref` in terms of `predict_into`, which means you can define `predict` in terms of `predict_into`, so we could just replace all definitions of `predict_ref` with `predict_into`.

algorithms/linfa-clustering/src/k_means/algorithm.rs
```diff
@@ -444,6 +444,31 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictRef<ArrayBase<DA, Ix1>
     }
 }
 
+impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictInto<ArrayBase<DA, Ix2>, Array1<usize>>
```
YuhanLiin, line 447: Can you add some tests for `predict_into` just to see that it doesn't crash? You can change some of the existing tests that call `predict` to call `predict_into`.

algorithms/linfa-clustering/src/k_means/algorithm.rs
```diff
@@ -444,6 +444,31 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> PredictRef<ArrayBase<DA, Ix1>
...
+    ///
+    /// # Panics
+    /// `targets` must be sufficiently large to hold the targets `observations.nrows() <= targets.len()`
+    fn predict_into<'a>(
+        &'a self,
+        observations: &'a ArrayBase<DA, Ix2>,
+        targets: &mut Array1<usize>,
+    ) {
+        assert!(
+            observations.nrows() <= targets.len(),
```
YuhanLiin, line 460: I think you should panic on inputs that are too long as well, since that's indicative of a programming error.

pYtoner: I was thinking that it is more flexible that way but an equality check should avoid more errors.

algorithms/linfa-clustering/src/k_means/algorithm.rs
pYtoner: Of course. Before the tests, I just wanted to get some guidance on the basic structure of how this will be done.

pYtoner:
> Would you mind making that change across the codebase?

I have a free day tomorrow. By the end of that day, it will most likely be done. Glad to be able to contribute to such a great library.

algorithms/linfa-clustering/src/k_means/algorithm.rs
YuhanLiin: I've got no problems with the structure of the new trait. It's close to the structure of `PredictRef` which makes replacement all the easier.

pYtoner:
How should we check that we have the same number of observations as targets?

```rust
assert_eq!(observations.len(), targets.len(), "The number of data points must match the number of output targets.");
```
Should we compare `len()` or `nrows()` or something else?

pYtoner:
Is it too restrictive to have an impl like this?

```rust
impl<F: Float, D, S, O> PredictRef<ArrayBase<D, Ix2>, Array2<S>> for O
where
    D: Data<Elem = F>,
    O: PredictInto<ArrayBase<D, Ix2>, Array2<S>>,
    S: Default,
{
    fn predict_ref(&self, records: &ArrayBase<D, Ix2>) -> Array2<S> {
        let mut targets = Array2::default((records.nrows(), 1));
        self.predict_into(records, &mut targets);
        targets
    }
}
```
Could there be things that predict into an `Array2` that has a shape other than `(records.nrows(), 1)`?

algorithms/linfa-clustering/src/k_means/algorithm.rs
pYtoner: Since `PredictRef` now uses `PredictInto` internally, does it still make sense to make tests for `PredictInto`?

src/dataset/impl_dataset.rs
```diff
@@ -1091,6 +1091,32 @@ where
...
+}
+
+impl<F: Float, D, S, O> PredictRef<ArrayBase<D, Ix2>, Array2<S>> for O
+where
+    D: Data<Elem = F>,
+    O: PredictInto<ArrayBase<D, Ix2>, Array2<S>>,
+    S: Default,
+{
+    fn predict_ref(&self, records: &ArrayBase<D, Ix2>) -> Array2<S> {
+        let mut targets = Array2::default((records.nrows(), 1));
```
pYtoner, line 1114: Could there be models that predict into an `Array2` that has a shape other than `(records.nrows(), 1)`?
If there are such cases, I am not sure how we could deal with them. Any ideas?

src/traits.rs
```diff
@@ -57,3 +57,8 @@ pub trait Predict<R: Records, T> {
 pub trait PredictRef<R: Records, T> {
     fn predict_ref<'a>(&'a self, x: &'a R) -> T;
 }
+
+/// Predict with model into a mutable reference of targets.
+pub trait PredictInto<R: Records, T> {
+    fn predict_into<'a>(&'a self, x: &'a R, y: &mut T);
```
bytesnake, line 63: `into` normally means that some value is consumed and consequently a new one produces, I think `PredictInplace` is more clear with function `predict_inplace`

bytesnake: you should also return an error with null type like `-> Result<(), E>`. I would then `unwrap` the error in the `Predict` implementations for simpler use. We could then continue making the distinction more explicit with `predict` and `predict_unwrap`, but in most cases an error there is a logical error in the program.

src/dataset/mod.rs
```diff
@@ -561,15 +561,25 @@ mod tests {
...
-        fn predict_ref<'a>(&'a self, _x: &'a ArrayView2<'b, f64>) -> Array2<f64> {
-            array![[0., 0.]]
+    impl<'b> PredictInto<ArrayView2<'b, f64>, Array2<f64>> for MockFittableResult {
+        fn predict_into<'a>(&'a self, x: &'a ArrayView2<'b, f64>, y: &mut Array2<f64>) {
+            assert_eq!(
+                x.nrows(),
+                y.nrows(),
+                "The number of data points must match the number of output targets."
+            );
+            *y = array![[0., 0.]];
```
bytesnake, line 582: as a note: your goal of efficient implementations is not achieved like this, you are making an allocation of array and just throwing the old one away. For example here, you would have to call `y.fill(0.0)` here. The infrastructure laid out here is just the first step, each implementation I've seen in this PR have to be modified to really benefit in terms of memory usage

src/dataset/impl_dataset.rs
bytesnake: yes for example partial least squares have multiple targets, add a function `num_targets` to your trait set?

algorithms/linfa-clustering/src/k_means/algorithm.rs
```diff
@@ -412,35 +412,41 @@ impl<F: Float, DA: Data<Elem = F>, D: Distance<F>> Transformer<&ArrayBase<DA, Ix
...
+            targets.len(),
+            "The number of data points must match the number of output targets."
+        );
+
         update_cluster_memberships(
             &self.dist_fn,
             &self.centroids,
             &observations.view(),
-            &mut memberships,
+            targets,
```
bytesnake, line 434: keep `memberships` here

bytesnake (commented):
> Actually now that I think about it you can define predict_ref in terms of predict_into, which means you can define predict in terms of predict_into, so we could just replace all definitions of predict_ref with predict_into. Would you mind making that change across the codebase?

yes this is possible and a good idea! You have to add a function to query the number of targets, because that may not be clear from context, but otherwise I'm happy to ditch `predict_ref`.

src/traits.rs
YuhanLiin: @bytesnake What type of errors do you envision being returned from prediction? Right now prediction doesn't return errors at all so I'm not sure whether we want to unwrap on possible prediction errors. For example if we throw parameter errors in `predict` I'm not sure about unwrapping that. I'm also not sure about returning Results from prediction traits, since current `predict` impls are infallible so we'd need to have a lot of useless unwraps.

src/dataset/impl_dataset.rs
```diff
@@ -1091,6 +1091,32 @@ where
     }
 }
 
+impl<F: Float, D, S, O> PredictRef<ArrayBase<D, Ix2>, Array1<S>> for O
```
YuhanLiin, line 1094: Once we finalize the new trait we can cut out `PredictRef` and define `Predict` in terms of `PredictInto` instead.

src/traits.rs
bytesnake: mainly for dimensionality errors, but note that the change is semantically. I would like to collect all variants of the error inside the corresponding error enum and avoid continuing sparkling asserts throughout the code. The behaviour is still the same

YuhanLiin: Unwrapping dimensionality errors make sense (we could even have a common error type), but we need to ensure that all the errors are ones that should be unwrapped. Maybe we should enforce that predictions only return dimensionality errors?

src/dataset/mod.rs
pYtoner: Yes, what I was doing is first replacing the `PredictRef`, and then actually implementing the more efficient versions. (The only one that I did implement a more efficient version is KMeans for now)

src/dataset/impl_dataset.rs
pYtoner: Good idea. I will remove `PredictRef` after the other changes are done.

src/traits.rs
pYtoner: Can you guys create a separate issue for the dimensionality error handling. I am happy to do this one but someone else will have to do that.

pYtoner:
What do should the signature of the `num_targets` function be?
```rust
fn num_targets(&self) -> usize
```
Is that good enough?

algorithms/linfa-clustering/src/k_means/algorithm.rs
pYtoner: Ok

src/traits.rs
bytesnake: you could add a variant `MismatchNumLabels` with expected, observed number and return that for prediction. In the future we may also have categorical features where invalid values may return errors.

> Maybe we should enforce that predictions only return dimensionality errors?

yes we can add a separate enum for prediction errors

bytesnake: > Can you guys create a separate issue for the dimensionality error handling. I am happy to do this one but someone else will have to do that.

yes sure perhaps just add a note to your issue?

bytesnake:
> What should the signature of the `num_targets` function be?
> 
> ```rust
> fn num_targets(&self) -> usize
> ```
> 
> Is that good enough?
> 
> Wouldn't `num_target_features` make more sense? `num_targets` seems to be a count of how many things we are predicting instead of how many values per prediction. What about `values_per_target`?

you could name it `num_target_variables`, but the shorter one is clear enough. You can have multiple classes per target variable, if they are not mutually exclusive they are called labels. `num_target_features` connects targets and features which are two separate things. `values_per_target` would indicate that you have something like distinct classes? They are encoded in the target type instead of array size.

pYtoner:
I need some help figuring out the correct `num_features` for some of the algorithms. (all of the non-obvious ones)

algorithms/linfa-ica/src/fast_ica.rs
```diff
@@ -247,6 +247,10 @@ impl<F: Float> PredictInplace<Array2<F>, Array2<F>> for FittedFastIca<F> {
         let xcentered = x - &self.mean.view().insert_axis(Axis(0));
         *y = xcentered.dot(&self.components.t());
     }
+
+    fn num_targets(&self) -> usize {
+        todo!()
```
pYtoner, line 252: What goes here?

algorithms/linfa-pls/src/pls_generic.rs
```diff
@@ -148,6 +148,10 @@ impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array2<F>> f
         x /= &self.x_std;
         *y = x.dot(&self.coefficients) + &self.y_mean;
     }
+
+    fn num_targets(&self) -> usize {
+        todo!()
```
pYtoner, line 153: What goes here?

algorithms/linfa-reduction/src/pca.rs
```diff
@@ -180,6 +180,10 @@ impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array2<F>> f
         );
         *targets = (records - &self.mean).dot(&self.embedding.t());
     }
+
+    fn num_targets(&self) -> usize {
+        todo!()
```
pYtoner, line 185: What goes here?

src/composing/multi_target_model.rs
```diff
@@ -50,6 +50,10 @@ impl<L, F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array2<L>
             .unwrap()
             .reversed_axes();
     }
+
+    fn num_targets(&self) -> usize {
+        todo!()
```
pYtoner, line 55: What about here?

src/dataset/mod.rs
```diff
@@ -581,6 +585,10 @@ mod tests {
             );
             *y = array![[0., 0.]];
         }
+
+        fn num_targets(&self) -> usize {
+            todo!()
```
pYtoner, line 590: Here?

YuhanLiin:
`num_targets` should take the same input as `predict_inplace`. For most algorithms you should just return the number of rows in the input data.

pYtoner:
> `num_targets` should take the same input as `predict_inplace`. For most algorithms you should just return the number of rows in the input data.

@bytesnake so the name is a bit confusing 😂

`num_targets` is not how many targets there are. It is how many targets there are per prediction. For example, partial least squares have multiple targets per prediction. The output is an `Array2` the first dimension is indeed the `input.nrows()` and the second is not 1.

src/traits.rs
```diff
@@ -57,3 +57,11 @@ pub trait Predict<R: Records, T> {
 pub trait PredictRef<R: Records, T> {
     fn predict_ref<'a>(&'a self, x: &'a R) -> T;
 }
+
+/// Predict with model into a mutable reference of targets.
+pub trait PredictInplace<R: Records, T> {
+    fn predict_inplace<'a>(&'a self, x: &'a R, y: &mut T);
+
+    /// The number of features each target has.
+    fn num_targets(&self) -> usize;
```
bytesnake, line 66: you can just add a default implementations with `fn num_targets -> 1` here

src/dataset/mod.rs
bytesnake: one?

src/composing/multi_target_model.rs
bytesnake: `self.models.len()`

algorithms/linfa-reduction/src/pca.rs
bytesnake: should be `self.embedding.nrows()`

algorithms/linfa-pls/src/pls_generic.rs
bytesnake: you can look at what is assigned to your targets. `y` here has (#obserations, #targets) and `self.y_mean` is added, so you can use `self.y_mean.ncols()`

algorithms/linfa-pls/src/lib.rs
```diff
@@ -157,12 +157,18 @@ macro_rules! pls_algo { ($name:ident) => {
...
-            fn predict_ref<'a>(&'a self, x: &ArrayBase<D, Ix2>) -> Array2<F> {
-                self.0.predict_ref(x)
+            fn predict_inplace<'a>(&'a self, x: &ArrayBase<D, Ix2>, y: &mut Array2<F>) {
+                assert_eq!(x.nrows(), y.nrows(), "The number of data points must match the number of output targets.");
+
+                self.0.predict_inplace(x, y);
+            }
+
+            fn num_targets(&self) -> usize {
+                self.0.num_targets()
```
bytesnake, line 171: I'm not sure where the trait bound is missing :confused:

algorithms/linfa-ica/src/fast_ica.rs
bytesnake: just `self.components.nrows()` like in PCA, because the number of columns in the transposed component matrix equals to the number of columns in the results which are the number of targets

src/dataset/mod.rs
bytesnake: ah I see, this is the way to go :rocket:

bytesnake:
> > `num_targets` should take the same input as `predict_inplace`. For most algorithms you should just return the number of rows in the input data.
> 
> @bytesnake so the name is a bit confusing joy
> 
> `num_targets` is not how many targets there are. It is how many targets there are per prediction. For example, partial least squares have multiple targets per prediction. The output is an `Array2` the first dimension is indeed the `input.nrows()` and the second is not 1.
> 
> For most of the algorithms, it is just simply 1 because their output is an `Array1`.

normally it's input: (#observations, #features), output: (#observations, #targets). Perhaps similar to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint we should add a `_hint` at the end to avoid name clashes with future implementations and call it `num_target_variables_hint` then?

pYtoner:
`num_target_variables_hint` is the most descriptive. I'll go with that for now.

algorithms/linfa-pls/src/lib.rs
pYtoner: For `pls_generic.rs` `self.y_mean` is an `Array1` though so it doesn't have `ncols()`. Maybe `self.coefficients.ncols()`?

pYtoner:
For pls_generic.rs self.y_mean is an Array1 though so it doesn't have ncols(). Maybe self.coefficients.ncols()?

pYtoner:
For `src/dataset/mod.rs` `num_target_variables_hint` in the `predict_inplace` we assign an `[1, 2]` shape so I'll make it `2`

algorithms/linfa-pls/src/lib.rs
pYtoner: The compiler didn't know which version of `PredictInplace` we wanted to call, because unlike in `predict_inplace` the generic target type isn't implied.

pYtoner:
Jesus what have I done

algorithms/linfa-pls/src/pls_generic.rs
```diff
@@ -136,11 +136,21 @@ impl<F: Float, D: Data<Elem = F>>
...
+        );
+
         let mut x = x - &self.x_mean;
         x /= &self.x_std;
-        x.dot(&self.coefficients) + &self.y_mean
+        *y = x.dot(&self.coefficients) + &self.y_mean;
+    }
+
+    fn num_target_variables_hint(&self) -> usize {
+        todo!()
```
pYtoner, line 153: So I am still unsure about this one.

bytesnake:
> For `src/dataset/mod.rs` `num_target_variables_hint` in the `predict_inplace` we assign a `[1, 2]` shape so I'll make it `2`

jep sorry, ndarray order is columns > rows



> After that last `todo!()` all is left is to remove the old `PredictRef` and make more efficient implementations.
> 
> Before removing `PredictRef` should we make `Predict` more generic so people who used `PredictRef` can just use `Predict` instead?
> 
> Should `Predict` be
> 
> ```rust
> fn predict<X: Borrow<R>>(&self, x: X) -> T;
> ```
> 
> Instead of:
> 
> ```rust
> fn predict(&self, x: R) -> T;
> ```
> 
> Because it calls `predict_inplace` anyway so all it needs is a reference. So why not make it so we can use it by value or by ref. `PredictRef` and `Predict` in one.
> 
> Do Some algorithms might need owned values to predict. Because then we can't do this.
> 
> Instead of doing the above, we could have this new generic predict inside `PredictInplace`. This would mean we could call `PredictInplace` `Predict` and `Predict` `PredictOwned`.
> 
> Any other ideas of how this should be done?

wait `predict` for `X: &ArrayBase<..>` is already delegated to `predict_ref` directly here: https://github.com/rust-ml/linfa/blob/master/src/dataset/impl_dataset.rs#L1092 the `predict_ref` implementation (and `predict_inplace`) just acts as a multiplexer so that you only have to implement one function to get all implementations

pYtoner:
@bytesnake oh yeah. You are right. Predict is perfect.

pYtoner:
Okay I tried to remove `PredictRef` but `Predict` is generic over targets so I can't do what I did before in `PredictRef` to get default targets.

src/dataset/impl_dataset.rs
```diff
@@ -1054,67 +1054,47 @@ impl<F: Float, D, E, T, O> Predict<ArrayBase<D, Ix2>, DatasetBase<ArrayBase<D, I
...
 
 impl<'a, F: Float, R, T, S, O> Predict<&'a DatasetBase<R, T>, S> for O
 where
     R: Records<Elem = F>,
-    O: PredictRef<R, S>,
+    O: PredictInplace<R, S>,
 {
     fn predict(&self, ds: &'a DatasetBase<R, T>) -> S {
-        self.predict_ref(&ds.records)
+        let mut targets = todo!();
```
pYtoner, line 1085: Not sure how we can get default targets here...

YuhanLiin:
I think the best way to deal with the default targets is to change `num_target_variables_hint` into `default_target`. Instead of returning the target dimensions, it'd return the default target with those dimensions (for example, if `num_target_variables_hint` returned `input.nrows()` and the target is 1D array, then `default_target` will return `Array1::zeros(input.nrows())`). This method works even if the target isn't an array.

bytesnake:
what should also work: add an associated type for the shape type and return the full shape with `targets_shape_hint`. Then use `Array::from_shape`. This gives you a bit more flexibility how to allocate

pYtoner:
> I think the best way to deal with the default targets is to change `num_target_variables_hint` into `default_target`. Instead of returning the target dimensions, it'd return the default target with those dimensions (for example, if `num_target_variables_hint` returned `input.nrows()` and the target is 1D array, then `default_target` will return `Array1::zeros(input.nrows())`). This method works even if the target isn't an array.

This is the way to go in my opinion.

I think the function could be a lot more useful if instead of passing in `x`:
```rust
    fn default_target(&self, x: &R) -> T;
```
We passed in `n_predictions`:
```rust
    fn default_target(&self, n_predictions: usize) -> T;
```

This would allow you to call it, without having to construct records. Should I do the second version or the first?

bytesnake:
> > I think the best way to deal with the default targets is to change `num_target_variables_hint` into `default_target`. Instead of returning the target dimensions, it'd return the default target with those dimensions (for example, if `num_target_variables_hint` returned `input.nrows()` and the target is 1D array, then `default_target` will return `Array1::zeros(input.nrows())`). This method works even if the target isn't an array.
> 
> This is the way to go in my opinion.
> 
> I think the function could be more useful if instead of passing in `x`:
> 
> ```rust
>     fn default_target(&self, x: &R) -> T;
> ```
> 
> We passed in `n_predictions`:
> 
> ```rust
>     fn default_target(&self, n_predictions: usize) -> T;
> ```
> 
> This would allow you to call it without having to construct records. Should I do the second version or the first?

you should do the first one, because we may have three-dimensional targets in the future with (#observations, #targets, #classes), predicting probabilities for different classes. Given the records we can easily extend that

src/composing/platt_scaling.rs
```diff
@@ -182,12 +182,19 @@ where
...
+
+        // TODO: why did we return a len 98 array when data had 100 rows??
+        let before: Array1<Pr> = self
             .obj
             .predict(data)
             .iter()
             .map(|x| platt_predict(*x, self.a, self.b))
             .collect();
+
+        assert_eq!(targets, &before);
```
pYtoner, line 197: Is this a bug in `self.obj.redict()`?

pYtoner:
Can we make a different issue for making more efficient `predict_inplace` inplementations? I did the ones that I knew how but the others are more complicated.

src/composing/platt_scaling.rs
pYtoner: It is a bug in `generate_dummy_values` I think.

bytesnake: https://github.com/rust-ml/linfa/blob/aeb354891323cca79d585d67ab67582237521643/src/composing/platt_scaling.rs#L461 this has to be 102 :D

bytesnake: we are not using `1.0` and `0.0` because of the inverse logit transformation

pYtoner: I instead made the `linspace` make `n` values instead of `n - 2`. It is more intuitive that way.

bytesnake:
> Can we make a different issue for making more efficient `predict_inplace` implementations? I did the ones that I knew how but the others are more complicated.
> 

can you update the issue https://github.com/rust-ml/linfa/issues/130 with bullet points, like:
```
### Implementation
 * [x] implement buffer reuse with `predict_inplace` 
 * [ ] add error handling for `predict_inplace`
 
## Update algorithms
 * [x] K-Means clustering
 * [ ] check other algorithms for feasibility
```

CONTRIBUTE.md
```diff
@@ -58,14 +58,19 @@ fn div_capped<F: Float>(num: F) {
 
 ## Implement prediction traits
 
-There are two different traits for predictions, `Predict` and `PredictRef`. `PredictRef` takes a reference to the records and produces a new set of targets. This should be implemented by a new algorithms, for example:
+There are three different traits for predictions, `Predict` and `PredictInplace`. `PredictInplace` takes a reference to the records and writes them into targets. This should be implemented by a new algorithms, for example:
```
bytesnake, line 61: this has to be updates, ~three~ two

bytesnake: I would also make the statement more precise:
 `PredictInplace` takes a reference to the records and writes the prediction into a target parameter.

pYtoner:
> > Can we make a different issue for making more efficient `predict_inplace` implementations? I did the ones that I knew how but the others are more complicated.
> 
> can you update the issue #130 with bullet points, like:
> 
> ```
> ### Implementation
>  * [x] implement buffer reuse with `predict_inplace` 
>  * [ ] add error handling for `predict_inplace`
>  
> ## Update algorithms
>  * [x] K-Means clustering
>  * [ ] check other algorithms for feasibility
> ```

Does it looks good?

src/composing/platt_scaling.rs
```diff
@@ -360,7 +367,7 @@ mod tests {
         let prob_values = Array1::linspace(
             F::one() / F::from(n).unwrap(),
             F::one() - F::one() / F::from(n).unwrap(),
-            n - 2,
+            n,
```
bytesnake, line 370: for evenly spaced values this has to be `n-2` with the given bounds, you have to update the call to `102`

bytesnake (approved):
did another quick check

bytesnake:
> > > Can we make a different issue for making more efficient `predict_inplace` implementations? I did the ones that I knew how but the others are more complicated.
> > 
> > 
> > can you update the issue #130 with bullet points, like:
> > ```
> > ### Implementation
> >  * [x] implement buffer reuse with `predict_inplace` 
> >  * [ ] add error handling for `predict_inplace`
> >  
> > ## Update algorithms
> >  * [x] K-Means clustering
> >  * [ ] check other algorithms for feasibility
> > ```
> 
> Does it look good?

yes looks fine, spotted a typo in the first sentence :mag:

CONTRIBUTE.md
pYtoner: Alright. It is better now.

src/composing/platt_scaling.rs
pYtoner: Right :D

bytesnake:
:drum:

PR #152, opened 2021-07-26
Params guard pattern with const generics

bytesnake:
Continues work on checked parameter sets of #142 by using const generics pattern for parameter set checking. Demonstrates how this may work on the Platt scaling implementation. WIP

 * [ ] change terminology from hyperparameter to parameters
 * [x] implement dummy implementations for transform, fit, predict, fit_with
 * [ ] port algorithms to new syntax

### DIFF
--- a/src/composing/platt_scaling.rs
+++ b/src/composing/platt_scaling.rs
@@ -18,8 +18,8 @@
 use std::marker::PhantomData;
 
 use crate::dataset::{DatasetBase, Pr};
+use crate::param_guard::{ParamGuard, ParamIntoCheckedConst};
 use crate::traits::{Predict, PredictRef};
-use crate::hyperparams::ParameterCheck;
 use crate::Float;
 
 use ndarray::{Array1, Array2, ArrayBase, ArrayView1, Data, Ix1, Ix2};
@@ -44,6 +44,7 @@ pub struct Platt<F, O> {
 }
 
 /// Parameters for Platt's Newton method
+#[derive(Clone, Debug)]
 pub struct PlattParams<F, O, const C: bool> {
     maxiter: usize,
     minstep: F,
@@ -64,10 +65,18 @@ impl<F: Float, O> Default for PlattParams<F, O, false> {
 
 impl<F: Float, O, const C: bool> PlattParams<F, O, C> {
     fn change_check_flag<const C2: bool>(self) -> PlattParams<F, O, C2> {
-        let PlattParams { maxiter, minstep, sigma, phantom } = self;
+        let PlattParams {
+            maxiter,
+            minstep,
+            sigma,
+            phantom,
+        } = self;
 
         PlattParams {
-            maxiter, minstep, sigma, phantom
+            maxiter,
+            minstep,
+            sigma,
+            phantom,
         }
     }
 
@@ -103,36 +112,28 @@ impl<F: Float, O, const C: bool> PlattParams<F, O, C> {
 
         self.change_check_flag()
     }
-
 }
 
-impl<F: Float, O> ParameterCheck for PlattParams<F, O, false> {
-    type Checked = PlattParams<F, O, true>;
-    type Error = PlattNewtonResult;
+impl<F: Float, O: Clone> ParamGuard for PlattParams<F, O, false> {
+    type Error = PlattError;
 
-    fn is_valid(&self) -> Result<(), PlattNewtonResult> {
+    fn is_valid(&self) -> Result<(), PlattError> {
         if self.maxiter == 0 {
-            return Err(PlattNewtonResult::MaxIterReached);
+            return Err(PlattError::MaxIterReached);
         }
         if self.minstep.is_negative() {
-            return Err(PlattNewtonResult::MinStepNegative(
-                self.minstep.to_f32().unwrap(),
-            ));
+            return Err(PlattError::MinStepNegative(self.minstep.to_f32().unwrap()));
         }
         if self.sigma.is_negative() {
-            return Err(PlattNewtonResult::SigmaNegative(
-                self.sigma.to_f32().unwrap(),
-            ));
+            return Err(PlattError::SigmaNegative(self.sigma.to_f32().unwrap()));
         }
 
         Ok(())
     }
+}
 
-    fn check(self) -> Result<Self::Checked, PlattNewtonResult> {
-        self.is_valid()?;
-
-        Ok(self.change_check_flag())
-    }
+unsafe impl<F: Float, O: Clone> ParamIntoCheckedConst for PlattParams<F, O, false> {
+    type Checked = PlattParams<F, O, true>;
 }
 
 #[cfg_attr(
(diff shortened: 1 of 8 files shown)


bytesnake:
@YuhanLiin got this idea and now trying to continue your work here. Hopefully this approach proves feasible

src/composing/platt_scaling.rs
```diff
@@ -43,14 +44,14 @@ pub struct Platt<F, O> {
 }
 
 /// Parameters for Platt's Newton method
-pub struct PlattParams<F, O> {
+pub struct PlattParams<F, O, const C: bool> {
```
YuhanLiin, line 47: It'd make sense to give this type param a more specific name such as `Check`/`CHECK` (or whatever the convention is for const generics).

src/composing/platt_scaling.rs
```diff
@@ -79,23 +88,29 @@ impl<F: Float, O> PlattParams<F, O> {
...
-        self
+        self.change_check_flag()
     }
 
-    fn validate(&self) -> Result<(), PlattNewtonResult> {
+}
+
+impl<F: Float, O> ParameterCheck for PlattParams<F, O, false> {
+    type Checked = PlattParams<F, O, true>;
+    type Error = PlattNewtonResult;
```
YuhanLiin, line 111: Shouldn't this be an error type, not a result type?

src/hyperparams.rs
```diff
@@ -5,22 +5,19 @@ use crate::{
...
-pub trait UncheckedHyperParams {
-    /// The checked hyperparameters
+pub trait ParameterCheck {
     type Checked;
-    /// Error type resulting from failed hyperparameter checking
     type Error: Error;
 
-    /// Checks the hyperparameters and returns a reference to the checked hyperparameters if
-    /// successful
-    fn check_ref(&self) -> Result<&Self::Checked, Self::Error>;
```
YuhanLiin, line 23: Removing `check_ref` is iffy because the blanket impls for `Predict`, `Fit`, and `IncrementalFit` rely on it. Without the blanket impls you'd need to have "checked" and "unchecked" impls of those 3 traits for every relevant parameter type, which is non-ideal. Admittedly `check_ref` doesn't work with the const generic guard pattern without some crazy reference casting, but I believe the ergonomics of const generic guards don't outweigh the benefits of having `check_ref` blanket impls.

YuhanLiin: [Here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=32e4579ff04cdfa78b8c5d685d6f5a51) is the reference cast required to make `check_ref` work with const generics. It's unsafe but probably not UB as long as the underlying layout is identical. At least Miri is fine with it.

src/composing/platt_scaling.rs
bytesnake: yes, I also thought about using an enum `IsChecked` to make it more descriptive

src/composing/platt_scaling.rs
bytesnake: this is the error enum and should be renamed to `PlattNewtonError`

src/hyperparams.rs
bytesnake: thanks I was looking for something like this

src/composing/platt_scaling.rs
bytesnake: done

bytesnake:
this actually is getting somewhere, I did the follow:
 * rename `hyperparameters.rs` to `param_guard.rs`
 * split the guarding trait into two separate. The first `ParamIntoChecked` implements the transmuting mechanism into a checked parameter set. The second `ParamGuard` provides the validation routine. This allows to provide a default implementation of the first one which transmutes the const generics (currently called `ParamIntoCheckedConst`, perhaps there ius a better name) and can be used for 99% of all implementations.
 * `as_checked`  returns a cow now to ease restriction for transitions which produces owned checked parameter sets in the future

for platt scaling it's now looking like this:
```rust
impl<F: Float, O: Clone> ParamGuard for PlattParams<F, O, false> {
    type Error = PlattError;

    fn is_valid(&self) -> Result<(), PlattError> {
        if self.maxiter == 0 { 
            return Err(PlattError::MaxIterReached);
        }
        if self.minstep.is_negative() {
            return Err(PlattError::MinStepNegative(self.minstep.to_f32().unwrap()));
        }
        if self.sigma.is_negative() {
            return Err(PlattError::SigmaNegative(self.sigma.to_f32().unwrap()));
        }

        Ok(())
    }   
}

unsafe impl<F: Float, O: Clone> ParamIntoCheckedConst for PlattParams<F, O, false> {
    type Checked = PlattParams<F, O, true>;
}
```
it provides `check`, `check_ref` and is really concise.

unclear points:
 * how can we hide the functions of `ParamIntoChecked` for internal use only
 * had to add `Clone` trait bound to model to satisfy cow
 * is there a special naming convention for functions returning cows?

src/composing/platt_scaling.rs
YuhanLiin: You could also rename the const generic itself (not sure if that's idiomatic)

YuhanLiin:
For hiding unsafe impls we can do something like this:
```rust
trait Verify {
  type Error;
  fn verify(&self) -> Result<(), Self::Error>;
}

struct ParamGuard<T: Verify, const C: bool>(pub(crate) T);

impl<T: Verify> ParamGuard<T, false> {
  fn check_ref(&self) -> &ParamGuard<T, true> {
    unsafe { // pointer casts }
  }

  fn check(self) -> ParamGuard<T, true> {
    ParamGuard(self.0)
  }
  
  // Use this as a helper to implement the builder/modification methods on unchecked params
  pub(crate) fn modify<F: FnOnce(T) -> T>(mut self, f: F) -> Self {
    Self(f(self.0))
  }
}
```
Users will work directly with `ParamGuard<MyParamSet, C>` rather than `MyParamSet`, but the Linfa traits will be implemented for `MyParamSet`. This means we need blanket impls for `ParamGuard<T, true>` and `ParamGuard<T, false>` (except for predict, which is only implemented on unchecked params). One drawback is that users may have problems implementing custom param types, since you can't define `impl`s for a type from another crate.

On another note, I don't think returning `Cow` is a good idea. It falls apart when we have params that don't implement `Clone`, and I don't foresee us needing to return owned values from `check_ref`. There's a higher chance we end up with params that can't implement `Clone` than params that actually need owned values.

YuhanLiin:
Yeah you can't implement for outside types. The struct I provided still works since all you need is the ability to modify unchecked params and to reference the inner struct, but it'd break a lot of existing code. Otherwise I can't think of a way to "hide" the unsafe impl, since each algorithm needs its own checked and unchecked param types and you'd have to ensure that each pair of types have identical layouts, so you'd have unsafe everywhere. IMO the ergonomics of const generics isn't worth having that much unsafe.

bytesnake:
I played around with different solutions and we could build a solution where we accept generic params for `Fit/FitWith/Transformer` or with GAT, but IMO the whole effort is not worth it and I'm just doing a branching here https://github.com/rust-ml/linfa/pull/152/files#diff-bc252d28dce9822c0c10dd023653314aa9c6c6cc64f070cdf9a1c444f6b1962cR113

YuhanLiin:
Will you still work on the other items in the checklist in this PR?

bytesnake:
yes sure!

bytesnake:
superseded by #158

Issue #124, opened 2021-04-22
Are we ndarray 0.15 ready yet?


relf:
I open this issue to keep track of stuff required to bump to `ndarray 0.15`. From #110 we get the following list:

- [x] approx = 0.4
- [x] argmin = 0.4.6
- [x] ndarray-csv = 0.5.1
- [x] ndarray-linalg = 0.14.0
- [x] ndarray-npy = 0.8
- [x] ndarray-rand = 0.14
- [x] ndarray-stats = 0.5
- [x] num-traits = 0.2
- [x] rand = 0.8
- [x] rand-isaac = 0.3
- [x] sprs

quietlychris:
I just submitted a pull request to `ndarray-csv` updating it to use `0.15`, and will add an update if it gets merged/a new version is posted to crates.io

quietlychris:
@relf I believe that we should be all set with `ndarray-csv` now--`0.15` should be supported with `ndarray-csv = "0.5.1"`

YuhanLiin:
Are we waiting for `argmin`, `ndarray-linalg`, and `sprs` to include `ndarray` 0.15 in their dependencies before bumping?

relf:
Well... I guess it was the plan, thinking that linfa's dependencies would upgrade rapidly. Do you suggest we could bump some of linfa's subcrates?

relf:
Yep, we agree. Btw it seems dificult to mix different versions of `ndarray` anyway (when it occurs by mistake it ends up with compilation errors).

bytesnake:
both `sprs` and `ndarray-linalg` are already using `ndarray = 0.15` in their git tree, but they are not yet published to `crates.io`. I also created an issue at `argmin` side https://github.com/argmin-rs/argmin/issues/123

relf:
`ndarray-linalg` and `argmin` on `ndarray = 0.15` are published, one to go...

bytesnake:
I would like to update all dependencies to `ndarray = 0.15` before publishing a new linfa version

mulimoen:
`sprs` 0.11.0 was published right now, and uses `ndarray = 1.15`

relf:
Closed by #156

PR #156, opened 2021-07-31
Bump ndarray to 0.15.0

YuhanLiin:
Addresses #124.
Having issues with new version of `sprs`. The compile error most likely has something to do with https://github.com/vbarrielle/sprs/issues/278 and was introduced by https://github.com/vbarrielle/sprs/pull/290.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ serde = ["serde_crate", "ndarray/serde"]
 [dependencies]
 num-traits = "0.2"
 rand = { version = "0.8", features = ["small_rng"] }
-approx = "0.5"
+approx = "0.4"
 
 ndarray = { version = "0.15", default-features = false, features = ["approx"] }
 ndarray-linalg = { version = "0.14", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,10 +37,10 @@ serde = ["serde_crate", "ndarray/serde"]
 [dependencies]
 num-traits = "0.2"
 rand = { version = "0.8", features = ["small_rng"] }
-approx = "0.4"
+approx = "0.5"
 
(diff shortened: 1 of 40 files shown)


bytesnake:
replacing the operator with `sprs::binop::csmat_binop(mat.view(), mat.transpose_view(), |x, y| x.add(*y));` seems to work as a workaround. The difference between making the explicit call to `add` and the operator makes the difference: https://docs.rs/sprs/0.11.0/src/sprs/sparse/binop.rs.html#63

bytesnake:
the construction of the sparse matrix fails though and I'm not yet sure why

bytesnake:
the new constructor assumes ordered indices, this solves it:
```rust
    // create CSR matrix from data, indptr and indices
    let mat = CsMatBase::new_from_unsorted((n_points, n_points), indptr, indices, data).unwrap();
    let mut mat = sprs::binop::csmat_binop(mat.view(), mat.transpose_view().to_other_storage().view(), |x, y| x.add(*y));
```

mulimoen:
It seems there are some strange type inference going on when using `linfa::Float` in `sprs`. You can use the following code instead:
```rust
    // create CSR matrix from data, indptr and indices
    let mat = CsMat::<F>::new_from_unsorted((n_points, n_points), indptr, indices, data).unwrap();
    let transpose = mat.transpose_view().to_other_storage();
    sprs::binop::csmat_binop(mat.view(), transpose.view(), |_, _| F::one())
```
This does the same as we do behind the scene. I have assumed `x + y != 0`, and moved this computation into `binop`, which might be more performant

bytesnake (approved):
:+1:

bytesnake:
the CI pipeline somehow failed to run here as well, I checked manually ..

Issue #151, opened 2021-07-25
Replace the term "hyperparam" with "param"


YuhanLiin:
This applies to type names, docs, as well as comments. The term "param" is shorter and more general than "hyperparam", and "hyperparam" is also a bit of a misnomer, since most of what we call "hyperparams" inside Linfa aren't actually hyperparameters. See [here](https://github.com/rust-ml/linfa/pull/142#issuecomment-874438697) for discussion on this.

YuhanLiin:
@bytesnake Should we rename instances of "hyperparam" in the docs/comments or just close this issue?

bytesnake:
Just removed `Hyper` from struct definitions in that PR, otherwise I'm still using it in most cases to distinguish them from learnable parameters.

PR #158, opened 2021-08-15
Introduce checked parameters to all algorithms

bytesnake:
Current state:

 * [x] port algorithm's parameters to new syntax
 * [ ] improve documentation (use `linfa-bayes` as prototype)
 * [ ] fix issues with `Transform` and double results


algorithms/linfa-logistic/src/hyperparams.rs
```diff
@@ -0,0 +1,142 @@
...
+        self
+    }
+}
+
+impl<F: Float, C> ParamGuard for LogisticRegressionParams<F, C> {
+    type Checked = LogisticRegressionValidParams<F, C>;
+    type Error = Error;
+
+    fn check_ref(&self) -> Result<&Self::Checked, Self::Error> {
+        Ok(&self.0)
```
YuhanLiin, line 135: So we don't need to check the params here?

algorithms/linfa-logistic/src/lib.rs
```diff
@@ -17,115 +17,28 @@
...
-
-impl<F: Float> Default for LogisticRegression<F> {
-    fn default() -> LogisticRegression<F> {
-        LogisticRegression::new()
-    }
-}
-
 type LBFGSType<F> = LBFGS<MoreThuenteLineSearch<ArgminParam<F>, F>, ArgminParam<F>, F>;
 
-impl<F: Float> LogisticRegression<F> {
```
YuhanLiin, line 82: LOL this is gonna have a massive merge conflict with my logistic regression changes.

algorithms/linfa-preprocessing/src/hyperparams.rs
```diff
@@ -0,0 +1,113 @@
...
+        self.document_frequency
+    }
+
+    pub fn stopwords(&self) -> &Option<HashSet<String>> {
+        &self.stopwords
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct CountVectorizerParams(CountVectorizerValidParams);
```
YuhanLiin, line 57: Where's the `ParamGuard` impl?

algorithms/linfa-pls/src/hyperparams.rs
```diff
@@ -0,0 +1,147 @@
...
+        self
+    }
+}
+
+impl<F: Float> ParamGuard for PlsParams<F> {
+    type Checked = PlsValidParams<F>;
+    type Error = PlsError;
+
+    fn check_ref(&self) -> Result<&Self::Checked, Self::Error> {
+        Ok(&self.0)
```
YuhanLiin, line 100: No verification here as well

algorithms/linfa-svm/src/hyperparams.rs
```diff
@@ -0,0 +1,215 @@
...
+        SvmParams::new()
+    }
+}
+
+impl<F: Float, L> ParamGuard for SvmParams<F, L> {
+    type Checked = SvmValidParams<F, L>;
+    type Error = SvmError;
+
+    fn check_ref(&self) -> Result<&Self::Checked, Self::Error> {
+        Ok(&self.0)
```
YuhanLiin, line 207: No verification here.

algorithms/linfa-tsne/src/hyperparams.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    /// Set the approximation threshold of the Barnes Hut algorithm
+    ///
+    /// The threshold decides whether a cluster centroid can be used as a summary for the whole
+    /// area. This was proposed by Barnes and Hut and compares the ratio of cell radius and
+    /// distance to a factor theta. This threshold lies in range (0, inf) where a value of 0
+    /// disables approximation and a positive value approximates the gradient with the cell center.
+    pub fn approx_threshold(mut self, threshold: F) -> Self {
+        self.0.approx_threshold = threshold;
+
```
YuhanLiin, line 110: Don't think you should put empty lines here since the methods are all 2-liners anyways.

YuhanLiin:
What's the full scope of this PR? I know it addresses #151 and also applies the `ParamGuard` trait everywhere, but it also looks like you're improving the impls of `PredictInplace` for other algorithms (#130)

YuhanLiin:
Also there's still many instances of the word `hyperparam` or `hyper-param` in the docs and the module/file names. If we want to rectify those names we should just run a ctrl/replace across the whole code-base.

algorithms/linfa-logistic/src/lib.rs
bytesnake: if it is the case (which I don't think) that this PR is merged prior to the multinomial logistic one, then I'll exclude the changes to make this PR orthogonal

bytesnake:
> What's the full scope of this PR? I know it addresses #151 and also applies the `ParamGuard` trait everywhere, but it also looks like you're improving the impls of `PredictInplace` for other algorithms (#130)

Sorry for not keeping updates here, I have to admit that the PR eats up a lot of time and I'm not so motivated communicating after working in a batch. The scope of the PR is to introduce checked parameters and uniformize the documentation and notation. I'm not sure what you mean with improving prediction, the only change in the function signature is for `linfa-bayes` where I've made it generic over the label type. I've postponed reviews for now, because this PR is still WIP. (but if needed I'll exclude changes to `linfa-logistic`)

bytesnake:
> Also there's still many instances of the word `hyperparam` or `hyper-param` in the docs and the module/file names. If we want to rectify those names we should just run a ctrl/replace across the whole code-base.

I'll keep referencing in documentation to hyper-parameters, so that it's clear to a SL/ML audience what is meant. I've also opted for the name `hyperparams.rs` as a module. The actual content is exported to the top-level mod, so that the name only matters when browsing the source-code and I find it there quite expressive. I'm removing the term `Hyper` from any structure making it shorter in total. For example for naive bayes algorithms we have the triplet `(GaussianNb, GaussianNbParams, GaussianNbValidParams)`

bytesnake:
I also took some time to think about documentation. A very useful recently stabilized feature is the use of macros for doc attributes. https://github.com/rust-lang/rust/blob/master/RELEASES.md#language-1 This allows us to include the `README.md` directly into rustdocs and avoids having to copy every change there.

bytesnake:
I will continue to split changes into separate commits, one for each algorithm

YuhanLiin:
I've added the `ParamGuard` trait to all applicable algorithms. If there's an algorithm that I missed then let me know. Otherwise this is good to merge.
The documentation improvements can go into a separate PR, since it has nothing to do with parameter checking and I want to merge this ASAP. I haven't run into any cases where `Transformer` returned a `Result`, so the double `Result` issue hasn't popped up. I can't think of a nice way to resolve the issue without mutually exclusive traits or specialization, which are both nightly, so I think we can leave it for now.

bytesnake:
> I've added the `ParamGuard` trait to all applicable algorithms. If there's an algorithm that I missed then let me know. Otherwise this is good to merge.

all algorithms are covered now :+1:

> The documentation improvements can go into a separate PR, since it has nothing to do with parameter checking and I want to merge this ASAP.

Would it be okay with you that we'll still merge the documentation changes (and also bump to MSRV to 1.54), then extend on that on a follow-up PR. I think it's too much effort to separate and has no real benefit.

> I haven't run into any cases where `Transformer` returned a `Result`, so the double `Result` issue hasn't popped up. I can't think of a nice way to resolve the issue without mutually exclusive traits or specialization, which are both nightly, so I think we can leave it for now.

you can find them by searching for `??`, for example https://github.com/rust-ml/linfa/pull/158/files#diff-7dd5b8319bc6045fdac67775d30a6f9f17a0ddd2696eafc28ab32e80834c5b65R30. This is caused by `ParamGuard` implementing `transform` with implicit check and wrapping dual results.

YuhanLiin:
Yeah merging current changes is fine. I just don't want to stuff any more unrelated changes into this already gargantuan PR. Followup PR for additional doc changes is fine.

YuhanLiin:
Resolved the double `Result` issue. The default blanket impl of `Transform` is now gated behind a marker trait, allowing custom `Transform` impl to be manually implemented on types where the original `transform` call returns `Result`. Might be worth it to turn the manual impl into a macro.

bytesnake:
> Resolved the double `Result` issue. The default blanket impl of `Transform` is now gated behind a marker trait, allowing custom `Transform` impl to be manually implemented on types where the original `transform` call returns `Result`. Might be worth it to turn the manual impl into a macro.

this seems to be reasonable straightforward

PR #159, opened 2021-08-16
Multinomial logistic regression

YuhanLiin:
Implementation of multinomial logistic regression. Also added some tests. Still need to add docs and possibly refactor.

### DIFF
--- a/algorithms/linfa-logistic/src/argmin_param.rs
+++ b/algorithms/linfa-logistic/src/argmin_param.rs
@@ -50,7 +50,7 @@ impl<F: Float, D: Dimension> ArgminDot<ArgminParam<F, D>, F> for ArgminParam<F, 
 
 impl<F: Float, D: Dimension> ArgminNorm<F> for ArgminParam<F, D> {
     fn norm(&self) -> F {
-        elem_dot(&self.0, &self.0)
+        num_traits::Float::sqrt(elem_dot(&self.0, &self.0))
     }
 }
 
--- a/algorithms/linfa-logistic/src/argmin_param.rs
+++ b/algorithms/linfa-logistic/src/argmin_param.rs
@@ -38,7 +38,7 @@ impl<F: Float, D: Dimension> ArgminSub<ArgminParam<F, D>, ArgminParam<F, D>> for
 
 impl<F: Float, D: Dimension> ArgminAdd<ArgminParam<F, D>, ArgminParam<F, D>> for ArgminParam<F, D> {
     fn add(&self, other: &ArgminParam<F, D>) -> ArgminParam<F, D> {
-        ArgminParam(&self.0 - &other.0)
+        ArgminParam(&self.0 + &other.0)
     }
 }
 
--- a/algorithms/linfa-logistic/src/argmin_param.rs
+++ b/algorithms/linfa-logistic/src/argmin_param.rs
@@ -1,4 +1,4 @@
-//! This module defines newtypes for ndarray's Array1.
+//! This module defines newtypes for ndarray's Array.
 //!
 //! This is necessary to be able to abstract over floats (f32 and f64) so that
 //! the logistic regression code can be abstract in the float type it works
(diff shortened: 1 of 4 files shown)


bytesnake:
is this PR in draft mode intentionally or ready for review?

YuhanLiin:
One observation I have of the multinomial implementation is that it tends to diverge when the input is not normalized well. Not sure if this is a natural property of the algorithm itself.

YuhanLiin:
The unnormalized wine quality also diverges on `scikit-learn`. However, the algorithm actually terminates since it hits the max iteration limit. With this implementation the algorithm just gets stuck on the first iteration of the solver forever. This likely has something to do with the linear search library we're using.

algorithms/linfa-logistic/src/lib.rs
```diff
@@ -93,79 +115,60 @@ impl<F: Float> LogisticRegression<F> {
...
     }
 
     /// Configure the initial parameters from where the optimization starts.
-    /// The `params` array must have the same size as the number of columns of
-    /// the feature matrix `x` passed to the `fit` method
-    pub fn initial_params(mut self, params: Array1<F>, intercept: F) -> LogisticRegression<F> {
-        self.initial_params = Some((params, intercept));
+    /// The `params` array must have at least the same number of rows as there are columns on the
+    /// feature matrix `x` passed to the `fit` method
+    pub fn initial_params(mut self, params: Array<F, D>) -> Self {
```
bytesnake, line 146: doesn't this try to achieve what we are doing with `FitWith`? (i.e. you can also initialize a linear model with these parameters and call `fit_with` on that)

algorithms/linfa-logistic/src/lib.rs
```diff
@@ -421,6 +476,37 @@ fn logistic_loss<F: Float, A: Data<Elem = F>>(
...
+fn multi_logistic_loss<F: Float, A: Data<Elem = F>>(
+    x: &ArrayBase<A, Ix2>,
+    y: &Array2<F>,
+    alpha: F,
+    w: &Array2<F>,
+) -> F {
+    let (log_prob, params) = multi_logistic_prob_params(x, w);
+    // Calculate loss
+    // XXX Should we divide cost by n_samples???
+    -elem_dot(&log_prob, y) + F::cast(0.5) * alpha * elem_dot(&params, &params)
```
bytesnake, line 507: because this is an estimate of the chained likelihood (and not MSE), I would not divide by the sample size

bytesnake (approved):
nice work :+1:

algorithms/linfa-logistic/src/argmin_param.rs
```diff
@@ -8,51 +8,60 @@
...
         ArgminParam(&self.0 + &other.0)
     }
 }
 
-impl<F: Float> ArgminDot<ArgminParam<F>, F> for ArgminParam<F> {
-    fn dot(&self, other: &ArgminParam<F>) -> F {
-        self.0.dot(&other.0)
+impl<F: Float, D: Dimension> ArgminDot<ArgminParam<F, D>, F> for ArgminParam<F, D> {
+    fn dot(&self, other: &ArgminParam<F, D>) -> F {
+        elem_dot(&self.0, &other.0)
```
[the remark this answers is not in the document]
bytesnake: I see, you can do `(&self.0 * &other.0).sum()`

algorithms/linfa-logistic/src/lib.rs
bytesnake: yes, but I think it's more intuitive if a fitted model can be constructed (i.e. having a `FittedLogisticRegression::new` function) with a given intercept and hyperplane. You can then update those initial weights with `FitWith` instead of putting them to the hyperparameters.

algorithms/linfa-logistic/src/lib.rs
bytesnake: on the other hand this is only true if the chain has always the same length otherwise the `alpha` would vary with the sample size

algorithms/linfa-logistic/src/argmin_param.rs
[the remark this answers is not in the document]
YuhanLiin: Wouldn't doing that create an intermediate array prior to summing?

algorithms/linfa-logistic/src/lib.rs
YuhanLiin: Both `FittedLogisticRegression` and `MultiFittedLogisticRegression` require the set of labels to be specified at construction, which isn't info that's available before fitting. If we make `FitWith` perform a single iteration then the users will have to write their own loops and stopping conditions, which is more work. We could make `FitWith` perform the complete algorithm to solve this. Also, if we have the `Fit` version of the algorithm always start with param arrays of 0s, it'd be pretty much useless since the optimal starting params are random values, not 0s. I just don't think it makes sense to delegate the offline and online versions of the algorithm to the same impl. What we can do is have separate param types for the online and offline versions of the algorithm, both implementing `FitWith`. The offline impl would run the entire algorithm while the online impl runs a single step. This makes sense because the offline algorithm requires a few hyperparams that the online version doesn't need.

PR #160, opened 2021-08-23
Extra Trees Implementation

mtandon18:
**Description**

This PR adds support for **Extremely Randomized Trees** (or Extra Trees for short) in `linfa-trees`. Here is a [link](https://quantdare.com/what-is-the-difference-between-extra-trees-and-random-forest/) detailing what Extra Trees are. Here is a summary of the main features:

1. Many trees w/all input observations used
2. Decision nodes consider only a subset of features and a random split for each feature. The best random split is assigned to the node.

**Primary Changes in Code**
- Updated the way fitting works in decision trees. First, not all features are considered at every decision node when `max_features` is not `None`. Additionally, I have broken out the code that finds the best split across all features and when the hyperparameter `random_split` is enabled, I skip finding the optimal split for a given feature. These latter changes were just incorporated into the existing for loop since there was very little additional code here.   

- Added fitting and prediction traits for `ExtraTrees` as well as the hyperparameters `ExtraTreesParams`. Fitting works by creating as many decision trees trees as specified by `num_estimators`. Predicting works by 

**Next Steps**
This is untested, except for one doc test (which passes). I will need to add a suite of tests in a follow up PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,8 +39,8 @@ num-traits = "0.2"
 rand = { version = "0.8", features = ["small_rng"] }
 approx = "0.4"
 
-ndarray = { version = "0.14", default-features = false, features = ["approx"] }
-ndarray-linalg = { version = "0.13", optional = true }
+ndarray = { version = "0.15", default-features = false, features = ["approx"] }
+ndarray-linalg = { version = "0.14", optional = true }
 
 thiserror = "=1.0.25"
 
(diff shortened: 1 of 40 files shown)


bytesnake:
did you intentionally close and force-push delete the content here? :grin:

mtandon18:
> did you intentionally close and force-push delete the content here? 😁

Yes, I did all that intentionally! This was accidentally submitted here but the goal is to follow up with a more polished implementation in a couple weeks.

PR #172, opened 2021-10-23
Remove linear regression normalization

YuhanLiin:
Remove normalization logic in `linfa-linear`, since that's handled in `linfa-preprocessing`.

Rewrite intercept fitting calculations.

### DIFF
--- a/algorithms/linfa-linear/src/ols.rs
+++ b/algorithms/linfa-linear/src/ols.rs
@@ -3,7 +3,6 @@
 use crate::error::{LinearError, Result};
 use ndarray::{concatenate, s, Array, Array1, Array2, ArrayBase, Axis, Data, Ix1, Ix2};
 use ndarray_linalg::{Lapack, LeastSquaresSvdInto, Scalar};
-use ndarray_stats::SummaryStatisticsExt;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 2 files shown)


algorithms/linfa-linear/src/ols.rs
```diff
@@ -267,7 +190,7 @@ mod tests {
     /// fit we only need to provide one point.
     #[test]
     fn without_intercept_fits_line_through_origin() {
-        let lin_reg = LinearRegression::new().with_intercept(false);
+        let lin_reg = LinearRegression::new().fit_intercept(false);
```
bytesnake, line 193: can you keep the original name here, writing `LinearRegression::new().fit_intercept(false).fit(..)` sounds weird

bytesnake (approved):
:+1: let's get rid of the `ndarray_stats` dependency here

algorithms/linfa-linear/src/ols.rs
YuhanLiin: Should I change it in logistic regression as well?

bytesnake: yes to keep the naming consistent

YuhanLiin: Turns out the name in logistic was already `with_intercept`

PR #177, opened 2021-11-20
Pin pdqselect dependency

YuhanLiin:
Bandaid fix for MSRV build to pass. Issue is with `pdqselect` 0.1.1 breaking older compilers.


ZuseZ4 (approved):
Thanks

Issue #179, opened 2021-11-23
The name of the plt file in linfa-tsne/examples is wrong


Alignof:
When I tried to run the code in linfa-tsne/examples as a tutorial, I got an error that the file does not exist.

```.zsh
$ git clone https://github.com/rust-ml/linfa.git
$ cd linfa/algorithms/linfa-tsne/
$ cargo run --example tsne --features linfa/intel-mkl-system
line 0: Cannot load input from 'examples/iris_plot.plt'.
````

Checking the corresponding directory, I found that only [mnist_plot.plt](https://github.com/rust-ml/linfa/blob/56c4d791054eb8e4705b31764561b023d74ed7d3/algorithms/linfa-tsne/examples/mnist_plot.plt) exists and its contents are for tsne.rs.
These are minor issues, but I'll report them for the sake of the project.

PR #180, opened 2021-11-23
Fix the plt files in linfa-tsne/examples

Alignof:
- Rename mnist_plot.plt to iris_plot.plt
- Add mnist_plotl.plt for mnist.rs


Alignof:
closes #179

algorithms/linfa-tsne/examples/mnist_plot.plt
```diff
@@ -1,7 +1,2 @@
-set style increment user
-set style line 1 lc rgb 'red'
-set style line 2 lc rgb 'blue'
-set style line 3 lc rgb 'green'
```
YuhanLiin, line 4: Can you include these 4 lines for `mnist_plot.plt` as well?

Alignof: OK. Fixed and amended.
It should be fixed correctly.

https://github.com/Alignof/linfa/commit/69ef8cb78973a47e0c6300725f6392b92cd0e5fa#diff-69a53e0c2347812bff90dd11e1ac389f7fc9793ac0b45deee06bd17f8186be05

YuhanLiin (approved):

PR #186, opened 2021-12-20
PLS Bug fix: PLS error when y constant

relf:
This PR handles constant residuals detection in singular value power method (which occurs if given training output is constant for instance). It raises a specific error which can be handled by the user.

### DIFF
--- a/algorithms/linfa-pls/src/pls_generic.rs
+++ b/algorithms/linfa-pls/src/pls_generic.rs
@@ -312,18 +312,14 @@ impl<F: Float> PlsValidParams<F> {
     ) -> Result<(Array1<F>, Array1<F>, usize)> {
         let eps = F::epsilon();
 
-        let mut y_score = Array1::ones(y.ncols());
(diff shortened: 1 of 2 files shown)


algorithms/linfa-pls/src/pls_generic.rs
```diff
@@ -313,12 +313,17 @@ impl<F: Float> PlsValidParams<F> {
         let eps = F::epsilon();
 
         let mut y_score = Array1::ones(y.ncols());
+        let mut found = false;
```
YuhanLiin, line 316: Instead of introducing `found`, you can make `y_score` into an `Option` that defaults to None and do `let y_score = y_score.ok_or(PlsError::PowerMethodConstantResidualError())?;`

relf: Sure! Defenitely the Rust way, thanks.

YuhanLiin (approved):

PR #189, opened 2022-01-09
Refactor `linfa-reduction` utility functions + move blob generation functions into `linfa-datasets`

YuhanLiin:
- Move `to_gaussian_similarity` into `linfa-nn` and parametrize it by the distance metric.
- Remove `generate_blobs`, since it already exists in `linfa-clustering`.


YuhanLiin:
One of the newest commits has the wrong commit message. Should instead be something like "move generate_blobs into linfa-dataset".

quietlychris (approved):
This all looks good to me!

Issue #183, opened 2021-11-29
Add Multinomial and Bernoulli Naive Bayes algorithms


sgrigory:
Currently `linfa-bayes` crate contains Gaussian Naive Bayes algorithm. It should not be very difficult to add other kinds of Naive Bayes present in `sklearn`:
- [x] Multinomial Naive Bayes
- [ ] Bernoulli Naive Bayes
 
For a new algorithm one needs to reimplement methods `joint_log_likelihood` and  `update_feature_log_prob` and the hyperparameters - the rest of the code stays more or less the same.

I have created a draft implementation of Multinomial Naive Bayes [in this branch](https://github.com/sgrigory/linfa/tree/add-multinomial-naive-bayes), based on the current code of Gaussian Naive Bayes and [the sklearn implementation of MultinomialNB](https://github.com/scikit-learn/scikit-learn/blob/0d378913be6d7e485b792ea36e9268be31ed52d0/sklearn/naive_bayes.py#L740). At the moment a large part of code is copy-pasted from Gaussian Naive Bayes, but it is possible to refactor both to deduplicate the shared code.

Would you consider this a useful feature to have? If yes, I can finalise the draft and open a PR .

@VasanthakumarV @bytesnake, tagging you since you authored and reviewed the original Gaussian Naive Bayes implementation in https://github.com/rust-ml/linfa/pull/51

bytesnake:
Hi @sgrigory, 

> It should not be very difficult to add other kinds of Naive Bayes present in sklearn

no not really, the question is rather how we want to add them. The type system should allow us to be generic over the distribution, there are some distribution libraries in Rust but few with MAP estimation.

> For a new algorithm one needs to reimplement methods joint_log_likelihood and update_feature_log_prob and the hyperparameters - the rest of the code stays more or less the same.

sounds like a good candidate for a trait

> I have created a draft implementation of Multinomial Naive Bayes in this branch, based on the current code of Gaussian Naive Bayes [..]

:+1:

> Would you consider this a useful feature to have? If yes, I can finalise the draft and open a PR .

yes, we would accept such a PR. To be really useful, we have to figure out how-to
1. handle mixed-type datasets
2. handle distributions with different parametrizations

It may therefore be refactored in the future, but nevertheless we will accept such a PR gladly :)

yuancc06:
[...]

PR #190, opened 2022-01-11
Add Multinomial Naive Bayes to linfa-bayes

sgrigory:
Closes "Multinomial Naive Bayes" part of #183

- Refactors `linfa-bayes` crate so that one can add different kinds of Naive Bayes without much code overlap
- Adds Multinomial Naive Bayes, in addition to Gaussian Naive Bayes which was already there. Implementation is inspired by `sklearn.naive_bayes.MultinomialNB` and it's used to compute the expected results for test cases

### Note on the type system

Initially I was planning to move the functionality which depends on the kind of Naive Bayes into a trait (let's call it `TraitNB` here), as suggested in #183, and implement this trait differently for Multinomial  and Gaussian NB. Then I would implement common NB functionality for any type T which implements `TraitNB`. However, I ran into a problem related to [specialization](https://github.com/rust-lang/rust/issues/31844): it seems impossible to implement traits like `Fit` for a generic type T, since it already has a generic implementation in [src/param_guard.rs](https://github.com/rust-ml/linfa/blob/b73a0fd9743a0df49baa251ba7707c36a964b693/src/param_guard.rs#L55). Please correct me if I'm missing some simple workaround here.

I therefore chose a different pattern: 
- Base types are enums - `base_nb::BaseNb` and `hyperparams::NbModel`
- Common functionality (e.g. `BaseNb::predict_inplace` method) is implemented for the whole enum
- Specific functionality is implemented for structures which correspond to the variants of enum, e.g. `GaussianNb::joint_log_likelihood` and `MultinomialNb::joint_log_likelihood`

I'm not sure this is the optimal solution and would be happy to hear your feedback on this.

### Note on dataset for the examples

[winequality](https://github.com/rust-ml/linfa/blob/master/datasets/data/winequality-red.csv.gz) doesn't seem to be the best dataset to illustrate the usage of Multinomial NB and it gives quite low accuracy in the example. Perhaps we could create a separate issue to add to `linfa-datasets` a dataset on which application of Multinomial NB would make sense, e.g. something like [20 Newsgroups data set](http://qwone.com/~jason/20Newsgroups/)?


algorithms/linfa-bayes/README.md
```diff
@@ -10,11 +10,12 @@
 
 `linfa-bayes` currently provides an implementation of the following methods: 
 
-- Gaussian Naive Bayes (GaussianNB)
+- Gaussian Naive Bayes ([`GaussianNb`](crate::GaussianNb))
+- Multinomial Naive Nayes ([`MultinomialNb`](crate::GaussianNb))
```
YuhanLiin, line 14: The link should say `MultinomialNb`

algorithms/linfa-bayes/src/base_nb.rs
```diff
@@ -0,0 +1,201 @@
...
+    // Returns a subset of x corresponding to the class specified by `ycondition`
+    pub fn filter(x: ArrayView2<F>, y: ArrayView1<L>, ycondition: &L) -> Array2<F> {
+        // We identify the row numbers corresponding to the class we are interested in
+        let index = y
+            .into_iter()
+            .enumerate()
+            .filter_map(|(i, y)| match *ycondition == *y {
+                true => Some(i),
+                false => None,
+            })
```
YuhanLiin, line 129: You can do `(*y_condition == *y).some().map(|_| i)`

YuhanLiin:
You can still use a Bayes trait to give you implementations for `predict_inplace` and `fit`. You just can't blanket impl for your trait, so you need to have individual impls of `PredictInplace` and `Fit` that act as thin wrappers. The trait can look like this:
```
trait NaiveBayes: FitWith { // FitWith impls have no common code, so they can be implemented for each individual type
  fn joint_log_likelihood() -> T;

  fn predict_inplace() {
    // call joint_log_likelihood
  }

  fn fit() {
    // call fit_with
  }
}
```
Then you impl `Fit` and `PredictInplace` for each NB type and just wrap the `NaiveBayes` methods with the same name, which minimizes amount of duplicated code. `filter` can just be a common function. Using a trait prevents needing to use an enum, which IMO adds unnecessary corner cases. This is especially true for the param type, where many methods panic when called with the wrong variant. The bonus is that the trait doesn't need to be public, since we don't blanket impl on top of it.

sgrigory:
@YuhanLiin Thanks for the detailed reply! That makes sense, let me try to refactor the code as you suggested and update this PR

bytesnake:
thank you for working on this

> Then I would implement common NB functionality for any type T which implements TraitNB. 

you can also introduce a trait `Distribution` (for loglikelihood estimation) and implement that for `univariate::{Gaussian, Multinomial}` structs containing the necessary statistics. The `NaiveBayes` algorithm would then be bounded to `Distribution` and contains a vector of that, one for each feature

algorithms/linfa-bayes/src/base_nb.rs
```diff
@@ -178,24 +46,50 @@ where
...
+}
+
+// Returns a subset of x corresponding to the class specified by `ycondition`
+pub fn filter<F: Float, L: Label + Ord>(
+    x: ArrayView2<F>,
+    y: ArrayView1<L>,
+    ycondition: &L,
+) -> Array2<F> {
+    // We identify the row numbers corresponding to the class we are interested in
+
```
YuhanLiin, line 80: Remove space

algorithms/linfa-bayes/src/base_nb.rs
```diff
@@ -3,152 +3,20 @@ use ndarray_stats::QuantileExt;
...
-#[derive(Debug, Clone)]
-pub enum BaseNb<F, L> {
-    MultinomialNb(MultinomialNb<F, L>),
-    GaussianNb(GaussianNb<F, L>),
-}
-
-
-impl<F, L, D, T> Fit<ArrayBase<D, Ix2>, T, NaiveBayesError> for NbValidParams<F, L>
+// Common functionality of fitted Naive Bayes models
+pub trait NaiveBayes<'a, F, L, D>
```
YuhanLiin, line 11: I don't think this trait and `NaiveBayesValidParams` need to be public.

YuhanLiin:
Can you move the examples on the param types to the main NB types? I'd rather have detailed docs on the main NB types than the param types. It's also a good replacement for the ignore examples we currently have for the NB types.

algorithms/linfa-bayes/src/base_nb.rs
```diff
@@ -48,8 +46,17 @@ where
...
 
+// Trait computing log-likelihood for Naive Bayes models
+pub(crate) trait NaiveBayesLogLikelihood<F, L>
+where
+    F: Float,
+    L: Label + Ord,
+{
+    fn joint_log_likelihood(&self, x: ArrayView2<F>) -> HashMap<&L, Array1<F>>;
+}
+
```
YuhanLiin, line 57: No need to have a separate trait for log likelihood.

sgrigory: @YuhanLiin Sure, I've moved log-likelihood back into the `NaiveBayes<'a, F, L, D>` trait. The only side effect is that now one needs to explicitly specify type parameter D when calling `joint_log_likelihood`, e.g. here:  https://github.com/rust-ml/linfa/pull/190/files#diff-20110e51d7148056af774f667a88af96f6abf6f9f3d3e79961e2098b6f11f441R261-R262. Compiler can't automatically infer type `D`  because `joint_log_likelihood` doesn't depend on it, but the whole `NaiveBayes` trait does.

YuhanLiin: In `NaiveBayes` can you move type parameter `D` from the trait to the `predict_inplace` method, since it's only relevant for that method? That should fix the type inference issues.

sgrigory: Yep, that worked, thanks

YuhanLiin (approved):

YuhanLiin:
Thanks for the PR!

sgrigory:
> Thanks for the PR!

Thanks for the guidance!

PR #194, opened 2022-01-19
Adding Multi-Task ElasticNet support

PABannier:
The goal of this PR is to add multi-task ElasticNet to the `elasticnet` crate.

A quick roadmap:

- [x] Write block coordinate descent
- [x] Write dual gap for multi-task 
- [x] Write tests for BCD
- [x] Make CD variable names consistent with variable names in BCD
- [x] Use a for loop when updating the residuals in CD
- [ ] Z-score + confidence level + variance
- [ ] Adapt Linfa ElasticNet API to the multi-task case
- [x] Write tests for ElasticNet multi-task


PABannier:
@YuhanLiin i'm implementing the `Fit` trait for the multi-task ENET. However, I don't know how to deal with Linfa API to handle the multi-task case. I created a `MultiTaskElasticNet` struct, but since both `ElasticNet` and `MultiTaskElasticNet` have the same set of parameters, I didn't create a `MultiTaskElasticNetValidParams`.

My question is how to restrict the trait bounds to implement the `Fit` trait for multi-task dataset?

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -56,6 +61,64 @@ where
     }
 }
 
+// impl<F, D, T> Fit<ArrayBase<D, Ix2>, T, ElasticNetError> for ElasticNetValidParams<F>
```
PABannier, line 64: @YuhanLiin If I uncomment this line, my compiler says that I'm redefining the `Fit` trait for `ElasticNetValidParams`.

YuhanLiin: Best way to resolve this is to change the `Fit` impl for single-task ElasticNet to `impl<F, D> Fit<ArrayBase<D, Ix2>, ArrayBase<D, Ix1>, ElasticNetError> for ElasticNetValidParams<F>`. This avoids conflicting between the two `Fit` impls on `ElasticNetValidParams`. The original `Fit` impl is only used for 1D outputs anyways so this change won't break anything. This way you won't need a multivariable version of the param struct.

src/dataset/mod.rs
```diff
@@ -198,6 +198,14 @@ pub trait Records: Sized {
...
 }
 
+/// MultiTask trait
+pub trait MultiTaskTarget: Sized {
+    type Elem;
+
+    fn nsamples(&self) -> usize;
+    fn ntasks(&self) -> usize;
+}
+
```
YuhanLiin, line 208: Why do you need to have this trait? Can't you just call `nrows()` and `ncols()` on 2D arrays directly?

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -56,6 +62,77 @@ where
...
+            &dataset.records(),
+            &hyperplane,
+            F::one(),
+            &mut y_est,
+        );
+        // Adding intercept
+        for i in 0..dataset.nsamples() {
+            let _res = &y_est.slice(s![i, ..]) + &intercept;
+            y_est.slice_mut(s![i, ..]).assign(&_res);
+        }
```
YuhanLiin, line 122: This can be better written as something like
```rust
for row in y_est.rows_mut() {
  row += &intercept;
}
```
You can even try just writing `y_est = y_est + &intercept`, since ndarray can broadcast 1D arrays into 2D arrays.

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -75,6 +152,28 @@ impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array1<F>> f
...
+{
+    /// Given an input matrix `X`, with shape `(n_samples, n_features)`,
+    /// `predict` returns the target variable according to elastic net
+    /// learned from the training data distribution.
+    fn predict_inplace(&self, x: &ArrayBase<D, Ix2>, y: &mut Array2<F>) {
+        assert_eq!(
+            x.nrows(),
+            y.nrows(),
+            "The number of data points must match the number of output targets."
+        );
```
YuhanLiin, line 166: You also need to assert that `x.ncols() == y.ncols()` right?

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -75,6 +152,28 @@ impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array1<F>> f
...
+
+        general_mat_mul(F::one(), &x, &self.hyperplane, F::one(), y);
+        // *y = *y + self.intercept;
+    }
+
+    fn default_target(&self, x: &ArrayBase<D, Ix2>) -> Array2<F> {
+        Array2::zeros((x.nrows(), x.nrows()))
```
YuhanLiin, line 173: Shouldn't it be `(x.nrows(), x.ncols())`

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -187,6 +345,88 @@ fn coordinate_descent<'a, F: Float>(
...
+}
+
+fn block_soft_thresholding<'a, F: Float>(x: ArrayView1<'a, F>, threshold: F) -> Array1<F> {
+    let norm_x = x.map(|&xi| xi.powi(2)).sum().sqrt();
+    let mut _res = Array1::<F>::zeros(x.len());
+    if norm_x >= threshold {
+        let scal = F::one() - threshold / norm_x;
+        for i in 0..x.len() {
+            _res[i] = scal * x[i];
+        }
```
YuhanLiin, line 425: This is equivalent to `res = x * scal;`

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -56,6 +62,77 @@ where
...
+        );
+
+        let mut y_est = Array2::<F>::zeros((dataset.nsamples(), dataset.ntargets()));
+        general_mat_mul(
+            F::one(),
+            &dataset.records(),
+            &hyperplane,
+            F::one(),
+            &mut y_est,
+        );
```
YuhanLiin, line 117: Equivalent to `let y_est = dataset.records() * hyperplane;`. This changed can be applied to most places where `general_mat_mul` is used.

algorithms/linfa-elasticnet/src/lib.rs
```diff
@@ -65,3 +68,39 @@ impl<F: Float> ElasticNet<F> {
...
+/// coefficients (a 2-dimensional array), (optionally) intercept (a 1-dimensional array), duality gaps
+/// and the number of steps needed in the computation.
+///
+/// ## Model implementation
+///
+/// The block coordinate descent is widely used to solve generalized linear models optimization problems,
+/// like Group Lasso, MultiTask Ridge or MultiTask Lasso. It cycles through a group of parameters and update
+/// the groups separately, holding all the others fixed. The optimization routine stops when a criterion is
+/// satisfied (dual sub-optimality gap or change in coefficients).
+pub struct MultiTaskElasticNet<F> {
```
YuhanLiin, line 84: Add the same `serde` derive that's on `ElasticNet`. Also can you add `#[derive(Debug, Clone, PartialEq)]` on both ElasticNet models?

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -153,25 +308,28 @@ fn coordinate_descent<'a, F: Float>(
...
-            let x_slc: ArrayView1<F> = x.slice(s![.., ii]);
-            if abs_diff_ne!(w_ii, F::zero()) {
-                // FIXME: direct addition with loop might be faster as it does not have to allocate
-                r += &(&x_slc * w_ii);
+            let old_w_j = w[j];
+            let x_j: ArrayView1<F> = x.slice(s![.., j]);
+            if abs_diff_ne!(old_w_j, F::zero()) {
+                for i in 0..x.shape()[0] {
+                    r[i] += x_j[i] * old_w_j;
+                }
```
YuhanLiin, line 320: Instead of adding in a loop you can do `r.scaled_add(w_ii, &x_slc)`. You can probably apply the same change in `block_coordinate_descent` as well.

YuhanLiin:
Is it possible to make functions like `coordinate_descent`, `duality_gap`, `variance_params`, and `compute_intercept` generic across 1D and 2D arrays? That way we won't need 2 sets of similar helper functions to handle the single-task and multi-task cases.

algorithms/linfa-elasticnet/src/algorithm.rs
YuhanLiin: If you want to stick with having 2 param types you can do something like:
```rust
pub struct ElasticNetParamsBase<const MULTI: bool> {
  ...
}

pub type ElasticNetParams = ElasticNetParamsBase<false>;
pub type MultiElasticNetParams = ElasticNetParamsBase<true>;
```
Then just put all the impls on `ElasticNetParamsBase` to avoid repeating everything.

src/dataset/mod.rs
PABannier: I removed it.

algorithms/linfa-elasticnet/src/algorithm.rs
PABannier: @YuhanLiin No. The number of columns in `x` corresponds to the number of features, while the number of columns in `y` is the number of tasks.

X_shape: (n_samples, n_features)
Y_shape: (n_samples, n_tasks)

algorithms/linfa-elasticnet/src/algorithm.rs
PABannier: It should actually be `(x.nrows(), y.ncols())`

PABannier:
@YuhanLiin Thanks for all the remarks! I passed your comments on the latest commit.

As for making `coordinate_descent` generic across 1D and 2D arrays, I don't think it is a good idea. `coordinate_descent` and `block_coordinate_descent` rely on two different proximal operators and there are more for loops in `block_coordinate_descent` specifically designed for the multi-task case. IMHO, a wiser choice would be to let `coordinate_descent` and `block_coordinate_descent` separate and use them as backbones for different penalties in the single task or multi task case. My intuition is that in future PR we could make it such that these two functions are generic over the regularization used in a model (for now it only supports L1 + L2), but there are more complex penalties that are very useful as well (non-convex ones for instance, see SCAD or MCP). 

`duality_gap` is also very specific in the single task or multi-task case and introducing a `match` syntax would make things more confusing IMHO.

For `variance_params` and `compute_intercept` I can certainly make it generic over 1D or 2D arrays.

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -153,25 +295,29 @@ fn coordinate_descent<'a, F: Float>(
...
-            let x_slc: ArrayView1<F> = x.slice(s![.., ii]);
-            if abs_diff_ne!(w_ii, F::zero()) {
-                // FIXME: direct addition with loop might be faster as it does not have to allocate
-                r += &(&x_slc * w_ii);
+            let old_w_j = w[j];
+            let x_j: ArrayView1<F> = x.slice(s![.., j]);
+            if abs_diff_ne!(old_w_j, F::zero()) {
+                // for i in 0..x.shape()[0] {
+                //     r[i] += x_j[i] * old_w_j;
+                // }
```
YuhanLiin, line 307: Delete comments

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -153,25 +295,29 @@ fn coordinate_descent<'a, F: Float>(
...
-                / (norm_cols_x[ii] + n_samples * (F::one() - l1_ratio) * penalty);
-            if abs_diff_ne!(w[ii], F::zero()) {
-                r -= &(&x_slc * w[ii]);
+            let tmp: F = x_j.dot(&r);
+            w[j] = tmp.signum() * F::max(tmp.abs() - n_samples * l1_ratio * penalty, F::zero())
+                / (norm_cols_x[j] + n_samples * (F::one() - l1_ratio) * penalty);
+            if abs_diff_ne!(w[j], F::zero()) {
+                for i in 0..x.shape()[0] {
+                    r[i] -= x_j[i] * w[j];
+                }
```
YuhanLiin, line 316: Can also use `scaled_add` here

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -187,6 +333,86 @@ fn coordinate_descent<'a, F: Float>(
...
+            }
+            let old_w_j: ArrayView1<F> = w.slice(s![j, ..]);
+            let x_j: ArrayView1<F> = x.slice(s![.., j]);
+            let norm_old_w_j = old_w_j.map(|wjt| wjt.powi(2)).sum().sqrt();
+            if abs_diff_ne!(norm_old_w_j, F::zero()) {
+                for i in 0..x.shape()[0] {
+                    for t in 0..n_tasks {
+                        r[[i, t]] += x_j[i] * old_w_j[t];
+                    }
+                }
```
YuhanLiin, line 372: Isn't this operation equivalent to `r += x_j.dot(old_w_j.t())`? If so you can replace these types of for loops with `general_mat_mul(1, x_j, old_w_j.t(), 1, r)`.

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -187,6 +333,86 @@ fn coordinate_descent<'a, F: Float>(
...
+            }
+        }
+    }
+
+    (w, gap, n_steps)
+}
+
+fn block_soft_thresholding<'a, F: Float>(x: ArrayView1<'a, F>, threshold: F) -> Array1<F> {
+    let norm_x = x.map(|&xi| xi.powi(2)).sum().sqrt();
+    let mut _res = Array1::<F>::zeros(x.len());
```
YuhanLiin, line 408: Why does the variable name start with _? Can you change it to just `res`?

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -56,6 +61,65 @@ where
...
+            y.view(),
+            self.tolerance(),
+            self.max_iterations(),
+            self.l1_ratio(),
+            self.penalty(),
+        );
+
+        // let y_est = dataset.records().dot(&hyperplane) + &intercept;
+        // let variance = variance_params_multi_task(dataset, y_est);
+        let variance = Ok(Array1::<F>::zeros(dataset.nfeatures()));
```
YuhanLiin, line 111: Testing purposes?

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -130,6 +216,62 @@ impl<F: Float> ElasticNet<F> {
...
+            .as_ref()
+            .map(|variance| {
+                self.hyperplane
+                    .iter()
+                    .zip(variance.iter())
+                    .map(|(a, b)| (*a - p * b.sqrt(), *a + p * b.sqrt()))
+                    .collect()
+            })
+            .map_err(|err| err.clone())
+    }
```
YuhanLiin, line 272: These 2 methods are identical between the 2 `ElasticNet`s, so they should be refactored out.

YuhanLiin:
Making `variance_params` and `compute_intercept` generic would be great.

algorithms/linfa-elasticnet/src/algorithm.rs
PABannier: This operation is equivalent to `np.outer` in Python (see: https://numpy.org/doc/stable/reference/generated/numpy.outer.html for a more detailed explanation). I don't think there is a built-in equivalent in ndarray. It is not yet available in the `ndarray` crate (see: https://github.com/rust-ndarray/ndarray/issues/1148) .

YuhanLiin: `outer(x, w)` is equivalent to `x` as a column vector multiplied with `w` as a row vector. The code in the `ndarray` PR you linked does the same thing. Using `general_mat_mul` allows you to add the matrix product of `x` and `w` to `r` in one operation. You do need to convert `x` and `y` into 2D arrays though, like in the PR.

YuhanLiin: Ideal way to convert 1D arrays into 2D is `insert_axis`. Something like `x.view().insert_axis(Axis(0_or_1))`

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -187,6 +306,88 @@ fn coordinate_descent<'a, F: Float>(
...
+    // the parameters of the model
+    let mut w = Array2::<F>::zeros((n_features, n_tasks));
+    // the residuals: `Y - XW` (since W=0, this is just `Y` for now),
+    // the residuals are updated during the algorithm as the parameters change
+    let mut r = y.to_owned();
+    let mut n_steps = 0u32;
+    let norm_cols_x = x.map_axis(Axis(0), |col| col.dot(&col));
+    let mut gap = F::one() + tol;
+    let d_w_tol = tol;
+    let tol = tol * y.fold(F::zero(), |sum, &y_ij| sum + y_ij.powi(2));
```
YuhanLiin, line 329: Instead of `fold` you can do `y.iter().map(|x| x*x).sum()`. Since `iter` is called before `map` it won't create a new array.  For 1D arrays it's even simpler since you can just call `y.dot(&y)` to dot product `y` with itself.

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -130,6 +215,66 @@ impl<F: Float> ElasticNet<F> {
...
+    /// Calculate the Z score
+    pub fn z_score(&self) -> Result<Array2<F>> {
+        self.variance
+            .as_ref()
+            .map(|variance| {
+                self.hyperplane
+                    .axis_iter(Axis(0))
+                    .zip(variance.iter())
+                    .map(|(a, b)| a.to_owned() / b.sqrt())
+                    .collect()
```
PABannier, line 251: @YuhanLiin My compiler is not happy when I try to build a 2d array from an iterator of 1d arrays: `a value of type ArrayBase<OwnedRepr<F>, Dim<[usize; 2]>> cannot be built from an iterator over elements of type ArrayBase<OwnedRepr<F>, Dim<[usize; 1]>>` . Do you know how to circumvent this issue?

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -130,6 +215,66 @@ impl<F: Float> ElasticNet<F> {
...
+        self.variance
+            .as_ref()
+            .map(|variance| {
+                self.hyperplane
+                    .axis_iter(Axis(1))
+                    .map(|coef| {
+                        coef.iter()
+                            .zip(variance.iter())
+                            .map(|(a, b)| (*a - p * b.sqrt(), *a + p * b.sqrt()))
+                            .collect()
```
PABannier, line 270: @YuhanLiin same issue here

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -217,22 +443,68 @@ fn duality_gap<'a, F: Float>(
...
+    };
+    let rty = r.t().dot(&y);
+    let trace_rty = rty.diag().sum();
+    let l21_norm = w.map_axis(Axis(1), |wj| (wj.dot(&wj)).sqrt()).sum();
+    gap += l1_reg * l21_norm - const_ * trace_rty
+        + half * l2_reg * (F::one() + const_ * const_) * w_norm2;
+    gap
+}
+
+fn variance_params<F: Float + Lapack, T: AsTargets<Elem = F>, D: Data<Elem = F>, I: Dimension>(
```
PABannier, line 480: This `AsTargets<Elem = F>` is very complex to deal with. I can't do much with it, since I need to have an `ArrayBase` in order to call `ndim()`, `shape()` and to make the computation `target - y_est`.

PABannier: Actually it should have some way to retrieve the dimension of the targets.

algorithms/linfa-elasticnet/src/algorithm.rs
YuhanLiin: You should also apply this change to all the other similar `fold` calls

algorithms/linfa-elasticnet/src/algorithm.rs
YuhanLiin: The typical iterator `map().collect()` pattern doesn't work on 2D arrays. Try `Zip::from(self.hyperplane).and_broadcast(variance).map_collect(|a, b| a / b.sqrt())`

algorithms/linfa-elasticnet/src/algorithm.rs
YuhanLiin: You can use the exact same logic here

algorithms/linfa-elasticnet/src/algorithm.rs
YuhanLiin: Currently `AsTargets` has the method `as_multi_targets`, which returns a 2D array view, so you can call it to retrieve the target for both cases. For the single target case this returns an array of dimension (n, 1). This means your code needs to treat single-task and multi-task-with-only-one-task as equivalent cases. `y_est` will need to be a 2D array in all cases; for single-task just insert `Axis(1)` into `y_est`.

After your other PR, this should be bounded with `AsMultiTargets` (now that I think about it, we need `AsMultiTargets` as a super-trait of `AsSingleTarget` for this to work).

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -353,9 +355,7 @@ fn block_coordinate_descent<'a, F: Float>(
             }
             let old_w_j: ArrayView1<F> = w.slice(s![j, ..]);
             let x_j: ArrayView1<F> = x.slice(s![.., j]);
-            let norm_old_w_j = old_w_j
-                .fold(F::zero(), |sum, &wjt| sum + wjt.powi(2))
-                .sqrt();
+            let norm_old_w_j = old_w_j.map(|wjt| wjt.powi(2)).sum().sqrt();
```
YuhanLiin, line 358: Use `old_w_j.dot(&old_w_j)` instead

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -366,10 +366,7 @@ fn block_coordinate_descent<'a, F: Float>(
                 &(block_soft_thresholding(tmp.view(), n_samples * l1_ratio * penalty)
                     / (norm_cols_x[j] + n_samples * (F::one() - l1_ratio) * penalty)),
             );
-            let norm_w_j = w
-                .slice(s![j, ..])
-                .fold(F::zero(), |sum, &wjt| sum + wjt.powi(2))
-                .sqrt();
+            let norm_w_j = w.slice(s![j, ..]).map(|wjt| wjt.powi(2)).sum().sqrt();
```
YuhanLiin, line 369: Call `dot` here as well

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -397,7 +394,7 @@ fn block_coordinate_descent<'a, F: Float>(
 }
 
 fn block_soft_thresholding<'a, F: Float>(x: ArrayView1<'a, F>, threshold: F) -> Array1<F> {
-    let norm_x = x.fold(F::zero(), |sum, &x_i| sum + x_i * x_i).sqrt();
+    let norm_x = x.map(|x| x.powi(2)).sum().sqrt();
```
YuhanLiin, line 397: Use `dot`

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -429,7 +426,7 @@ fn duality_gap<'a, F: Float>(
     } else {
         (F::one(), r_norm2)
     };
-    let l1_norm = w.fold(F::zero(), |sum, w_i| sum + w_i.abs());
+    let l1_norm = w.map(|w_i| w_i.abs()).sum();
```
YuhanLiin, line 429: Use `dot`

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -452,8 +449,8 @@ fn duality_gap_mtl<'a, F: Float>(
     let dual_norm_xta = xta
         .map_axis(Axis(1), |x| x.dot(&x).sqrt())
         .fold(F::zero(), |max_norm, &nrm| max_norm.max(nrm));
-    let r_norm2 = r.fold(F::zero(), |sum, &rij| sum + rij.powi(2));
-    let w_norm2 = w.fold(F::zero(), |sum, &wij| sum + wij.powi(2));
+    let r_norm2 = r.map(|rij| rij.powi(2)).sum();
+    let w_norm2 = w.map(|wij| wij.powi(2)).sum();
```
YuhanLiin, line 453: Call `iter()` before `map` to prevent creating a new array

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -580,7 +577,7 @@ mod tests {
         let mut resid = x.dot(beta);
         resid = &resid * -1.;
         resid = &resid - intercept + y;
-        let mut datafit = resid.fold(0., |sum, &r| sum + r.powi(2));
+        let mut datafit = resid.map(|rij| rij.powi(2)).sum();
```
YuhanLiin, line 580: Add `iter` here as well

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -594,7 +591,7 @@ mod tests {
     }
 
     fn elastic_net_mtl_penalty(beta: &Array2<f64>, alpha: f64) -> f64 {
-        let frob_norm = beta.fold(0., |sum, &beta_ij| sum + beta_ij.powi(2));
+        let frob_norm = beta.map(|beta_ij| beta_ij.powi(2)).sum();
```
YuhanLiin, line 594: Add `iter` here

PABannier:
Since #206 has been merged, ElasticNet is now easier to adapt to the multi-task case. I'm still working on it. 
Roadmap before merging:

- [ ] Make `variance_params` generic for single task and multi-task
- [ ] Make `compute_intercept` generic for single task and multi-task
- [ ] Write tests for MultiTaskElasticNet (Lasso + Ridge...)

YuhanLiin:
Work continued in #238

relf:
I am closing as superseded by #238

Issue #195, opened 2022-01-23
Split `AsTarget` and `AsTargetMut` into 1D and 2D traits


YuhanLiin:
`AsTarget` and `AsTargetMut` support return 1D and 2D targets. However, most uses of the traits require either 1D targets or 2D targets, but never both. Also, some impls don't support 1D targets, so `try_single_target()` has to return a `Result` to handle that case. I propose that we split `AsTarget` into `AsSingleTarget` and `AsMultiTarget` (same with `AsTargetMut`) and implement them on `ArrayView1` and `ArrayView2` respectively. Trait bounds will be changed to use one or the other. What do you think @bytesnake ?

PABannier:
I think that would be of great help! I can work on it. Actually without that kind of tweak, I'm stuck with #194 .

YuhanLiin:
Closed by #206

PR #203, opened 2022-02-20
Split `AsTargets` into `AsSingleTargets` and `AsMultiTargets` 

PABannier:
**1) What was done before?**  `linfa` offers one trait `AsTargets` which supports both the single-task and the multi-task case.

**2) What does this PR change?** The goal of this PR is to split `AsTargets` into `AsSingleTargets` and `AsMultiTargets`, add the corresponding implementations and integrate it to `linfa` codebase.

**3) Why is it better**? As explained in #195 , splitting `AsTargets` into `AsSingleTargets` and `AsMultiTargets` would be greatly beneficial to implement multi-task models.

### DIFF
--- a/algorithms/linfa-svm/src/regression.rs
+++ b/algorithms/linfa-svm/src/regression.rs
@@ -153,10 +153,6 @@ macro_rules! impl_regression {
     };
 }
 
-// impl_regression!(Array2<f32>, Array2<f32>, f32);
-// impl_regression!(Array2<f64>, Array2<f64>, f64);
-// impl_regression!(ArrayView2<'_, f32>, ArrayView2<'_, f32>, f32);
-// impl_regression!(ArrayView2<'_, f64>, ArrayView2<'_, f64>, f64);
 impl_regression!(Array2<f32>, Array1<f32>, f32);
 impl_regression!(Array2<f64>, Array1<f64>, f64);
 impl_regression!(ArrayView2<'_, f32>, ArrayView1<'_, f32>, f32);
--- a/algorithms/linfa-svm/src/regression.rs
+++ b/algorithms/linfa-svm/src/regression.rs
@@ -1,6 +1,6 @@
 //! Support Vector Regression
 use linfa::{
-    dataset::{AsTargets, DatasetBase},
+    dataset::{AsSingleTargets, DatasetBase},
     traits::Fit,
     traits::Transformer,
     traits::{Predict, PredictInplace},
@@ -153,10 +153,10 @@ macro_rules! impl_regression {
     };
 }
 
-impl_regression!(Array2<f32>, Array2<f32>, f32);
-impl_regression!(Array2<f64>, Array2<f64>, f64);
-impl_regression!(ArrayView2<'_, f32>, ArrayView2<'_, f32>, f32);
-impl_regression!(ArrayView2<'_, f64>, ArrayView2<'_, f64>, f64);
+// impl_regression!(Array2<f32>, Array2<f32>, f32);
+// impl_regression!(Array2<f64>, Array2<f64>, f64);
+// impl_regression!(ArrayView2<'_, f32>, ArrayView2<'_, f32>, f32);
+// impl_regression!(ArrayView2<'_, f64>, ArrayView2<'_, f64>, f64);
 impl_regression!(Array2<f32>, Array1<f32>, f32);
 impl_regression!(Array2<f64>, Array1<f64>, f64);
 impl_regression!(ArrayView2<'_, f32>, ArrayView1<'_, f32>, f32);
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -244,6 +244,8 @@ impl<L, R: Records, T: AsMultiTargets<Elem = L>> AsMultiTargets for DatasetBase<
     }
 }
 
+impl<L, R: Records, T: AsSingleTargets<Elem = L>> AsSingleTargets for DatasetBase<R, T> {}
+
 impl<L, R: Records, T: AsMultiTargetsMut<Elem = L>> AsMultiTargetsMut for DatasetBase<R, T> {
     type Elem = L;
 
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -25,7 +25,7 @@ impl<R: Records, S> DatasetBase<R, S> {
     /// ```ignore
     /// let dataset = Dataset::new(records, targets);
     /// ```
-    pub fn new<T: IntoTargets<S>>(records: R, targets: T) -> DatasetBase<R, S> {
+    pub fn new(records: R, targets: S) -> DatasetBase<R, S> {
         let targets = targets.into();
 
         DatasetBase {
@@ -1116,21 +1116,5 @@ impl<L: Label, S: Labels<Elem = L>> CountedTargets<L, S> {
         let labels = targets.label_count();
 
         CountedTargets { targets, labels }
-    }
-}
-
-pub trait IntoTargets<T> {
-    fn into(self) -> T;
-}
-
-// impl<F, D: Data<Elem = F>> IntoTargets<ArrayBase<D, Ix2>> for ArrayBase<D, Ix1> {
-//     fn into(self) -> ArrayBase<D, Ix2> {
-//         self.insert_axis(Axis(1))
-//     }
-// }
-
-impl<T> IntoTargets<T> for T {
-    fn into(self) -> T {
-        self
     }
 }
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -795,87 +795,6 @@ where
         objs.into_iter().zip(self.sample_chunks(fold_size))
     }
 
-    /// Cross validation for multi-target algorithms
-    ///
-    /// Given a list of fittable models, cross validation
-    /// is used to compare their performance according to some
-    /// performance metric. To do so, k-folding is applied to the
-    /// dataset and, for each fold, each model is trained on the training set
-    /// and its performance is evaluated on the validation set. The performances
-    /// collected for each model are then averaged over the number of folds.
-    ///
-    /// ### Parameters:
-    ///
-    /// - `k`: the number of folds to apply
-    /// - `parameters`: a list of models to compare
-    /// - `eval`: closure used to evaluate the performance of each trained model
-    ///
-    /// ### Returns
-    ///
-    /// An array of model performances, in the same order as the models in input, if no errors occur.
-    /// The performance of each model is given as an array of performances, one for each target.
-    /// Otherwise, it might return an Error in one of the following cases:
-    ///
-    /// - An error occurred during the fitting of one model
-    /// - An error occurred inside the evaluation closure
-    ///
-    /// ### Example
-    ///
-    /// ```rust, ignore
-    ///
-    /// use linfa::prelude::*;
-    ///
-    /// // mutability needed for fast cross validation
-    /// let mut dataset = linfa_datasets::diabetes();
-    ///
-    /// let models = vec![model1, model2, ... ];
-    ///
-    /// let r2_scores = dataset.cross_validate_multi(5,&models, |prediction, truth| prediction.r2(truth))?;
-    ///
-    /// ```
-    pub fn cross_validate_multi<O, ER, M, FACC, C>(
-        &'a mut self,
-        k: usize,
-        parameters: &[M],
-        eval: C,
-    ) -> std::result::Result<Array2<FACC>, ER>
-    where
-        ER: std::error::Error + std::convert::From<crate::error::Error>,
-        M: for<'c> Fit<ArrayView2<'c, F>, ArrayView2<'c, E>, ER, Object = O>,
-        O: for<'d> PredictInplace<ArrayView2<'a, F>, Array2<E>>,
-        FACC: Float,
-        C: Fn(&Array2<E>, &ArrayView2<E>) -> std::result::Result<Array1<FACC>, crate::error::Error>,
-    {
-        let mut evaluations = Array2::from_elem((parameters.len(), self.ntargets()), FACC::zero());
-        let folds_evaluations: std::result::Result<Vec<_>, ER> = self
-            .iter_fold(k, |train| {
-                let fit_result: std::result::Result<Vec<_>, ER> =
-                    parameters.iter().map(|p| p.fit(train)).collect();
-                fit_result
-            })
-            .map(|(models, valid)| {
-                let targets = valid.targets();
-                let models = models?;
-                let mut eval_predictions =
-                    Array2::from_elem((models.len(), targets.len()), FACC::zero());
-                for (i, model) in models.iter().enumerate() {
-                    let predicted = model.predict(valid.records());
-                    let eval_pred = match eval(&predicted, targets) {
-                        Err(e) => Err(ER::from(e)),
-                        Ok(res) => Ok(res),
-                    }?;
-                    eval_predictions.row_mut(i).add_assign(&eval_pred);
-                }
-                Ok(eval_predictions)
-            })
-            .collect();
-
-        for fold_evaluation in folds_evaluations? {
-            evaluations.add_assign(&fold_evaluation)
-        }
-        Ok(evaluations / FACC::from(k).unwrap())
-    }
-
     /// Cross validation for single target algorithms
     ///
     /// Given a list of fittable models, cross validation
@@ -929,7 +848,10 @@ where
         M: for<'c> Fit<ArrayView2<'c, F>, ArrayView2<'c, E>, ER, Object = O>,
         O: for<'d> PredictInplace<ArrayView2<'a, F>, ArrayBase<OwnedRepr<E>, I>>,
         FACC: Float,
-        C: Fn(&ArrayView1<E>, &ArrayView1<E>) -> std::result::Result<FACC, crate::error::Error>,
+        C: Fn(
+            &ArrayView2<E>,
+            &ArrayView2<E>,
+        ) -> std::result::Result<Array1<FACC>, crate::error::Error>,
         I: Dimension,
     {
         // construct shape as either vector or matrix
@@ -973,14 +895,21 @@ where
                         let predicted: Array2<_> =
                             predicted.into_shape((nsamples, ntargets)).unwrap();
 
-                        predicted
-                            .columns()
-                            .into_iter()
-                            .zip(targets.columns().into_iter())
-                            .map(|(p, t)| eval(&p.view(), &t).map_err(ER::from))
-                            .collect()
+                        let eval_pred = match eval(&predicted.view(), &targets.view()) {
+                            Err(e) => Err(ER::from(e)),
+                            Ok(res) => Ok(res),
+                        }
+                        .unwrap();
+
+                        eval_pred
+                        // predicted
+                        //     .columns()
+                        //     .into_iter()
+                        //     .zip(targets.columns().into_iter())
+                        //     .map(|(p, t)| eval(&p.view(), &t).map_err(ER::from))
+                        //     .collect()
                     })
-                    .collect::<std::result::Result<Vec<Vec<FACC>>, ER>>()?
+                    .collect::<Vec<Array1<_>>>()
                     .into_iter()
                     .flatten()
                     .collect();
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -123,7 +123,7 @@ impl<R: Records, S> DatasetBase<R, S> {
     }
 }
 
-impl<L, R: Records, T: AsSingleTargets<Elem = L>> DatasetBase<R, T> {
+impl<L, R: Records, T: AsMultiTargets<Elem = L>> DatasetBase<R, T> {
     /// Map targets with a function `f`
     ///
     /// # Example
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -424,7 +424,7 @@ where
 }
 
 impl<F, E, D, S> From<(ArrayBase<D, Ix2>, ArrayBase<S, Ix1>)>
-    for DatasetBase<ArrayBase<D, Ix2>, ArrayBase<S, Ix2>>
+    for DatasetBase<ArrayBase<D, Ix2>, ArrayBase<S, Ix1>>
 where
     D: Data<Elem = F>,
     S: Data<Elem = E>,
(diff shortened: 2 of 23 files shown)


src/dataset/impl_dataset.rs
```diff
@@ -123,7 +123,7 @@ impl<R: Records, S> DatasetBase<R, S> {
     }
 }
 
-impl<L, R: Records, T: AsTargets<Elem = L>> DatasetBase<R, T> {
+impl<L, R: Records, T: AsSingleTargets<Elem = L>> DatasetBase<R, T> {
```
PABannier, line 126: @YuhanLiin The implementation for `T: AsMultiTargets<Elem = L>` would be very similar. What do you recommend to avoid duplicating very similar code?

src/dataset/mod.rs
```diff
@@ -222,6 +222,17 @@ pub trait AsTargets {
     }
```
YuhanLiin, line 222: This trait should only have the method `as_single_target()` and should only be implemented for 1D datasets. Same with the mutable version of the trait.

src/dataset/impl_targets.rs
```diff
@@ -1,22 +1,54 @@
...
     }
 }
 
+impl<'a, L, S: Data<Elem = L>> AsSingleTargets for ArrayBase<S, Ix2> {
+    type Elem = L;
+
+    fn as_multi_targets(&self) -> ArrayView2<L> {
+        self.view()
+    }
+}
```
YuhanLiin, line 26: This impl shouldn't exist

src/dataset/impl_targets.rs
```diff
@@ -30,57 +62,109 @@ impl<'a, L: Clone + 'a, S: Data<Elem = L>> FromTargetArray<'a, L> for ArrayBase<
...
-impl<L, S: Data<Elem = L>> AsTargets for ArrayBase<S, Ix2> {
+impl<L, S: DataMut<Elem = L>> AsSingleTargetsMut for ArrayBase<S, Ix2> {
     type Elem = L;
 
-    fn as_multi_targets(&self) -> ArrayView2<L> {
-        self.view()
+    fn as_multi_targets_mut(&mut self) -> ArrayViewMut2<'_, Self::Elem> {
+        self.view_mut()
     }
 }
```
YuhanLiin, line 79: This shouldn't exist either

src/dataset/impl_targets.rs
```diff
@@ -30,57 +62,109 @@ impl<'a, L: Clone + 'a, S: Data<Elem = L>> FromTargetArray<'a, L> for ArrayBase<
...
+
+    fn ntargets(&self) -> usize {
+        self.targets.ntargets()
+    }
 }
 
 impl<'a, L: Label + 'a, T> FromTargetArray<'a, L> for CountedTargets<L, T>
 where
-    T: AsTargets<Elem = L> + FromTargetArray<'a, L>,
+    T: AsSingleTargets<Elem = L> + FromTargetArray<'a, L>,
```
YuhanLiin, line 167: Is the `AsSingleTargets` bound necessary here?

src/dataset/impl_targets.rs
```diff
@@ -152,7 +236,7 @@ impl<L: Label, S: Data<Elem = L>, I: Dimension> Labels for ArrayBase<S, I> {
 }
 
 /// Counted labels can act as labels
-impl<L: Label, T: AsTargets<Elem = L>> Labels for CountedTargets<L, T> {
+impl<L: Label, T: AsSingleTargets<Elem = L>> Labels for CountedTargets<L, T> {
```
YuhanLiin, line 239: This `AsSingleTargets` bound isn't needed

src/dataset/impl_targets.rs
```diff
@@ -163,7 +247,7 @@ impl<L: Label, T: AsTargets<Elem = L>> Labels for CountedTargets<L, T> {
 impl<F: Copy, L: Copy + Label, D, T> DatasetBase<ArrayBase<D, Ix2>, T>
 where
     D: Data<Elem = F>,
-    T: AsTargets<Elem = L>,
+    T: AsSingleTargets<Elem = L>,
```
YuhanLiin, line 250: This bound should be `AsMultiTargets`

src/metrics_clustering.rs
```diff
@@ -62,8 +62,13 @@ impl<F: Float> DistanceCount<F> {
...
+impl<
+        'a,
+        F: Float,
+        L: 'a + Label,
+        D: Data<Elem = F>,
+        T: AsSingleTargets<Elem = L> + Labels<Elem = L>,
+    > SilhouetteScore<F> for DatasetBase<ArrayBase<D, Ix2>, T>
 {
     fn silhouette_score(&self) -> Result<F> {
         if self.ntargets() > 1 {
```
YuhanLiin, line 74: Maybe we can delete the error variant below this line

src/metrics_regression.rs
```diff
@@ -137,7 +139,9 @@ impl<F: Float, D: Data<Elem = F>, T: AsTargets<Elem = F>> SingleTargetRegression
 /// The shape of the compared targets must match.
 ///
 /// To compare single-dimensional arrays use [`SingleTargetRegression`](trait.SingleTargetRegression.html)
-pub trait MultiTargetRegression<F: Float, T: AsTargets<Elem = F>>: AsTargets<Elem = F> {
+pub trait MultiTargetRegression<F: Float, T: AsSingleTargets<Elem = F>>:
+    AsSingleTargets<Elem = F>
```
YuhanLiin, line 143: This should be multitarget

src/dataset/impl_targets.rs
```diff
@@ -1,22 +1,54 @@
...
+    type Elem = L;
+
+    fn as_multi_targets(&self) -> ArrayView2<L> {
+        self.view().insert_axis(Axis(1))
+    }
+
+    fn ntargets(&self) -> usize {
+        1
+    }
+}
```
YuhanLiin, line 38: I don't know if this is possible, but maybe you can do a blanket impl like `impl AsMultiTargets for T where T: AsSingleTargets` instead.

src/dataset/impl_dataset.rs
YuhanLiin: Use `AsMultiTargets` instead of `AsSingleTarget` as the bound for these dataset impls. `AsMultiTarget` is more general than `AsSingleTarget`, and even without a blanket impl, types that implement `AsSingleTarget` probably also implement `AsMultiTarget`.

src/dataset/impl_targets.rs
YuhanLiin: Actually, scratch that. We should make `AsMultiTargets` as supertrait of `AsSingleTarget`. Maybe `AsMultiTargets` should have its name changed back to `AsTargets`, since it encapsulates both single and multi-target data.

PABannier: Indeed, I think making `AsMultiTargets` a supertrait of `AsSingleTarget` would be helpful. I would opt for a clearer naming however, by keeping `AsSingleTarget` and `AsMultiTarget`. We should just precise in the doc that `AsSingleTarget` is simply an edge case of `AsMultiTarget` with `ntarget = 1`.

YuhanLiin: That makes sense

src/dataset/mod.rs
```diff
@@ -247,12 +249,17 @@ pub trait FromTargetArray<'a, F> {
...
-pub trait AsSingleTargetsMut {
+pub trait AsMultiTargetsMut {
     type Elem;
 
-    /// Returns a mutable view on targets as two-dimensional array
+    /// Convert to a multi-target
     fn as_multi_targets_mut(&mut self) -> ArrayViewMut2<Self::Elem>;
 
+    /// Returns the number of targets
+    fn ntargets(&mut self) -> usize;
```
YuhanLiin, line 259: No need for `mut` here

src/metrics_regression.rs
```diff
@@ -139,7 +139,7 @@ impl<F: Float, D: Data<Elem = F>, T: AsSingleTargets<Elem = F>> SingleTargetRegr
 /// The shape of the compared targets must match.
 ///
 /// To compare single-dimensional arrays use [`SingleTargetRegression`](trait.SingleTargetRegression.html)
-pub trait MultiTargetRegression<F: Float, T: AsSingleTargets<Elem = F>>:
+pub trait MultiTargetRegression<F: Float, T: AsMultiTargets<Elem = F>>:
     AsSingleTargets<Elem = F>
```
YuhanLiin, line 143: This should also be changed to multi target

src/dataset/mod.rs
```diff
@@ -198,13 +198,26 @@ pub trait Records: Sized {
...
+/// AsMultiTargets trait
+///
+/// Provides useful methods to handle target variables. Note that AsMultiTargets
+/// is a supertrait of AsSingleTargets. AsMultiTargets is specifically designed
+/// to handle multi-target variables.
+pub trait AsMultiTargets {
     type Elem;
 
-    /// Returns a view on targets as two-dimensional array
+    /// Convert to a multi-target
```
YuhanLiin, line 209: Why the documentation change?

src/dataset/mod.rs
YuhanLiin: The `as_single_target` method shouldn't be fallible, so it won't have a default impl. Otherwise if we mark the wrong type with this trait we'd get panics. This trait should only be implemented on "single-target" types, such as `Array1`.

src/dataset/impl_targets.rs
```diff
@@ -49,6 +33,10 @@ impl<'a, L, S: Data<Elem = L>> AsMultiTargets for ArrayBase<S, Ix2> {
     }
 }
 
+impl<'a, L, S: Data<Elem = L>> AsSingleTargets for ArrayBase<S, Ix1> {}
+
+impl<'a, L, S: Data<Elem = L>> AsSingleTargets for ArrayBase<S, Ix2> {}
```
YuhanLiin, line 38: Implementing `AsSingleTargets` for 2D arrays makes no sense

src/dataset/impl_targets.rs
```diff
@@ -62,54 +50,18 @@ impl<'a, L: Clone + 'a, S: Data<Elem = L>> FromTargetArray<'a, L> for ArrayBase<
...
-    type Elem = L;
-
-    fn as_multi_targets_mut(&mut self) -> ArrayViewMut2<'_, Self::Elem> {
-        self.view_mut().insert_axis(Axis(1))
-    }
-
-    fn ntargets(&self) -> usize {
-        1
-    }
-}
```
YuhanLiin, line 47: This impl should stay, so we can implement `AsSingleTargetMut` for 1D arrays.

PABannier:
@YuhanLiin I'm changing `AsTargets` into `AsSingleTargets` and `AsMultiTargets` in `linfa` crates. I'm having a lot of compiler errors in the tests because of the object `DatasetView`, which implicitly converts a 1d array into a `2d array`. Since 2d arrays do not implement the `AsSingleTargets` trait, and we try to pass a `DatasetView` object to a function `fit` that expects a struct that implements the `AsSingleTargets`, we have compiler errors. For instance, see `GaussianNB` tests.

Since we don't want to implement `AsSingleTargets` for 2d arrays, does that mean we should create a `SingleDatasetView` or something like this? What do you recommend?

PABannier:
@YuhanLiin Actually the issue is that `Dataset` and `DatasetView` automatically creates a multi-target vector (2d array), which prohibits the use of the trait `AsSingleTargets`. 

I'd be tempted to create  a `SingleTargetDataset` and `MultiTargetDataset` types, but the naming is cumbersome and it's change with far-reaching consequences. One option would be to keep `Dataset` as the default option with 2d array as the target, and to create a `SingleDataset` type which takes in a 1d array as the target. But in this case, to be consistent with the naming, we should rename `AsMultiTargets` into `AsTargets` as you suggested above.

WDYT?

YuhanLiin:
Try this: change the definition of `Dataset` into `pub type Dataset<D, T, I = Ix2> = DatasetBase<ArrayBase<OwnedRepr<D>, Ix2>, ArrayBase<OwnedRepr<T>, I>>;` Do the same for `DatasetView`. Adding an extra type param for the dimensionality of targets allows the type to be used for both 1D and 2D targets, and the default value of `Ix2` minimizes breakage. After that all you need to do is change the single-target dataset examples (such as `GaussianNB`) to use 1D instead of 2D arrays.

algorithms/linfa-svm/src/regression.rs
```diff
@@ -153,10 +153,10 @@ macro_rules! impl_regression {
     };
 }
 
-impl_regression!(Array2<f32>, Array2<f32>, f32);
-impl_regression!(Array2<f64>, Array2<f64>, f64);
-impl_regression!(ArrayView2<'_, f32>, ArrayView2<'_, f32>, f32);
-impl_regression!(ArrayView2<'_, f64>, ArrayView2<'_, f64>, f64);
+// impl_regression!(Array2<f32>, Array2<f32>, f32);
+// impl_regression!(Array2<f64>, Array2<f64>, f64);
```
PABannier, line 157: I commented out these tests, since the macro calls `try_single_target`. Should we remove them?

src/dataset/impl_dataset.rs
```diff
@@ -1113,11 +1123,11 @@ pub trait IntoTargets<T> {
     fn into(self) -> T;
 }
 
-impl<F, D: Data<Elem = F>> IntoTargets<ArrayBase<D, Ix2>> for ArrayBase<D, Ix1> {
-    fn into(self) -> ArrayBase<D, Ix2> {
-        self.insert_axis(Axis(1))
-    }
-}
+// impl<F, D: Data<Elem = F>> IntoTargets<ArrayBase<D, Ix2>> for ArrayBase<D, Ix1> {
```
PABannier, line 1126: I don't think we need this impl anymore. It converts Array1d into Array2d in `Dataset`, which creates conflicting implementations in the tests.

PABannier: This change has a lot of consequences. A lot of 1d targets were implicitly converted to Array2d, which is expected for functions like `cross_validate` to work. For `DatasetBase<ArrayBase<D, Ix2>, ArrayBase<S, Ix2>>`, there are `cross_validate` and `cross_validate_multi`. IMHO, we should remove `cross_validate_multi` from the trait and leave `cross_validate`, then we implement `cross_validate` for `DatasetBase<ArrayBase<D, Ix2>, ArrayBase<S, Ix1>>`. WDYT?

algorithms/linfa-svm/src/regression.rs
YuhanLiin: I'm pretty sure those are impls used by the rest of the library, not tests. This fitting algorithm is clearly single-target, so just remove the impls with 2D targets and leave the other ones.

src/dataset/impl_dataset.rs
YuhanLiin: That whole trait should be removed. The fact that it has `into` as a method is a red flag, since that makes it a worse version of `Into`. From what I understand the problem you're having currently is that, because most `Dataset` methods were implemented on 2D targets, they won't the compile now that we have datasets with 1D targets. For methods that are applicable for both 1D and 2D targets (most methods are like this), place them in a generic impl block bound by `AsMultiTargets`, since `as_multi_targets` normalizes 1D and 2D targets into 2D arrays. Methods applicable only for 1D targets should be implemented for `AsSingleTarget` if possible; otherwise implement for `ArrayBase<S, Ix1>`. Methods applicable only for 2D targets should be implemented for `AsMultiTargets` if possible; otherwise implement for `ArrayBase<S, Ix2>`. This strategy should minimize caller breakage.

For `cross_validate` and `cross_validate_multi`, I'm pretty sure that `cross_validate` applies to both 1D and 2D targets, while `cross_validate_multi` applies only to 2D and works a bit differently. `multi` allows the evaluation function to differentiate between different targets, since its evaluation function takes `Array2` instead of `Array1`. However, it's possible to express the semantics of 2D `cross_validate` in terms of `cross_validate_multi`, so we don't really need 2 functions. All we really need is one generic `cross_validate` function that covers both 1D and 2D targets, and it'll behave like `cross_validate_multi` in the 2D case. This should be easily doable by making the evaluation closure generic on the target dimensionality.

PABannier: It is well noted. I'm left confused by the `FromTargetArray` trait. It is not implemented for Array1d. However, `bootstrap`, `bootstrap_samples`, etc. are implemented with the `FromTargetArray` and `AsMultiTargets` traits. I only see one implementation of `FromTargetArray` for `CountedTargets` but none for Array1d and Array2d. I don't know how to deal with it.

YuhanLiin: `FromTargetArray` acts like the reverse of `AsMultipleTargets`. It takes a 2D target array and converts it back into the original "target type", such as `CountedTargets`. If implemented on `Array2` it converts the target array into itself (identity operation pretty much). If implemented on `CountedTargets` it wraps the target array into a `CountedTargets`. Typically, a function like `bootstrap` takes some targets, calls `as_multi_targets` on it, manipulates the 2D target array, then calls `from_target` to turn it back into the original target type. This doesn't work on 1D targets right now, since `FromTargetArray` is not implemented on `Array1`. The impl should turn the 2D target array (which we know only has one target) into a 1D target array. Something like:
```rust
impl<'a, L: Clone + 'a, S: Data<Elem = L>> FromTargetArray<'a, L> for ArrayBase<S, Ix1> {
    type Owned = Array1<L>;
    type View = ArrayView1<'a, L>;

    fn new_targets(targets: Array2<L>) -> Self::Owned {
        // something like targets.reshape(targets.nrows()).unwrap()
    }

    fn new_targets_view(targets: ArrayView2<'a, L>) -> Self::View {
        // same here
    }
}
```

PABannier:
@YuhanLiin Hey. Could you have a look at what I've done? I'm running into a lot of issues and it feels like it's never ending. I have errors stating `Unsatisfied trait bounds`. I don't have a holistic view of the crate and it's becoming increasingly hard to fix those issues. I'd be glad if you could help in fixing those issues.

src/dataset/mod.rs
```diff
@@ -758,53 +758,6 @@ mod tests {
...
-            .cross_validate_multi(5, &params, |_pred, _truth| {
-                if false {
-                    Ok(array![0f32])
-                } else {
-                    Err(Error::Parameters("eval".to_string()))
-                }
-            })
-            .unwrap_err();
-        assert_eq!(err.to_string(), "invalid parameter eval".to_string());
-    }
```
YuhanLiin, line 787: Is it possible to adapt these tests to use `cross_validate_multi`?

src/dataset/impl_dataset.rs
```diff
@@ -929,7 +849,10 @@ where
         M: for<'c> Fit<ArrayView2<'c, F>, ArrayView2<'c, E>, ER, Object = O>,
         O: for<'d> PredictInplace<ArrayView2<'a, F>, ArrayBase<OwnedRepr<E>, I>>,
         FACC: Float,
-        C: Fn(&ArrayView1<E>, &ArrayView1<E>) -> std::result::Result<FACC, crate::error::Error>,
+        C: Fn(
+            &ArrayView2<E>,
+            &ArrayView2<E>,
+        ) -> std::result::Result<Array1<FACC>, crate::error::Error>,
```
YuhanLiin, line 855: What I meant in my comment was that since `cross_validate` can be expressed in terms of `cross_validate_multi`, we should only have one cross validation API that behaves like `cross_validate_multi`. This means `cross_validate` should behave the same as the old `cross_validate_multi`, and it should return a 2D array instead of 1D.

PABannier: Actually old `cross_validate_multi` returns a 1d array, while the previous implementation of `cross_validate` returns a scalar

YuhanLiin: The signature of `cross_validate_multi` on master is
```
pub fn cross_validate_multi<O, ER, M, FACC, C>(
        &'a mut self,
        k: usize,
        parameters: &[M],
        eval: C,
    ) -> std::result::Result<Array2<FACC>, ER>
```
The return is actually a 2D array

PABannier: Ok my bad, i got confused

YuhanLiin:
After looking at the change in-depth, I'd like to propose a rework to the target traits (sorry that I'm suggesting this at this point in the PR, but I do believe it will make everything significantly easier). The main thorn in our side has been the fact that `as_multi_targets()` yields `ArrayView2` from both 1D and 2D targets, which is bad when we want to turn the view back into the original target type. This was the problem with `FromTargetArray`, `map_targets()`, and a bunch of other functions. I propose changing the target traits as such:
```rust
trait AsTargets {
  type Elem;
  type Ix: Dimensions;
  fn as_targets(&self) -> ArrayView<Self::Elem, Self::Ix>;
  fn ntargets(&self) -> usize { ... }
}

impl AsTargets for ArrayBase<S, Ix1> {
  type Ix = Ix1;
  ...
}
impl AsTargets for ArrayBase<S, Ix2> {
  type Ix = Ix2;
  ...
}

trait AsSingleTarget: AsTargets<Ix = Ix1> {}

// Might not need this
trait AsMultiTargets: AsTargets<Ix = Ix2> {}
```
If we then change most of the trait bounds to use `AsTargets` instead of `AsMultiTargets`, it allows 1D targets to return 1D views in generic contexts, which solves the problem described above. Since this is a pretty big change I'd be willing to make the change.

As for the errors I see on your branch, they come from `tfidf_validation.rs` and `whitening.rs`. For `tfidf` you need `load_set()` to return `Array1` instead of `Array2`, since Gaussian Naive Bayes now expects a 1D array. This can be done by changing the `from_shape_vec()` call in `load_set()`. For `whitening` the issue is due to `map_targets()` and `datasets::winequality()` always returning 2D arrays. The problem with `map_targets()` would be solved with the above rework, which would also allow you to change `winequality()` to return a 1D target.

PABannier:
@YuhanLiin no worries, the chance is possibly very disruptive so if a solution can significantly alleviate the work load, it is more than welcome. If you could give it shot, that'd be great. I'll try to give it a shot myself today, and push if it is any successful.

YuhanLiin:
Just let me know how it goes, so we don't duplicate work.

PABannier:
@YuhanLiin I tried to start from scratch again and implement what you just said. It does seem to be a less breaking change, but I still run into some issues which I don't know how to solve. I let you take the lead on what you proposed to implement.

YuhanLiin:
I reworked the PR in #206. The refactoring of the main `linfa` crate is entirely complete. Can you fix any remaining failures and update the docs in the main `linfa` crate?

YuhanLiin:
Closed by #206

PR #207, opened 2022-03-28
Eliminate BLAS dependency for ALL crates

YuhanLiin:
Integrates [`ndarray-linalg-rs`](https://github.com/rust-ml/ndarray-linalg-rs), a pure-Rust linalg library, into `linfa-clustering`, `linfa-ica`, `linfa-linear`, `linfa-preprocessing`, `linfa-reduction`, `linfa-elasticnet` and `linfa-pls`. By default those crates will use a Rust implementation of linalg operations, and BLAS versions from `ndarray-linalg` can be opted in via the `blas` feature (this is separate from the `linfa` features for selecting the BLAS library).

## Concerns
- ~~The Rust SVD impl returns a "compact" SVD decomposition, while the `ndarray-linalg` impl returns a "full" SVD decomposition. The two SVDs hold the same information but have different dimensions. This hasn't caused any issues in the tests but could be a concern.~~
- ~~Unlike `ndarray-linalg`, the Rust versions of eigendecomposition and SVD do not return sorted decompositions by default. Some algorithms, such as ICA, use unsorted decompositions only when using the Rust impl, because I've found that sorting did not matter in those cases. I'd like someone more familiar with the algorithms to confirm that opinion.~~
- ~~The Rust impl doesn't support truncated eigen or SVD, so the full decomposition is used instead. This makes the Rust versions of affected algorithms, like PCA, significantly slower.~~
- Diffusion maps in `linfa-reduction` was affected by this change but there are not tests to validate its correctness.
- ~~`test_explained_variance_cutoff` in `linfa-reducton/src/pca.rs` still fails. For some reasons the number of embeddings expected in the output is unequal to the number specified in the input, which I do not understand.~~
- ~~`test_cca` in `linfa-pls` does not converge when BLAS is disabled. It might not have anything to do with linalg, but instead be related to the pure-Rust implementation of matrix multiplication, which is used in non-BLAS builds.~~
- ~~For least squares, `ndarray-linalg` uses the SVD-pseudoinverse algorithm while the Rust library uses the QR algorithm. Both should work fine, so this shouldn't make a difference.~~

## Todo
- [x] Release and possibly rename `ndarray-linalg-rs`.
- [x] Update README with new build features
- [x] Update changelog
- [x] Update feature flags in CI to test both linalg impls


quietlychris (commented):
This all looks good! The non-convergence of `test_cca` is a little concerning, but since you mentioned it doesn't look like a BLAS change issue, that shouldn't be a blocker for this. I also think that the re-organization of the `dataset` module makes sense.

YuhanLiin:
Update:
After some linalg hacking, I also integrated the pure-Rust linalg library into `linfa-elasticnet`, `linfa-preprocessing`, and `linfa-linear`. This means that, if things go as planned, **BLAS will no longer be required to use Linfa at all**.

YuhanLiin:
I integrated the new pure-Rust LOBPCG impl Lorenz wrote into `linfa-reduction` things all the PCA tests pass without taking too long, which is great. However, the doctest for diffusion map is failing after I changed the RNGs in `linfa-reduction` to be deterministically seeded. The failure is due to NaNs appearing in the sorted eigenvalue calculations in the `lobpcg` implementation, which leads to an unwrap. @bytesnake, do you know if this is something we can ignore or is it a bug?

YuhanLiin:
OK it looks like the diffusion map doctest now passes after I lower the precision on the LOBPCG call.

algorithms/linfa-ica/src/error.rs
```diff
@@ -18,9 +18,12 @@ pub enum FastIcaError {
...
     #[error("tolerance should be positive but is {0}")]
     InvalidTolerance(f32),
+    #[cfg(feature = "blas")]
+    #[error("Linalg BLAS error: {0}")]
+    LinalgBlasError(#[from] ndarray_linalg::error::LinalgError),
+    #[error("Linalg error: {0}")]
     /// Errors encountered during linear algebra operations
-    #[error("Linalg Error: {0}")]
-    Linalg(#[from] LinalgError),
+    LinalgError(#[from] linfa_linalg::LinalgError),
```
relf, line 26: `#[cfg(not(feature = "blas"))]` missing before this one or it is not really a problem?

YuhanLiin: It's supposed to be absent, to make the `blas` feature strictly additive. If I gated `LinalgError` on the absence of `blas`, then code that depends on the non-`blas` version would break if `blas` was enabled anywhere else in the dependency tree.

Issue #209, opened 2022-03-31
CI testing on Windows missing


relf:
Trying to compile `linfa` master, I stumbled on the following:
```
   > cargo test --all --release
   [...]
   Compiling mnist v0.5.0
   Compiling time v0.2.26
error[E0433]: failed to resolve: could not find `unix` in `os`
  --> D:\rlafage\.cargo\registry\src\github.com-1ecc6299db9ec823\mnist-0.5.0\src\download.rs:88:34
   |
88 |                     use std::os::unix::fs::MetadataExt;
   |                                  ^^^^ could not find `unix` in `os`

error[E0599]: no method named `size` found for struct `Metadata` in the current scope
  --> D:\rlafage\.cargo\registry\src\github.com-1ecc6299db9ec823\mnist-0.5.0\src\download.rs:99:49
   |
99 | ...                   current_size = meta.size() as usize;
   |                                           ^^^^ method not found in `Metadata`

Some errors have detailed explanations: E0433, E0599.
For more information about an error, try `rustc --explain E0433`.
error: could not compile `mnist` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error: build failed
```
Actually, it is a `mnist` 0.5 bug. I will open an issue on `mnist` repo. 
The issue for linfa is the lack of automated tests on Windows to catch this.

YuhanLiin:
I can bump the `mnist` version but I don't have experience with Windows in GitHub Actions

quietlychris:
There's not a new version to bump to yet; I just created a branch that should have the fix, but Remi will be testing it locally before merging it will be merged upstream into `mnist` main. Once that happens, it can get pushed onto `crates.io` (it shouldn't be a breaking change, so will just be a patch version bump). I'm not sure what that timeline will look like; it doesn't look like the owner of the `mnist` crate has been very active on Github recently, so it may be at least a few days before everything resolves itself.

relf:
I understand that it has not a very high priority. Such problem may not arise that often.... It has not so far. At the moment, ML in Rust is not that common and on Windows... Well I feel like a weirdo for sure 🤣 . Anyway I will try to submit a PR to implement CI tests for Windows.

quietlychris:
Okay, a PR with the fix for the `mnist` crate has been submitted upstream, just waiting on a response from the maintainer

relf:
Fixed by #210

PR #210, opened 2022-04-01
Add CI windows testing

relf:
Tentative PR to get windows testing for linfa. 

I've removed the usage of the container `rustmath:mkl-rust` to factorize actions and as I have some doubt [this Docker image](https://hub.docker.com/r/rustmath/mkl/tags) will be maintained in the future. Moreover, it might also prepare the field for no-blas PR landing.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -4,8 +4,8 @@ name: Run Tests
 
 jobs:
   testing:
-    name: testing-${{ matrix.toolchain }}-${{ matrix.platform.os }}
-    runs-on: ${{ matrix.platform.os }}
+    name: testing-${{ matrix.toolchain }}-${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
     if: github.event.pull_request.draft == false
     strategy:
       fail-fast: false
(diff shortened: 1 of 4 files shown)


relf:
The PR fails on windows as expected. Waiting for `mnist 0.5` windows bug fix.

.github/workflows/testing.yml
```diff
@@ -4,19 +4,20 @@ name: Run Tests
...
       matrix:
         toolchain:
           - 1.54.0
           - stable
-
+        platform: [{ os: "ubuntu-18.04", 
+            rust-target: "x86_64-unknown-linux-gnu"},
+          { os: "windows-2019", 
+            rust-target: "x86_64-pc-windows-msvc"},
+        ]
```
YuhanLiin, line 20: Fix up formatting

YuhanLiin:
The container is definitely unmaintained, and the tests actually run faster without it. We should also parametrize the build checking step by the OS.

.github/workflows/testing.yml
relf: Done

relf:
I fiddled quite a bit with the github actions yml. I did not succeed in making `include` work with `os` definition, so in the end I removed it and handled `nightly` like `1.54` and `stable`.

relf:
@YuhanLiin, @quietlychris, I've finally disabled the `mnist` example in `linfa-tsne` (it can be enabled again once `mnist` > 0.5 is released).

YuhanLiin (approved):

YuhanLiin:
We can keep the issue open until the issue is fixed in `mnist`

relf:
Ok, I've removed the text in the PR description which would triggered the closing of the related issue #209. 

Note: once this PR is merged, required status checks have to be adjusted.

PR #212, opened 2022-04-19
Add Follow The Regularized Leader - Proximal model

MilaKyr:
The goal of this PR is to add FTRL model with L1 and L2 regularization. This model is primarily used in CTR prediction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,23 +28,24 @@ Where does `linfa` stand right now? [Are we learning yet?](http://www.arewelearn
 `linfa` currently provides sub-packages with the following algorithms: 
 
 
-| Name | Purpose | Status | Category |  Notes | 
-| :--- | :--- | :---| :--- | :---| 
-| [clustering](algorithms/linfa-clustering/) | Data clustering | Tested / Benchmarked  | Unsupervised learning | Clustering of unlabeled data; contains K-Means, Gaussian-Mixture-Model, DBSCAN and OPTICS | 
-| [kernel](algorithms/linfa-kernel/) | Kernel methods for data transformation  | Tested  | Pre-processing | Maps feature vector into higher-dimensional space| 
-| [linear](algorithms/linfa-linear/) | Linear regression | Tested  | Partial fit | Contains Ordinary Least Squares (OLS), Generalized Linear Models (GLM) | 
-| [elasticnet](algorithms/linfa-elasticnet/) | Elastic Net | Tested | Supervised learning | Linear regression with elastic net constraints |
-| [logistic](algorithms/linfa-logistic/) | Logistic regression | Tested  | Partial fit | Builds two-class logistic regression models
-| [reduction](algorithms/linfa-reduction/) | Dimensionality reduction | Tested | Pre-processing | Diffusion mapping and Principal Component Analysis (PCA) |
-| [trees](algorithms/linfa-trees/) | Decision trees | Experimental  | Supervised learning | Linear decision trees
-| [svm](algorithms/linfa-svm/) | Support Vector Machines | Tested  | Supervised learning | Classification or regression analysis of labeled datasets | 
-| [hierarchical](algorithms/linfa-hierarchical/) | Agglomerative hierarchical clustering | Tested | Unsupervised learning | Cluster and build hierarchy of clusters |
-| [bayes](algorithms/linfa-bayes/) | Naive Bayes | Tested | Supervised learning | Contains Gaussian Naive Bayes |
-| [ica](algorithms/linfa-ica/) | Independent component analysis | Tested | Unsupervised learning | Contains FastICA implementation |
-| [pls](algorithms/linfa-pls/) | Partial Least Squares | Tested | Supervised learning | Contains PLS estimators for dimensionality reduction and regression |
-| [tsne](algorithms/linfa-tsne/) | Dimensionality reduction| Tested | Unsupervised learning | Contains exact solution and Barnes-Hut approximation t-SNE |
-| [preprocessing](algorithms/linfa-preprocessing/) |Normalization & Vectorization| Tested | Pre-processing | Contains data normalization/whitening and count vectorization/tf-idf |
-| [nn](algorithms/linfa-nn/) | Nearest Neighbours & Distances | Tested / Benchmarked | Pre-processing | Spatial index structures and distance functions |
+| Name                                             | Purpose                                  | Status | Category | Notes                                                                                     | 
+|:-------------------------------------------------|:-----------------------------------------| :---| :--- |:------------------------------------------------------------------------------------------| 
+| [clustering](algorithms/linfa-clustering/)       | Data clustering                          | Tested / Benchmarked  | Unsupervised learning | Clustering of unlabeled data; contains K-Means, Gaussian-Mixture-Model, DBSCAN and OPTICS | 
+| [kernel](algorithms/linfa-kernel/)               | Kernel methods for data transformation   | Tested  | Pre-processing | Maps feature vector into higher-dimensional space                                         | 
+| [linear](algorithms/linfa-linear/)               | Linear regression                        | Tested  | Partial fit | Contains Ordinary Least Squares (OLS), Generalized Linear Models (GLM)                    | 
+| [elasticnet](algorithms/linfa-elasticnet/)       | Elastic Net                              | Tested | Supervised learning | Linear regression with elastic net constraints                                            |
+| [logistic](algorithms/linfa-logistic/)           | Logistic regression                      | Tested  | Partial fit | Builds two-class logistic regression models                                               
+| [reduction](algorithms/linfa-reduction/)         | Dimensionality reduction                 | Tested | Pre-processing | Diffusion mapping and Principal Component Analysis (PCA)                                  |
+| [trees](algorithms/linfa-trees/)                 | Decision trees                           | Experimental  | Supervised learning | Linear decision trees                                                                     
+| [svm](algorithms/linfa-svm/)                     | Support Vector Machines                  | Tested  | Supervised learning | Classification or regression analysis of labeled datasets                                 | 
+| [hierarchical](algorithms/linfa-hierarchical/)   | Agglomerative hierarchical clustering    | Tested | Unsupervised learning | Cluster and build hierarchy of clusters                                                   |
+| [bayes](algorithms/linfa-bayes/)                 | Naive Bayes                              | Tested | Supervised learning | Contains Gaussian Naive Bayes                                                             |
+| [ica](algorithms/linfa-ica/)                     | Independent component analysis           | Tested | Unsupervised learning | Contains FastICA implementation                                                           |
+| [pls](algorithms/linfa-pls/)                     | Partial Least Squares                    | Tested | Supervised learning | Contains PLS estimators for dimensionality reduction and regression                       |
+| [tsne](algorithms/linfa-tsne/)                   | Dimensionality reduction                 | Tested | Unsupervised learning | Contains exact solution and Barnes-Hut approximation t-SNE                                |
+| [preprocessing](algorithms/linfa-preprocessing/) | Normalization & Vectorization            | Tested | Pre-processing | Contains data normalization/whitening and count vectorization/tf-idf                      |
+| [nn](algorithms/linfa-nn/)                       | Nearest Neighbours & Distances           | Tested / Benchmarked | Pre-processing | Spatial index structures and distance functions                                           |
+| [ftrl](algorithms/linfa-ftrl/)                   | Follow The Reguralized Leader - proximal | Tested  | Partial fit | Contains L1 and L2 regularization. Possible incremental update                            | 
 
 We believe that only a significant community effort can nurture, build, and sustain a machine learning ecosystem in Rust - there is no other way forward.
 
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -13,6 +13,10 @@ readme = "README.md"
 keywords = ["machine-learning", "linfa", "ai", "ml", "ftrl"]
 categories = ["algorithms", "mathematics", "science"]
 
+[features]
+default = []
+serde = ["serde_crate"]
+
 [dependencies]
 ndarray = { version = "0.15"}
 ndarray-rand = "0.14.0"
@@ -25,3 +29,10 @@ linfa = { version = "0.5.0", path = "../.."}
 [dev-dependencies]
 approx = "0.4"
 linfa-datasets = { version = "0.5.0", path = "../../datasets", features = ["winequality"] }
+
+[dependencies.serde_crate]
+package = "serde"
+optional = true
+version = "1.0"
+default-features = false
+features = ["std", "derive"]
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -25,7 +25,7 @@ serde = {version = "0.9.15", features = ["derive"]}
 linfa = { version = "0.5.0", path = "../.."}
 
 [dev-dependencies]
-iai = "0.1"
+criterion = "0.3"
 approx = "0.4"
 linfa-datasets = { version = "0.5.0", path = "../../datasets", features = ["winequality"] }
 
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -13,16 +13,14 @@ readme = "README.md"
 keywords = ["machine-learning", "linfa", "ai", "ml", "ftrl"]
 categories = ["algorithms", "mathematics", "science"]
 
-[features]
-default = []
-serde = ["serde_crate"]
-
 [dependencies]
 ndarray = { version = "0.15"}
 ndarray-rand = "0.14.0"
 argmin = { version = "0.4.7", features = ["ndarray", "ndarray-rand"]}
 thiserror = "1.0"
 rand = "0.8.5"
+rand_xoshiro = "0.6.0"
+serde = {version = "0.9.15", features = ["derive"]}
 
 linfa = { version = "0.5.0", path = "../.."}
 
@@ -30,13 +28,6 @@ linfa = { version = "0.5.0", path = "../.."}
 iai = "0.1"
 approx = "0.4"
 linfa-datasets = { version = "0.5.0", path = "../../datasets", features = ["winequality"] }
-
-[dependencies.serde_crate]
-package = "serde"
-optional = true
-version = "1.0"
-default-features = false
-features = ["std", "derive"]
 
 [[bench]]
 name = "ftrl"
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -14,13 +14,13 @@ keywords = ["machine-learning", "linfa", "ai", "ml", "ftrl"]
 categories = ["algorithms", "mathematics", "science"]
 
 [dependencies]
-ndarray = { version = "0.15"}
+ndarray = { version = "0.15.4", features = ["serde"]}
 ndarray-rand = "0.14.0"
 argmin = { version = "0.4.7", features = ["ndarray", "ndarray-rand"]}
 thiserror = "1.0"
 rand = "0.8.5"
 rand_xoshiro = "0.6.0"
-serde = {version = "0.9.15", features = ["derive"]}
+serde = {version = "1.0.137", features = ["derive"]}
 
 linfa = { version = "0.5.0", path = "../.."}
 
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -14,20 +14,13 @@ keywords = ["machine-learning", "linfa", "ai", "ml", "ftrl"]
 categories = ["algorithms", "mathematics", "science"]
 
 [dependencies]
-ndarray = { version = "0.15", features = ["blas"] }
+ndarray = { version = "0.15"}
 ndarray-rand = "0.14.0"
-argmin = { version = "0.4.6", features = ["ndarrayl"] }
+argmin = { version = "0.5.1", features = ["ndarray", "ndarray-rand"]}
 thiserror = "1.0"
 rand = "0.8.5"
 
-linfa = { version = "0.5.0", path = "../..", features=["serde"] }
-
-[dependencies.serde_crate]
-package = "serde"
-optional = true
-version = "1.0"
-default-features = false
-features = ["std", "derive"]
+linfa = { version = "0.5.0", path = "../.."}
 
 [dev-dependencies]
 approx = "0.4"
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -27,6 +27,7 @@ rand = "0.8.5"
 linfa = { version = "0.5.0", path = "../.."}
 
 [dev-dependencies]
+iai = "0.1"
 approx = "0.4"
 linfa-datasets = { version = "0.5.0", path = "../../datasets", features = ["winequality"] }
 
@@ -36,3 +37,7 @@ optional = true
 version = "1.0"
 default-features = false
 features = ["std", "derive"]
+
+[[bench]]
+name = "ftrl"
+harness = false
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.5.0"
 authors = ["Liudmyla Kyrashchuk <himila@tutanota.com>"]
 
 description = "A Machine Learning framework for Rust"
-edition = "2021"
+edition = "2018"
 license = "MIT/Apache-2.0"
 
 repository = "https://github.com/rust-ml/linfa"
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -1,0 +1,34 @@
+[package]
+name = "linfa-ftrl"
+version = "0.5.0"
+authors = ["Liudmyla Kyrashchuk <himila@tutanota.com>"]
+
+description = "A Machine Learning framework for Rust"
+edition = "2021"
+license = "MIT/Apache-2.0"
+
+repository = "https://github.com/rust-ml/linfa"
+readme = "README.md"
+
+keywords = ["machine-learning", "linfa", "ai", "ml", "ftrl"]
+categories = ["algorithms", "mathematics", "science"]
+
+[dependencies]
+ndarray = { version = "0.15", features = ["blas"] }
+ndarray-rand = "0.14.0"
+argmin = { version = "0.4.6", features = ["ndarrayl"] }
+thiserror = "1.0"
+rand = "0.8.5"
+
+linfa = { version = "0.5.0", path = "../..", features=["serde"] }
+
+[dependencies.serde_crate]
+package = "serde"
+optional = true
+version = "1.0"
+default-features = false
+features = ["std", "derive"]
+
+[dev-dependencies]
+approx = "0.4"
+linfa-datasets = { version = "0.5.0", path = "../../datasets", features = ["winequality"] }
--- a/algorithms/linfa-ftrl/Cargo.toml
+++ b/algorithms/linfa-ftrl/Cargo.toml
@@ -20,7 +20,7 @@ serde = ["serde_crate"]
 [dependencies]
 ndarray = { version = "0.15"}
 ndarray-rand = "0.14.0"
-argmin = { version = "0.5.1", features = ["ndarray", "ndarray-rand"]}
+argmin = { version = "0.4.7", features = ["ndarray", "ndarray-rand"]}
 thiserror = "1.0"
 rand = "0.8.5"
 
--- a/algorithms/linfa-ftrl/README.md
+++ b/algorithms/linfa-ftrl/README.md
@@ -31,7 +31,7 @@ Show source code
 ```rust
 use linfa::prelude::*;
 use linfa::dataset::{AsSingleTargets, Records};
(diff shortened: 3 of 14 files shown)


algorithms/linfa-ftrl/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[cfg(feature = "serde")]
+use serde_crate::{Deserialize, Serialize};
+
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(crate = "serde_crate")
+)]
+pub struct FollowTheRegularizedLeader<F: Float> {
```
YuhanLiin, line 38: Rename this to `FTRL`, since the full name is a bit long. Also add docs and an example to this type.

algorithms/linfa-ftrl/src/lib.rs
```diff
@@ -0,0 +1,63 @@
...
+//! `linfa-ftrl` provides a pure Rust implementation of an [algorithm](struct.FollowTheRegularizedLeader.html).
+//!
+//! ## Examples
+//!
+//! There is an usage example in the `examples/` directory. To run, use:
+//!
+//! ```bash
+//! $ cargo run --example winequality
+//! ```
+//!
```
YuhanLiin, line 17: Try replacing with `#![doc = include_str!("../README.md")]`. This syncs the README docs with the crate docs and runs the README example as a doctest.

YuhanLiin:
@bytesnake @quietlychris Do you think this should be moved into an existing crate (I'm thinking `linfa-logistic`) or should it get its own crate?

algorithms/linfa-ftrl/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+license = "MIT/Apache-2.0"
+
+repository = "https://github.com/rust-ml/linfa"
+readme = "README.md"
+
+keywords = ["machine-learning", "linfa", "ai", "ml", "ftrl"]
+categories = ["algorithms", "mathematics", "science"]
+
+[dependencies]
+ndarray = { version = "0.15", features = ["blas"] }
```
YuhanLiin, line 17: You don't need the `blas` feature

algorithms/linfa-ftrl/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+let mut model = FTRL::new(&params, train.nfeatures());
+
+// Bootstrap each row from the train dataset to imitate online nature of the data flow
+let mut rng = SmallRng::seed_from_u64(42);
+let mut row_iter = train.bootstrap_samples(1, &mut rng);
+for _ in 0..train.nsamples() {
+let b_dataset = row_iter.next().unwrap();
+model = params.fit_with(Some(model), &b_dataset)?;
+}
```
YuhanLiin, line 56: Fix indentation

src/metrics_classification.rs
```diff
@@ -475,12 +476,30 @@ impl BinaryClassification<&[bool]> for &[Pr] {
...
+            .iter()
+            .map(|v| v.clamp(f32::EPSILON, 1. - f32::EPSILON))
+            .collect();
+        clipped_probs
+            .iter()
+            .zip(y.iter())
+            .map(|(a, b)| if b == &true { -a.ln() } else { -(1. - a).ln() })
+            .collect::<Array1<f32>>()
+            .mean()
+            .ok_or(Error::NotEnoughSamples)
```
YuhanLiin, line 491: Move this code into the impl for `AsSingleTargets`. In this impl call `aview1` on the slices and invoke the `log_loss` impl for `AsSingleTargets`. Doing this allows `log_loss` to work for non-standard array layouts.

src/dataset/mod.rs
```diff
@@ -115,6 +115,14 @@ impl PartialOrd for Pr {
     }
 }
 
+impl Sub for Pr {
+    type Output = Pr;
+
+    fn sub(self, rhs: Self) -> Self::Output {
+        Pr(self.0 - rhs.0)
```
YuhanLiin, line 122: Should we check for negatives here?

algorithms/linfa-ftrl/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use crate::hyperparams::FtrlParams;
+pub use algorithm::Result;
+pub use error::FtrlError;
+use linfa::Float;
+use ndarray::Array1;
+use ndarray_rand::RandomExt;
+use rand::distributions::Uniform;
+
+#[cfg(feature = "serde")]
```
YuhanLiin, line 15: Feature-gating by `serde` does nothing here, because you don't have a `serde` feature in this crate. And there's no point in having a `serde` feature because `argmin` doesn't feature-gate its dependency on `serde`.

algorithms/linfa-ftrl/src/hyperparams.rs
```diff
@@ -0,0 +1,123 @@
+use crate::error::FtrlError;
+use linfa::{Float, ParamGuard};
+
+#[derive(Clone)]
+pub struct FtrlParams<F: Float>(pub(crate) FtrlValidParams<F>);
+
+/// A verified hyper-parameter set ready for the estimation of a Follow the regularized leader - proximal model
+///
+/// See [`FtrlParams`](crate::FtrlParams) for more information.
+#[derive(Clone)]
```
YuhanLiin, line 10: Implement `Debug`

algorithms/linfa-ftrl/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+use serde_crate::{Deserialize, Serialize};
+
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(crate = "serde_crate")
+)]
+pub struct FTRL<F: Float> {
```
YuhanLiin, line 23: Derive `Debug` and `Clone`

algorithms/linfa-ftrl/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    ///
+    /// It requires data preprocessing done in the separate step.
+
+    /// Create default hyperparameters
+    pub fn params() -> FtrlParams<F> {
+        FtrlParams::default()
+    }
+
+    /// Create a new model with given parameters and number of features
+    pub fn new(params: &FtrlParams<F>, nfeatures: usize) -> FTRL<F> {
```
YuhanLiin, line 54: Pass in `params` by value so you don't clone inside the method

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// Get the hyperparameters
+    pub fn get_params(&self) -> &FtrlParams<F> {
+        &self.params
+    }
+
+    /// Calculate weights for model prediction
+    pub fn get_weights(&self) -> Array1<F> {
+        Zip::from(self.z.view())
+            .and(self.n.view())
+            .map_collect(|z_, n_| {
```
YuhanLiin, line 96: Rename those variables to z and n

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+        let weights = self.get_weights();
+        let mut probabilities = x.dot(&weights);
+        probabilities.mapv_inplace(sigmoid);
+        probabilities
+    }
+
+    fn calculate_sigma(&self, gradients: ArrayView1<F>) -> Array1<F> {
+        Zip::from(&self.n)
+            .and(gradients)
+            .map_collect(|n_, grad| calculate_sigma(*n_, *grad, self.params.0.alpha))
```
YuhanLiin, line 131: Rename `n_` to `n`

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    fn update_params(&mut self, gradient: Array1<F>, sigma: Array1<F>) {
+        self.z = &self.z + &gradient - sigma * self.get_weights();
```
YuhanLiin, line 135: Do:
```rust
self.z += gradient;
self.z -= sigma * self.get_weights();
```

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+        self.n = &self.n + gradient.mapv(|grad| grad.powf(F::cast(2.)));
```
YuhanLiin, line 136: `self.n += gradient.mapv(...)`

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+        dataset: &DatasetBase<ArrayBase<D, Ix2>, T>,
+        probabilities: ArrayView1<Pr>,
+    ) {
+        let probabilities = probabilities.mapv(|prob| F::cast(prob.0));
+        let gradient = calculate_gradient(probabilities.view(), dataset);
+        let sigma = self.calculate_sigma(gradient.view());
+        self.update_params(gradient, sigma);
+    }
+
+    fn predict_probabilities<D: Data<Elem = F>>(&self, x: &ArrayBase<D, Ix2>) -> Array1<F> {
```
YuhanLiin, line 121: Makes more sense for this to return `Array1<Pr>`

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+        });
+    diff.dot(x)
+}
+
+fn calculate_sigma<F: Float>(n: F, gradient: F, alpha: F) -> F {
+    (F::sqrt(n + gradient.powf(F::cast(2.))) - F::sqrt(n)) / alpha
+}
+
+fn sigmoid<F: Float>(prediction: F) -> F {
+    F::one() / (F::one() + F::exp(-F::max(F::min(prediction, F::cast(35.)), F::cast(-35.))))
```
YuhanLiin, line 160: Write this as `F::one() / (F::one() + (-prediction.min(F::cast(35.)).max(F::cast(-35.))).exp())`

src/metrics_classification.rs
MilaKyr: I'm not sure I understand. 
As I understand it, calculate `log_loss` would only make sense when there is binary classification, so moving it to the impl for `AsSingleTargets` would cause to calculate it for different types `<T>`.
Can you help me understand it?

src/dataset/mod.rs
MilaKyr: I agree but I'm not sure if it should be here. We can hide `f32` within `Pr` and instead use some method, e.g. `try_from` to create `Pr` while checking for negative. Something like that:
```rust
pub struct Pr(f32)

// and later initialize with
Pr::try_from(0.7)?
```

What do you think?

src/metrics_classification.rs
YuhanLiin: I'm basically saying to move the content of this impl into the impl at line 519. That impl is defined for `AsSingleTargets<Elem = Pr>`, so it's not going to be generic across `T`.

src/dataset/mod.rs
YuhanLiin: The main drawback of doing this is that it breaks existing code that constructs `Pr`. While there's not much code doing that in `linfa`, there might be external users who'll be broken. However, `Pr` is basically useless if we don't restrain its value, and we already have breaking changes in this release. I'd say add a `try_from` impl that checks `0 <= prob <= 1.0` and returns `Result<Pr, f32>`. Then add `new`, which constructs `Pr` infallibly and panics on bad input (this is how `Pr` will be constructed most of the time). Finally add `new_unchecked` which constructs `Pr` without any checking, for converting large batches. You'll need to add the invariant of `0 <= prob <= 1.0` to `Pr`'s docs, and get rid of the `Sub` impl.

If you really want to keep the `Sub` impl, then just panic in the `Sub` impl if the result is under 0. Regardless of how you approach this, I'd like you to add `#[repr(transparent)]` to `Pr`.

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    fn update_params(&mut self, gradient: Array1<F>, sigma: Array1<F>) {
+        self.z = &self.z + &gradient - sigma * self.get_weights();
+        self.n = &self.n + gradient.mapv(|grad| grad.powf(F::cast(2.)));
+    }
+}
+
+fn calculate_gradient<F: Float, D: Data<Elem = F>, T: AsSingleTargets<Elem = bool>>(
+    probabilities: ArrayView1<F>,
```
YuhanLiin, line 141: Makes more sense for this to be `ArrayView1<Pr>`

src/metrics_classification.rs
MilaKyr: Thank you. Using it in line 519 caused the error `stack overflow`. I moved it to the implementation of `ArrayBase`.

algorithms/linfa-ftrl/src/lib.rs
```diff
@@ -4,22 +4,15 @@ mod algorithm;
...
 
-#[cfg(feature = "serde")]
-use serde_crate::{Deserialize, Serialize};
-
-#[cfg_attr(
-    feature = "serde",
-    derive(Serialize, Deserialize),
-    serde(crate = "serde_crate")
-)]
+#[derive(Clone, Debug)]
```
YuhanLiin, line 15: I'd still derive `serde` on these types, since we'te including `Serde` unconditionally

src/dataset/mod.rs
```diff
@@ -94,10 +95,36 @@ impl<L: Label> Label for Option<L> {}
...
+        }
+    }
+}
 
 impl Pr {
+    pub fn new(prob: f32) -> Self {
+        match Pr::try_from(prob) {
+            Ok(pr) => pr,
+            Err(_) => panic!(),
+        }
```
YuhanLiin, line 122: Just do `prob.try_into().unwrap()`

YuhanLiin:
Do you mind adding a benchmark for FTRL. I'd prefer one written in `iai` but `Criterion` works too. Also update the changelog and the main `linfa` README with the new crate.

src/metrics_classification.rs
YuhanLiin: That doesn;t make sense, since you don't have any recursion. Can you show me the error?

MilaKyr: > thread 'metrics_classification::tests::log_loss' has overflowed its stack
fatal runtime error: stack overflow

I commited this change in `1e47aa0`, if you run `cargo test` it will fail for the `log_loss` with this error

MilaKyr:
> Do you mind adding a benchmark for FTRL. I'd prefer one written in `iai` but `Criterion` works too. Also update the changelog and the main `linfa` README with the new crate.

Sure!

I see that some checks failed due to the error `feature edition2021 is required`, caused by `argmin=0.5.1`. Should I downgrade it?

quietlychris:
Sorry for the delay, just saw this. I replied in the Zulip chat as well, but figured I'd leave a comment here as well. For now, I'd suggest integrating this as a new crate by itself. 

I think we're starting to reach a point on the larger project level that a general consolidation of sub-crates may begin to make sense, but since there are a couple other large-ish changes that are pending for integration into the master branch, I think it makes sense to hold off on new organizational approaches until things have settled a little bit. Since `linfa-logistic` is not currently a composite sub-crate like `linfa-clustering` is, I think a new crate makes the most sense. As always, if someone feels differently, happy to discuss further

src/metrics_classification.rs
```diff
@@ -478,17 +478,8 @@ impl BinaryClassification<&[bool]> for &[Pr] {
...
-            .map(|v| v.clamp(f32::EPSILON, 1. - f32::EPSILON))
-            .collect();
-        clipped_probs
-            .iter()
-            .zip(y.iter())
-            .map(|(a, b)| if b == &true { -a.ln() } else { -(1. - a).ln() })
-            .collect::<Array1<f32>>()
-            .mean()
-            .ok_or(Error::NotEnoughSamples)
+        let probabilities = Array1::from_vec(self.to_vec());
```
YuhanLiin, line 481: Instead do `let probabilities = aview1(self)`

src/metrics_classification.rs
```diff
@@ -498,7 +489,17 @@ impl<D: Data<Elem = Pr>> BinaryClassification<&[bool]> for ArrayBase<D, Ix1> {
...
+            .iter()
+            .map(|v| (*v).clamp(f32::EPSILON, 1. - f32::EPSILON))
+            .collect();
+        clipped_probs
+            .iter()
+            .zip(y.iter())
+            .map(|(a, b)| if *b { -a.ln() } else { -(1. - a).ln() })
+            .collect::<Array1<f32>>()
+            .mean()
+            .ok_or(Error::NotEnoughSamples)
```
YuhanLiin: To avoid creating arrays. do:
```rust
let sum = self
  .iter()
  .map(|v| ...)
  .zip(y.iter())
  .map(|(a, b)| ...)
  .sum();
let len = self.len();
if len == 0 {
  Err(Error::NotEnoughSamples)
} else {
  Ok(sum / len)
}
```

YuhanLiin:
> Sure!
> 
> I see that some checks failed due to the error `feature edition2021 is required`, caused by `argmin=0.5.1`. Should I downgrade it?

Set argmin version to `0.4.6` or `0.4.7`, since that's what other `linfa` crates use. We can't upgrade it to `0.5` until we bump MSRV to 2021 edition.

src/metrics_classification.rs
YuhanLiin: The issue was because the first two `log_loss` impls were calling each other, instead of the 3rd impl. I don't think you can define the first two impls in terms of the 3rd, so the way you have it now is fine.

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,346 @@
...
+        .and(targets)
+        .map_collect(|prob, y| {
+            let truth = if *y { F::one() } else { F::zero() };
+            F::cast(**prob) - truth
+        });
+    diff.dot(x)
+}
+
+fn calculate_sigma<F: Float>(n: F, gradient: F, alpha: F) -> F {
+    (F::sqrt(n + gradient.powf(F::cast(2.))) - F::sqrt(n)) / alpha
```
bytesnake, line 157: why not just `gradient * gradient`?

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,346 @@
...
+    let diff = Zip::from(&probabilities)
+        .and(targets)
+        .map_collect(|prob, y| {
+            let truth = if *y { F::one() } else { F::zero() };
+            F::cast(**prob) - truth
+        });
+    diff.dot(x)
+}
+
+fn calculate_sigma<F: Float>(n: F, gradient: F, alpha: F) -> F {
```
bytesnake, line 156: this weights the current solution to the dual sum, so why not call it `calculate_weight_in_average` so that people reading the `update` function have more pointers how it relates to RDA (than just comparing 1-to-1 to the paper with sigma)

bytesnake (commented):
Thank you!

algorithms/linfa-ftrl/src/algorithm.rs
```diff
@@ -0,0 +1,346 @@
...
+    (F::sqrt(n + gradient.powf(F::cast(2.))) - F::sqrt(n)) / alpha
+}
+
+fn sigmoid<F: Float>(prediction: F) -> F {
+    F::one() / (F::one() + (-prediction.min(F::cast(35.)).max(F::cast(-35.))).exp())
```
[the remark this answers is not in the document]
MilaKyr: Cool! Thanks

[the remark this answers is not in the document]
YuhanLiin: Do we have sigmoid defined in `linfa` somewhere? Maybe we should have common definitions for common numerical operations.

README.md
```diff
@@ -28,23 +28,24 @@ Where does `linfa` stand right now? [Are we learning yet?](http://www.arewelearn
...
+| [pls](algorithms/linfa-pls/)                     | Partial Least Squares                    | Tested | Supervised learning | Contains PLS estimators for dimensionality reduction and regression                       |
+| [tsne](algorithms/linfa-tsne/)                   | Dimensionality reduction                 | Tested | Unsupervised learning | Contains exact solution and Barnes-Hut approximation t-SNE                                |
+| [preprocessing](algorithms/linfa-preprocessing/) | Normalization & Vectorization            | Tested | Pre-processing | Contains data normalization/whitening and count vectorization/tf-idf                      |
+| [nn](algorithms/linfa-nn/)                       | Nearest Neighbours & Distances           | Tested / Benchmarked | Pre-processing | Spatial index structures and distance functions                                           |
+| [ftrl](algorithms/linfa-ftrl/)                   | Follow The Reguralized Leader - proximal | Tested  | Partial fit | Contains L1 and L2 regularization. Possible incremental update                            |
```
YuhanLiin: Status should be "Tested / Benchmarked"

algorithms/linfa-ftrl/Cargo.toml
```diff
@@ -35,4 +36,8 @@ package = "serde"
 optional = true
 version = "1.0"
 default-features = false
-features = ["std", "derive"]
\ No newline at end of file
+features = ["std", "derive"]
+
+[[bench]]
+name = "ftrl"
+harness = false
```
YuhanLiin, line 43: add newline

algorithms/linfa-ftrl/benches/ftrl.rs
```diff
@@ -0,0 +1,55 @@
...
+    let mut rng = SmallRng::seed_from_u64(42);
+    let params = FTRL::params();
+    for nfeatures in (10..100).step_by(10) {
+        for size in (1_000..10_000).step_by(2_000) {
+            let dataset = get_dataset(&mut rng, size, nfeatures);
+            let model = params.fit_with(None, iai::black_box(&dataset)).unwrap();
+            model.predict(iai::black_box(dataset));
+        }
+    }
+}
```
YuhanLiin, line 20: I'm not too familiar with `iai`, but this benchmark only counts the `fit_with` and `predict` calls right? It shouldn't be including the `get_dataset` call in the measurement.

algorithms/linfa-ftrl/src/lib.rs
YuhanLiin: What I meant in this comment is that this you should add `Serialize, Deserialize` to this derive, and delete the `cfg_attr` above, since it depends on an non-existent `serde` feature. Do this for all types with `serde` annotations.

algorithms/linfa-ftrl/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub fn params() -> FtrlParams<F> {
+        FtrlParams::default()
+    }
+
+    /// Create a new model with given parameters and number of features
+    pub fn new(params: FtrlValidParams<F>, nfeatures: usize) -> FTRL<F> {
+        Self {
+            params,
+            n: Array1::zeros(nfeatures),
+            z: Array1::random(nfeatures, Uniform::new(F::zero(), F::one())),
```
YuhanLiin, line 57: `Array1::random` is problematic because it always uses `SmallRng` and is seeded by `ThreadRng`. For testing we want an RNG we seed ourselves for reproduceability and in the real world we want a higher-quality RNG like `Xoshiro` (assuming that the initial values of `z` actually matter for performance, if it doesn't then we might not even need to use an RNG here). Therefore I'd like this to call `Array1::random_using` instead with a custom RNG provided by `FtrlParams`. This means `FtrlParams` will need a type parameter of `R: Rng` and have a field for the RNG. This would also mean you can no longer clone the params directly into `FTRL`. In that case I suggest replicating the current fields of `FtrlParams` in `FTRL` and copying those fields instead.

README.md
MilaKyr: Yes, fixed it during merging master into ftrl

algorithms/linfa-ftrl/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+use ndarray_rand::RandomExt;
+use rand::distributions::Uniform;
+
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(crate = "serde_crate")
+)]
+#[derive(Clone, Debug)]
+pub struct FTRL<F: Float> {
```
relf, line 22: Hi, thanks for your contribution! @YuhanLiin nitpicking: should not it be `Ftrl` instead of `FTRL` as we already have `Pls`, `Pca` or `Svm` in `linfa`?

algorithms/linfa-ftrl/src/lib.rs
MilaKyr: Oh, I got it wrong, apologies!

algorithms/linfa-ftrl/benches/ftrl.rs
MilaKyr: I'm only discovering this crate, but from what I can see in their documentation, specifically `Con: Iai cannot exclude setup code from the measurements, where Criterion-rs can.`, it does include everything. On the side note, I can rewrite benchmark with Criterion.

algorithms/linfa-ftrl/src/lib.rs
MilaKyr: I tried to do this, but run into some problems. Maybe you could help me with them. Adding `R: Rng` to the `FtrlParams` by cascade causes in `fit_with` refer to `&mut self` instead of `&self`, because if we pass to `fit_with` `None` as model parameter, we must create default model and to do this we must use &mut params (and consequently &mut rng) meaning &mut self in `fit_with`. Maybe there is simple solution, it would be great if you can point me towards it.
On the other hand, `fit_with` IMHO is very important in online settings.

I would propose to use another approach. I would suggest to add to `FTRL::new(..)` parameter `rng: &mut dyn Rng`  to initialize `z` values. Additionally, add `get_rng()` to `FtrlParams` which returns `rand_xoshiro::Xoshiro256Plus`. Let me know what you think

algorithms/linfa-ftrl/src/algorithm.rs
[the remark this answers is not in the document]
YuhanLiin: We should figure out all the common math functions we want to refactor out before making any changes. Just keep it as is for this PR and open an issue afterwards.

algorithms/linfa-ftrl/src/lib.rs
YuhanLiin: The way it's done in other parts of `linfa` is to clone the RNG when calling `fit_with` for the first time. This isn't as bad as it sounds because the clone is only done at the start. With your approach, I'm not sure how `fit_with(None)` would work, since we'd need to create a new RNG out of nowhere. `get_rng()` isn't good because `Xoshiro` doesn't necessarily fit every usecase, since sometimes a non-seeded RNG like `ThreadRng` is required (I misspoke in my comment about real world RNGs earlier). This is why we prefer having users provide their own RNGs.

algorithms/linfa-ftrl/benches/ftrl.rs
YuhanLiin: Better to rewrite in Criterion then. Iai seems to lack a lot of features and isn't being actively developed.

algorithms/linfa-ftrl/src/lib.rs
YuhanLiin: I can't really find any counterexamples so yeah, it makes sense to name this `Ftrl`.

algorithms/linfa-ftrl/src/hyperparams.rs
```diff
@@ -56,27 +59,32 @@ impl<F: Float> FtrlValidParams<F> {
...
             alpha,
             beta,
             l1_ratio,
             l2_ratio,
+            rng,
+        })
+    }
+
+    /// Create new hyperparameters with pre-defined random number generator
+    pub fn new_with_rng(rng: R) -> Self {
```
YuhanLiin, line 81: Rename to `default_with_rng`

YuhanLiin (approved):

YuhanLiin:
Final suggestion: Update the changelog's breaking changes with your change to `Pr`.

MilaKyr:
> Final suggestion: Update the changelog's breaking changes with your change to `Pr`.

Did it with the line `Pr has been constrained to 0. <= prob <= 1.. Creating probability value now possible with Pr::new(prob) or Pr::try_from(prob)?`. Should I add something to this?

src/dataset/mod.rs
```diff
@@ -94,10 +95,33 @@ impl<L: Label> Label for Option<L> {}
 /// This helper struct exists to distinguish probabilities from floating points. For example SVM
 /// selects regression or classification training, based on the target type, and could not
 /// distinguish them without a new-type definition.
+///
+/// # Panics
+/// Panics if probability is negative or bigger than one.
```
YuhanLiin, line 100: Move this to the docs of `Pr::new`, and add similar docs to `Pr::try_from` and `Pr::new_unchecked`

YuhanLiin:
> > Final suggestion: Update the changelog's breaking changes with your change to `Pr`.
> 
> Did it with the line `Pr has been constrained to 0. <= prob <= 1.. Creating probability value now possible with Pr::new(prob) or Pr::try_from(prob)?`. Should I add something to this?

Do something like `Constructor Pr(prob) replaced with Pr::new(prob), Pr::try_from(prob), and Pr::new_unchecked(prob)`

src/dataset/mod.rs
```diff
@@ -95,13 +95,17 @@ impl<L: Label> Label for Option<L> {}
...
 #[derive(Debug, Copy, Clone, Default)]
 pub struct Pr(f32);
 
+/// Tries to convert float to probability type.
+///
+/// # Returns
+/// Either probability type Pr(f32) or error as Err(f32)
+///
+/// # Panics
+/// Panics if probability is negative or bigger than one.
```
YuhanLiin, line 108: Delete this part (this method doesn't panic)

YuhanLiin:
Thanks for your contribution!

PR #218, opened 2022-05-03
Fix t-SNE example gnuplot path

Wint3rmute:
# Expected behaviour

When running `cargo run --example tsne  --features linfa/intel-mkl-system` from `linfa/algorithms/linfa-tsne` directory, `gnuplot` should show a plot with t-SNE's result.

# Actual behaviour

```
"examples/iris_plot.plt" line 7: warning: Cannot find or open file "iris.dat"
"examples/iris_plot.plt" line 7: No data in plot
```

It seems like the example code `linfa/algorithms/linfa-tsne/examples/tsne.rs` expects to be run from the `linfa/algorithms/linfa-tsne/` directory (as it writes its result to `examples/iris.dat`), but at the end it calls `gnuplot` using just `iris.dat` path.


YuhanLiin:
Should the same fix be applied to the other tsne example?

Wint3rmute:
> Should the same fix be applied to the other tsne example?

Yes, @YuhanLiin, I've applied it

YuhanLiin (approved):

PR #223, opened 2022-05-27
Isotonic Regression

wildart:
An isotonic regression  based the pool adjacent violators algorithm from *Best, M.J., Chakravarti, N. Active set algorithms for isotonic regression; A unifying framework. Mathematical Programming 47, 425–439 (1990)*.

### DIFF
--- a/algorithms/linfa-linear/src/isotonic.rs
+++ b/algorithms/linfa-linear/src/isotonic.rs
@@ -1,0 +1,337 @@
+//! Isotonic
+#![allow(non_snake_case)]
+use crate::error::{LinearError, Result};
+use ndarray::{s, stack, Array1, ArrayBase, Axis, Data, Ix1, Ix2};
+use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
+
+use linfa::dataset::{AsSingleTargets, DatasetBase};
+use linfa::traits::{Fit, PredictInplace};
+
+pub trait Float: linfa::Float {}
+impl Float for f32 {}
+impl Float for f64 {}
+
+#[derive(Serialize, Deserialize)]
+/// An isotonic regression model.
+///
+/// IsotonicRegression solves an isotonic regression problem using the pool
+/// adjacent violators algorithm.
+///
+/// /// ## Examples
+///
+/// Here's an example on how to train an isotonic regression model on one feature
+/// from the `diabetes` dataset.
+/// ```rust
+/// use linfa::{traits::Fit, traits::Predict, Dataset};
+/// use linfa_linear::IsotonicRegression;
+/// use linfa::dataset::Records;
+/// use linfa::prelude::SingleTargetRegression;
+///
+/// let diabetes = linfa_datasets::diabetes();
+/// let feature1d = diabetes.records().column(0) // first feature of the datasets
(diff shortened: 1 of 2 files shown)


algorithms/linfa-linear/src/isotonic.rs
```diff
@@ -0,0 +1,304 @@
...
+pub struct FittedIsotonicRegression<F> {
+    regressor: Array1<F>,
+    response: Array1<F>,
+}
+
+impl Default for IsotonicRegression {
+    fn default() -> Self {
+        IsotonicRegression::new()
+    }
+}
```
YuhanLiin, line 50: This can just be derived

algorithms/linfa-linear/src/isotonic.rs
```diff
@@ -0,0 +1,304 @@
...
+        let y = dataset.as_single_targets();
+
+        // Check the input dimension
+        assert_eq!(dim, 1, "The input dimension must be 1.");
+
+        // Check that our inputs have compatible shapes
+        assert_eq!(y.dim(), n_samples);
+
+        // use correlation for determining relationship between x & y
+        let x = X.slice(s![.., 0]);
```
YuhanLiin, line 87: Do `X.col(0)`

algorithms/linfa-linear/src/isotonic.rs
```diff
@@ -0,0 +1,304 @@
...
+        let rho = DatasetBase::from(stack![Axis(1), x, y]).pearson_correlation();
+        let incresing = rho.get_coeffs()[0] >= F::zero();
```
YuhanLiin, line 89: `increasing`

algorithms/linfa-linear/src/isotonic.rs
```diff
@@ -0,0 +1,304 @@
...
+            let B_plus = &J[i + 1];
+            if B_zero.0 <= B_plus.0 {
+                i += 1;
+            } else {
+                let w0 = B_zero.1 + B_plus.1;
+                let v0 = (B_zero.0 * B_zero.1 + B_plus.0 * B_plus.1) / w0;
+                let mut i0 = B_zero.2.clone();
+                i0.extend(&(B_plus.2));
+                J[i] = (v0, w0, i0);
+                J.remove(i + 1);
```
YuhanLiin, line 115: Vector removal is O(n), which is inefficient inside loops. Can you change `J` to a different data structure so removals are less expensive? `Hashmap` or `Vec<Option<...>>` are good options, but you'll need to change the way you index into `J`.

wildart:
I rewrote algorithm, and added an another implementation.

YuhanLiin:
What's the difference and tradeoffs between PVA and AlgorithmA? Is there a reason why we want to include both instead of just including one?

wildart:
I removed AlgorithmA code, and left PVA code only. Both algorithms are of *O(n)* complexity, it's just AlgorithmA is dual to PVA, so the solutions they provide are almost similar.

After all, I do not think we need two algorithms implemented.

YuhanLiin:
Mention the paper/textbook the algorithm is from in the public docs and you're good to go

YuhanLiin (approved):

PR #226, opened 2022-06-09
Bernoulli Naive Bayes

wildart:
Following #183 tips. This is an implementation of Bernoulli Naive Bayes classifier derived from Multinomial NB.

### DIFF
--- a/algorithms/linfa-bayes/src/bernoulli_nb.rs
+++ b/algorithms/linfa-bayes/src/bernoulli_nb.rs
@@ -1,0 +1,345 @@
+use linfa::dataset::{AsSingleTargets, DatasetBase, Labels};
+use linfa::traits::{Fit, FitWith, PredictInplace};
+use linfa::{Float, Label};
+use ndarray::{Array1, ArrayBase, ArrayView2, Axis, Data, Ix2, OwnedRepr};
+use std::collections::HashMap;
+use std::hash::Hash;
+
+use crate::base_nb::{filter, NaiveBayes, NaiveBayesValidParams};
+use crate::error::{NaiveBayesError, Result};
+use crate::hyperparams::{BernoulliNbParams, BernoulliNbValidParams};
+
+impl<'a, F, L, D, T> NaiveBayesValidParams<'a, F, L, D, T> for BernoulliNbValidParams<F, L>
+where
+    F: Float,
+    L: Label + 'a,
+    D: Data<Elem = F>,
+    T: AsSingleTargets<Elem = L> + Labels<Elem = L>,
+{
+}
+
+impl<F, L, D, T> Fit<ArrayBase<D, Ix2>, T, NaiveBayesError> for BernoulliNbValidParams<F, L>
+where
+    F: Float,
+    L: Label + Ord,
+    D: Data<Elem = F>,
+    T: AsSingleTargets<Elem = L> + Labels<Elem = L>,
+{
+    type Object = BernoulliNb<F, L>;
+    // Thin wrapper around the corresponding method of NaiveBayesValidParams
+    fn fit(&self, dataset: &DatasetBase<ArrayBase<D, Ix2>, T>) -> Result<Self::Object> {
+        let model = NaiveBayesValidParams::fit(self, dataset, None)?;
+        Ok(model.unwrap())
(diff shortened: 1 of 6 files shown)


algorithms/linfa-bayes/src/bernoulli_nb.rs
```diff
@@ -0,0 +1,336 @@
...
+impl<'a, F, L, D, T> FitWith<'a, ArrayBase<D, Ix2>, T, NaiveBayesError>
+    for BernoulliNbValidParams<F, L>
+where
+    F: Float,
+    L: Label + 'a,
+    D: Data<Elem = F>,
+    T: AsSingleTargets<Elem = L> + Labels<Elem = L>,
+{
+    type ObjectIn = Option<BernoulliNb<F, L>>;
+    type ObjectOut = Option<BernoulliNb<F, L>>;
```
YuhanLiin, line 46: This doesn't need to be an `Option`

algorithms/linfa-bayes/src/bernoulli_nb.rs
```diff
@@ -0,0 +1,336 @@
...
+impl<'a, F, L> BernoulliNbValidParams<F, L>
+where
+    F: Float,
+{
+    // Update log probabilities of features given class
+    fn update_feature_log_prob(
+        &self,
+        info_old: &MultinomialClassInfo<F>,
+        x_new: ArrayView2<F>,
+    ) -> (Array1<F>, Array1<F>) {
```
YuhanLiin, line 129: This method looks like an exact copy of the same method in `multinomial_nb.rs`

algorithms/linfa-bayes/src/bernoulli_nb.rs
YuhanLiin: NVM I see that the rest of the `linfa-bayes` types also return `Option`s. We could fix this later but I'm not sure, so just leave this as is.

algorithms/linfa-bayes/src/bernoulli_nb.rs
```diff
@@ -0,0 +1,336 @@
...
+    where
+        D: Data<Elem = F>,
+    {
+        let thr = self.binarize.unwrap_or(F::zero());
+        let xbin = x.map(|v| if v > &thr { F::one() } else { F::zero() });
+        return if self.binarize.is_some() {
+            xbin.to_owned()
+        } else {
+            x.to_owned()
+        };
```
YuhanLiin, line 232: The signature should return `CowArray2<F>` so that you don't create a new array when there's no threshold.

The body is better written as:
```rust
if let Some(thr) = self.binarize {
  x.map(|v| if v > &thr { F::one() } else { F::zero() }).into()
} else {
  x.into()
}
```

algorithms/linfa-bayes/src/bernoulli_nb.rs
```diff
@@ -0,0 +1,336 @@
...
+        dataset: &DatasetBase<ArrayBase<D, Ix2>, T>,
+    ) -> Result<Self::ObjectOut> {
+        let x = dataset.records();
+        let y = dataset.as_single_targets();
+
+        let mut model = match model_in {
+            Some(temp) => temp,
+            None => BernoulliNb {
+                class_info: HashMap::new(),
+                binarize: self.binarize().to_owned(),
```
YuhanLiin, line 60: Just do `self.binarize`

algorithms/linfa-bayes/src/multinomial_nb.rs
```diff
@@ -202,11 +202,11 @@ pub struct MultinomialNb<F: PartialEq, L: Eq + Hash> {
...
-struct MultinomialClassInfo<F> {
-    class_count: usize,
-    prior: F,
-    feature_count: Array1<F>,
-    feature_log_prob: Array1<F>,
+pub struct MultinomialClassInfo<F> {
+    pub class_count: usize,
+    pub prior: F,
+    pub feature_count: Array1<F>,
+    pub feature_log_prob: Array1<F>,
```
YuhanLiin, line 209: Make all of these `pub(crate)`

bytesnake:
continued in #242

Issue #199, opened 2022-02-16
Random Forest and Ensemble Learning


jk1015:
I've recently been looking into adding Random Forest to linfa. Since Ensemble Learning is on the roadmap anyway I think the best way to do this would be to add Bootstrap Aggregation for any classifier rather than specialising the implementation to Decision Trees. I'm not totally sure what the design of this should look like though, especially since there don't seem to be any fixed conventions for implementing classifiers in linfa.

Would general bootstrap aggregation be a useful addition? If so I'm interested in other's opinions on how this should interface with existing/future classifiers in linfa along with any other design considerations.

YuhanLiin:
In [`impl_dataset.rs`](https://github.com/rust-ml/linfa/blob/master/src/dataset/impl_dataset.rs#L454) we already have bootstrap aggregation code that produces sub-samples from a dataset. We just need a generalized way of fitting classifiers over the subsamples. We have the trait `linfa::traits::Fit` that represents the training of a model using a set of hyperparameters, and we have `linfa::traits::PredictInplace` representing prediction using a trained model. You can define a new ensemble classifier that's generic over these traits, similar to how [`cross_validate`](https://github.com/rust-ml/linfa/blob/master/src/dataset/impl_dataset.rs#L911) is defined. Its `Fit` impl fits its "inner" classifier/regressor multiple times over the subsamples, and its `Predict` impl averages/votes on predictions made across its inner models.

EricTulowetzke:
Here is a WIP PR for RF

https://github.com/rust-ml/linfa/pull/43

YuhanLiin:
The work for that PR for ensemble learning ended up in #66 which didn't pan out for some reason. The current work is in #229.

PR #229, opened 2022-07-07
Bootstrap Aggregation

jk1015:
Following on from the discussion in issue #199 here is an implementation of Bootstrap Aggregation following the approach laid out there. There is also an example showing how the provided methods can be used together with the existing linfa-trees package to implement a Random Forest. 

Outside of the linfa-ensemble package the only required change was to factor the trait FromTargetArray to create an additional trait FromTargetArrayOwned. This was done to allow owned data to be used without needing to specify a lifetime, which was impossible under the previous FromTargetArray trait.

### DIFF
--- a/algorithms/linfa-ensemble/src/ensemble.rs
+++ b/algorithms/linfa-ensemble/src/ensemble.rs
@@ -1,0 +1,153 @@
+use linfa::{
+    dataset::{AsTargets, AsTargetsMut, FromTargetArrayOwned, Records},
+    error::{Error},
+    traits::*,
+    DatasetBase,
+};
+use ndarray::{
+    Array2, Axis, Array, Dimension
+};
+use std::{
+    cmp::Eq,
+    collections::HashMap,
+    hash::Hash,
+};
+
+pub struct EnsembleLearner<M> {
+    pub models: Vec<M>,
+}
+
+impl<M> EnsembleLearner<M> {
+
+    // Generates prediction iterator returning predictions from each model
+    pub fn generate_predictions<'b, R: Records, T>(&'b self, x: &'b R) -> impl Iterator<Item = T> + 'b
+    where M: Predict<&'b R, T> + PredictInplace<R, T> {
+        self.models.iter().map(move |m| m.predict(x))
+    }
+
+    // Consumes prediction iterator to return all predictions
+    // Orders predictions by total number of models giving that prediciton
+    pub fn aggregate_predictions<Ys: Iterator>(&self, ys: Ys)
+    -> impl Iterator<Item = Vec<(Array<<Ys::Item as AsTargets>::Elem, <<Ys::Item as AsTargets>::Ix as Dimension>::Smaller >, usize)>>
+    where
+        Ys::Item: AsTargets,
+        <Ys::Item as AsTargets>::Elem: Copy + Eq + Hash,
+    {
+        let mut prediction_maps = Vec::new();
+
+        for y in ys {
+            let targets = y.as_targets();
+            let no_targets = targets.shape()[0];
+
+            for i in 0..no_targets {
+                if prediction_maps.len() == i {
+                    prediction_maps.push(HashMap::new());
+                }
+                *prediction_maps[i].entry(y.as_targets().index_axis(Axis(0), i).to_owned()).or_insert(0) += 1;
+            }
+        }
+
+        prediction_maps.into_iter().map(|xs| {
+            let mut xs: Vec<_> = xs.into_iter().collect();
+            xs.sort_by(|(_, x), (_, y)| y.cmp(x));
+            xs
+        })
+    }
+}
+
+impl<F: Clone, T, M>
+PredictInplace<Array2<F>, T> for EnsembleLearner<M>
+where
+    M: PredictInplace<Array2<F>, T>,
+    <T as AsTargets>::Elem: Copy + Eq + Hash,
+    T: AsTargets + AsTargetsMut<Elem = <T as AsTargets>::Elem>,
+{
+    fn predict_inplace(&self, x: &Array2<F>, y: &mut T) {
+        let mut y_array = y.as_targets_mut();
+        assert_eq!(
+            x.nrows(),
+            y_array.len(),
+            "The number of data points must match the number of output targets."
+        );
+
+        let mut predictions = self.generate_predictions(x);
+        let aggregated_predictions = self.aggregate_predictions(&mut predictions);
+
+        for (target, output) in y_array.axis_iter_mut(Axis(0)).zip(aggregated_predictions.into_iter()) {
+            for (t, o) in target.into_iter().zip(output[0].0.iter()) {
+                *t = *o;
+            }
+        }
+    }
+
+    fn default_target(&self, x: &Array2<F>) -> T {
+        self.models[0].default_target(x)
+    }
+}
+
+pub struct EnsembleLearnerParams<P> {
+    pub ensemble_size: usize,
+    pub bootstrap_proportion: f64,
+    pub model_params: Option<P>,
+}
+
+impl<P> EnsembleLearnerParams<P> {
+    pub fn new() -> EnsembleLearnerParams<P> {
+        EnsembleLearnerParams {
+            ensemble_size: 1,
+            bootstrap_proportion: 1.0,
+            model_params: None,
+        }
+    }
+
+    pub fn ensemble_size(&mut self, size: usize) -> &mut EnsembleLearnerParams<P> {
+        self.ensemble_size = size;
+        self
+    }
+
+    pub fn bootstrap_proportion(&mut self, proportion: f64) -> &mut EnsembleLearnerParams<P> {
+        self.bootstrap_proportion = proportion;
+        self
+    }
+
+    pub fn model_params(&mut self, params: P) -> &mut EnsembleLearnerParams<P> {
+        self.model_params = Some(params);
+        self
+    }
+}
+
+impl<D, T, P: Fit<Array2<D>, T::Owned, Error>>
+     Fit<Array2<D>, T, Error> for EnsembleLearnerParams<P>
+where
+    D: Clone,
+    T: FromTargetArrayOwned,
+    T::Elem: Copy + Eq + Hash,
+    T::Owned: AsTargets,
+{
+    type Object = EnsembleLearner<P::Object>;
+
+    fn fit(&self, dataset: &DatasetBase<Array2<D>, T>) -> Result<Self::Object, Error> {
+        assert!(
+            self.model_params.is_some(),
+            "Must define an underlying model for ensemble learner",
+        );
+
+        let mut models = Vec::new();
+        let rng =  &mut rand::thread_rng();
+
+        let dataset_size = ((dataset.records.shape()[0] as f64) * self.bootstrap_proportion) as usize;
+
+        let iter = dataset.bootstrap_samples(dataset_size, rng);
+
+        for train in iter {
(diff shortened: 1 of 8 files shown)


algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -0,0 +1,153 @@
...
+
+pub struct EnsembleLearner<M> {
+    pub models: Vec<M>,
+}
+
+impl<M> EnsembleLearner<M> {
+
+    // Generates prediction iterator returning predictions from each model
+    pub fn generate_predictions<'b, R: Records, T>(&'b self, x: &'b R) -> impl Iterator<Item = T> + 'b
+    where M: Predict<&'b R, T> + PredictInplace<R, T> {
```
YuhanLiin, line 24: Don't need `PredictInplace` bound here

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    fn default_target(&self, x: &Array2<F>) -> T {
+        self.models[0].default_target(x)
+    }
+}
+
+pub struct EnsembleLearnerParams<P> {
+    pub ensemble_size: usize,
+    pub bootstrap_proportion: f64,
+    pub model_params: Option<P>,
```
YuhanLiin, line 91: This field shouldn't be optional and should be provided by the constructor

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn fit(&self, dataset: &DatasetBase<Array2<D>, T>) -> Result<Self::Object, Error> {
+        assert!(
+            self.model_params.is_some(),
+            "Must define an underlying model for ensemble learner",
+        );
+
+        let mut models = Vec::new();
+        let rng =  &mut rand::thread_rng();
+
+        let dataset_size = ((dataset.records.shape()[0] as f64) * self.bootstrap_proportion) as usize;
```
YuhanLiin, line 138: Need to ensure this value isn't 0

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -0,0 +1,153 @@
...
+    type Object = EnsembleLearner<P::Object>;
+
+    fn fit(&self, dataset: &DatasetBase<Array2<D>, T>) -> Result<Self::Object, Error> {
+        assert!(
+            self.model_params.is_some(),
+            "Must define an underlying model for ensemble learner",
+        );
+
+        let mut models = Vec::new();
+        let rng =  &mut rand::thread_rng();
```
YuhanLiin, line 136: Make the RNG a field of `EnsembleLearnerParams` (something like `R: impl Rng`). This allows us to use seeded RNGs for repeatable tests and benchmarks.

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    fn default_target(&self, x: &Array2<F>) -> T {
+        self.models[0].default_target(x)
+    }
+}
+
+pub struct EnsembleLearnerParams<P> {
+    pub ensemble_size: usize,
```
YuhanLiin, line 89: Why do we need a separate field for `ensemble_size`? Isn't this value implied by `bootstrap_proportion`?

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -0,0 +1,153 @@
...
+    M: PredictInplace<Array2<F>, T>,
+    <T as AsTargets>::Elem: Copy + Eq + Hash,
+    T: AsTargets + AsTargetsMut<Elem = <T as AsTargets>::Elem>,
+{
+    fn predict_inplace(&self, x: &Array2<F>, y: &mut T) {
+        let mut y_array = y.as_targets_mut();
+        assert_eq!(
+            x.nrows(),
+            y_array.len(),
+            "The number of data points must match the number of output targets."
```
YuhanLiin, line 70: You should be comparing against `y_array.nrows()`, which is the number of output data points, not the number of targets (that would be `ncols()`)

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -0,0 +1,153 @@
...
+        );
+
+        let mut predictions = self.generate_predictions(x);
+        let aggregated_predictions = self.aggregate_predictions(&mut predictions);
+
+        for (target, output) in y_array.axis_iter_mut(Axis(0)).zip(aggregated_predictions.into_iter()) {
+            for (t, o) in target.into_iter().zip(output[0].0.iter()) {
+                *t = *o;
+            }
+        }
```
YuhanLiin, line 80: Replace with this:
```rust
// prediction map has same shape as y_array, but the elements are maps
let mut prediction_maps = y_array.map(|_| HashMap::new());

for prediction in predictions {
  let p_arr = prediction.as_targets();
  assert_eq!(p_arr.shape(), y_array.shape());
  // Insert each prediction value into the corresponding map
  Zip::from(&mut prediction_maps).and(&p_arr).for_each(|(&mut map, &val)| map.entry(val).or_insert(0) += 1);
}

// For each prediction, pick the result with the highest number of votes
y_array = prediction_maps.mapv_into(|map| map.iter().max_by_key(|(_, v)| v).0);
```
It picks out the predictions with the highest number of votes without the complexity of `aggregate_predictions`

algorithms/linfa-ensemble/src/ensemble.rs
jk1015: Removing as suggested.

algorithms/linfa-ensemble/src/ensemble.rs
jk1015: Removing option as suggested and setting in constructor. Also removing builder method.

algorithms/linfa-ensemble/src/ensemble.rs
jk1015: I think other fit methods assume that the dataset is non-empty. So I am going to solve this by ensuring bootstrap_proportions is greater than 0 and then taking the ceiling.

YuhanLiin: Sounds good

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -85,39 +87,45 @@ where
...
             bootstrap_proportion: 1.0,
-            model_params: None,
+            model_params: model_params,
+            rng: rng
         }
     }
 
-    pub fn ensemble_size(&mut self, size: usize) -> &mut EnsembleLearnerParams<P> {
+    pub fn ensemble_size(&mut self, size: usize) -> &mut EnsembleLearnerParams<P, R> {
+        assert!(size > 0, "ensemble_size cannot be less than 1. Ensembles must consist of at least one model.");
```
YuhanLiin, line 114: In `linfa` we generally use the consuming builder pattern rather than the `&mut` builder pattern. Also, instead of using `assert` to check param values can you use the `ParamGuard` pattern? You can use [the DBSCAN params](https://github.com/rust-ml/linfa/blob/master/algorithms/linfa-clustering/src/dbscan/hyperparams.rs#L63) as an example. The gist is that you use the `ParamGuard` trait to convert your current param type to a validated and immutable newtype of the same params, and then implement `Fit` on the newtype.  This allows param validation to be done without panicking.

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -85,39 +87,45 @@ where
...
         self.ensemble_size = size;
         self
     }
 
-    pub fn bootstrap_proportion(&mut self, proportion: f64) -> &mut EnsembleLearnerParams<P> {
+    pub fn bootstrap_proportion(&mut self, proportion: f64) -> &mut EnsembleLearnerParams<P, R> {
+        assert!(proportion > 0.0, "bootstrap_proportion must be greater than 0. Must provide some data to each model.");
```
YuhanLiin, line 120: Proportion should also be <1.0

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -66,8 +68,8 @@ where
         let mut y_array = y.as_targets_mut();
         assert_eq!(
             x.nrows(),
-            y_array.len(),
-            "The number of data points must match the number of output targets."
+            y_array.len_of(Axis(0)),
```
YuhanLiin, line 71: change to `nrows`

algorithms/linfa-ensemble/src/ensemble.rs
```diff
@@ -127,20 +135,16 @@ where
...
-            self.model_params.is_some(),
-            "Must define an underlying model for ensemble learner",
-        );
 
         let mut models = Vec::new();
-        let rng =  &mut rand::thread_rng();
+        let mut rng = self.rng.clone();
 
-        let dataset_size = ((dataset.records.shape()[0] as f64) * self.bootstrap_proportion) as usize;
+        let dataset_size = ((dataset.records.shape()[0] as f64) * self.bootstrap_proportion).ceil() as usize;
```
YuhanLiin, line 142: Replace `shape()[0]` with `nrows()`

algorithms/linfa-ensemble/src/ensemble.rs
jk1015: This doesn't seem to be possible since I've written this to be as general as possible in the dimensionality of y. As far as I can tell `nrows` is only defined for 2d arrays -- `ArrayBase<S, Ix2>` -- while `y_array` here has type `ArrayBase<ViewRepr<&mut <T as AsTargets>::Elem>, <T as AsTargetsMut>::Ix>`

YuhanLiin: OK makes sense

algorithms/linfa-ensemble/src/ensemble.rs
jk1015: `ensemble_size` gives the number of models in the ensemble while `bootstrap_proportion` gives the proportion of the total number of training samples that should be given to each model for training. These should be distinct parameters.

YuhanLiin: Shouldn't `bootstrap_proportion` be the same as `1/ensemble_size`?

jk1015: Not necessarily, each model in the ensemble just needs its own random set of samples of training data from the complete training data set. There are no constraints on the size of this set other than it being non-empty, so we let the user tune this size as a hyperparameter.

YuhanLiin: OK so `bootstrap_samples` just grabs random sets of samples from the input and yields them infinitely. I thought it divided the input into random subsamples. This makes sense now.

YuhanLiin: Can you add this behaviour to the docs, along with a general description of `EnsembleLearner`? We should also have top level docs in `src/lib.rs` like with the other crates.

algorithms/linfa-ensemble/src/ensemble.rs
YuhanLiin: This comment still applies I believe.

HridayM25:
Hi!
Could you please guide me as to what is remaining in this?
Thank You!

YuhanLiin:
Merge/rebase with the latest master and address the open review comments. That's pretty much it.

PR #231, opened 2022-07-28
Fixed p norm creation

jorgehermo9:
I could not find a way to create an arbitrary p-norm distance function, it is implemented but if I'm not mistaken, the struct cannot be initialized from outside of the module due to private field on tuple struct `LpDist`.

[compiler error](https://imgur.com/a/zsUwqh7)

The solution is to make that field public, or provide a public `new` function. 

I'm ok with allowing both ways (pub field and new method).


YuhanLiin:
I think you have tabs instead of spaces as your indents, which is why `rustfmt` is failing. Just fix that and you're good.

YuhanLiin (approved):

YuhanLiin:
Alternatively you could also just run `cargo fmt` on the failing file

jorgehermo9:
Done! Thanks for the review

PR #233, opened 2022-08-01
Support for Hamming distance (l0 norm) 

jorgehermo9:
Currently, Lp distance when using p=0 is broken. It tries to calculate 1/0. I have implemented [Hamming Distance](https://en.wikipedia.org/wiki/Lp_space#When_p_=_0)  (l0 norm) that counts the number of positions which have different values.

Also, I changed the distance tests to increase code readability and norm symmetry/homogeneity checks.

### DIFF
--- a/algorithms/linfa-nn/src/distance.rs
+++ b/algorithms/linfa-nn/src/distance.rs
@@ -67,7 +67,7 @@ impl<F: Float> Distance<F> for L2Dist {
     }
 }
 
-/// L-infinte or [Chebyshev](https://en.wikipedia.org/wiki/Chebyshev_distance) distance
+/// L-infinite or [Chebyshev](https://en.wikipedia.org/wiki/Chebyshev_distance) distance
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct LInfDist;
 impl<F: Float> Distance<F> for LInfDist {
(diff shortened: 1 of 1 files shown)


YuhanLiin:
IDK if it should be referred to as L0 norm, since the Wiki article states that Hamming distance is not technically a "real" norm. I think it'd be better to rename it to something like `Hamming` and leave the 0-case for `Lp` distance broken, since we don't technically support the real "L0" norm.

algorithms/linfa-nn/src/distance.rs
```diff
@@ -77,6 +77,21 @@ impl<F: Float> Distance<F> for LInfDist {
...
 }
 
+/// L0 or [Hamming](https://en.wikipedia.org/wiki/Hamming_distance)
+#[derive(Debug, Clone, PartialEq)]
+pub struct L0Dist;
+impl<F: Float> Distance<F> for L0Dist {
+    #[inline]
+    fn distance<D: Dimension>(&self, a: ArrayView<F, D>, b: ArrayView<F, D>) -> F {
+        Zip::from(&a).and(&b).fold(F::zero(), |acc, &a, &b| {
+            if (a - b).is_zero() {
```
bytesnake, line 87: comparing two floats like this introduces unclear behavior when rounding errors occur

read in order 
    (1) https://stackoverflow.com/questions/5595425/what-is-the-best-way-to-compare-floats-for-almost-equality-in-python 
    (2) https://floating-point-gui.de/errors/comparison/ 
    (3) https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

and then use the https://docs.rs/approx/latest/approx/ crate to compare by absolute and relative epsilone

algorithms/linfa-nn/src/distance.rs
```diff
@@ -77,6 +77,21 @@ impl<F: Float> Distance<F> for LInfDist {
     }
 }
 
+/// L0 or [Hamming](https://en.wikipedia.org/wiki/Hamming_distance)
+#[derive(Debug, Clone, PartialEq)]
+pub struct L0Dist;
```
bytesnake, line 82: it kinda makes sense to continue 2/1/inf with zero norm, but like Yuhan said, it's not a real norm (and putting it into the framework of norms is misleading) - therefore distinguishing it from other norms by naming it after hamming code, makes it discrete behaviour more clear

algorithms/linfa-nn/src/distance.rs
```diff
@@ -177,6 +221,28 @@ mod test {
...
+
+        test_triangle_inequality(&dist);
+        test_symmetry(&dist);
+    }
+
+    #[test]
+    #[should_panic]
+    fn l0_dist_non_properties() {
+        let dist = L0Dist;
+        test_infinite_distance(&dist);
```
bytesnake, line 242: conflicting with homogeneity is actually never tested when infinite distance already panics :D

bytesnake (commented):
thanks for extending the distance bundle!

Issue #232, opened 2022-07-29
PCA works only for f64


jorgehermo9:
[PCA algorithm](algorithms/linfa-reduction/src/pca.rs] is only implemented for f64. Is there a reason for this? Can't it be implemented for trait Float?

EricTulowetzke:
I can take look at this over the next few days if no one else wants too

bytesnake:
This depends on https://github.com/rust-ndarray/ndarray-linalg/pull/275, this is already possible in the native Rust implementation but for fast LAPACK not yet

YuhanLiin:
While the `f32` impl would be broken when we use LAPACK, by default we use the pure-Rust linalg library instead, so `f32` should work. I'll see if I can enable `f32` for the pure-Rust library only without too much copypaste.

PR #234, opened 2022-08-07
Implement PCA for f32 and f64

YuhanLiin:
Addresses #232 

Adds PCA impl for f32 for all BLAS backends using macros. Does not involve the `Float` trait.

There is currently a logic bug, as the `test_explained_variance_diag` and `test_whitening_small` tests are currently failing on `f32`. It might be related to the `TruncateSvd` code. @bytesnake any ideas?

### DIFF
--- a/algorithms/linfa-reduction/src/pca.rs
+++ b/algorithms/linfa-reduction/src/pca.rs
@@ -21,6 +21,8 @@
 //! let dataset = embedding.predict(dataset);
 //! ```
 //!
+use std::marker::PhantomData;
+
 use crate::error::{ReductionError, Result};
 #[cfg(not(feature = "blas"))]
 use linfa_linalg::{lobpcg::TruncatedSvd, Order};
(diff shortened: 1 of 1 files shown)


YuhanLiin:
I also ran the tests with BLAS backend and `white_small` passes, but `explained_variance_diag` still fails.

bytesnake:
for `explained_variance_diag` this relates to https://github.com/rust-ml/linfa-linalg/pull/11 when decreasing to single-precision floats we can't extract as many eigenvalues without losing coherence

bytesnake:
the same occurs for `test_whitening_small`, it calculate covariance of 2x2 https://github.com/rust-ml/linfa-linalg/blob/main/src/lobpcg/svd.rs#L200

bytesnake:
I propose first fixing first https://github.com/rust-ml/linfa-linalg/pull/11 by adding the 1/5 rules to `TruncatedSvd` and `TruncatedEig` and then test again

Issue #48, opened 2020-09-27
Suggestion: Make serde an optional dependency


xd009642:
One suggestion to prevent bringing in more dependencies than a user needs would be to make serde an optional dependency and adjust the `Serialize` and `Deserialize` derives be like `#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]`. This would be a breaking change so would need a minor version update.

paulkoerbitz:
This sounds like a good idea, thanks for pointing this out! We're currently contributing a lot of changes for a 0.2 version, so the minor version update is not a problem IMO.

bytesnake:
we can't move the serde dependency behind a feature flag for `linfa-linear` and `linfa-logistic` because of `argmin`

stefan-k:
I just merged a [PR](https://github.com/argmin-rs/argmin/pull/180) in `argmin` which adds a feature that allows users to remove the `serde` dependency. This isn't released yet though and unfortunately  I can't make any promises yet as to when there will be a new release.

YuhanLiin:
Update: We just need to bump `argmin` in `linfa-logistic` and `linfa-ftrl` to 0.7, the same way it was done in `linfa-linear`.

relf:
Closed by #289

PR #236, opened 2022-08-12
Update argmin to version 0.6.0

stefan-k:
This PR updates argmin to the most recent version 0.6.0. In this version the `serde` dependency is optional, which addresses #48.
It only concerns `linfa-linear` and `linfa-logistic`.

# linfa-linear 

This one was fairly easy to do. I'm unsure what the exact reasons for introducing `ArgminParam` were. I managed to remove `ArgminParam` by adding the necessary trait bounds to `TweedyRegressorValidParams` which avoids having to re-implement the math traits. Let me know if I missed something and I'll introduce `ArgminParam` again.

# linfa-logistic

Here I did not remove `ArgminParam` here, but I think it should also be possible. 

**EDIT:**
I see now why you haven't upgraded to 0.5 ;). I guess it only makes sense to upgrade once the MSRV is increased. Also, I noticed that you want to avoid pulling in `ndarray-linalg` whenever possible. I somehow by accident removed the linalg feature from `argmin-math` and I'll have to see how to add that again (in the current setup it will always have `ndarray-linalg` as a dependency). 
I'm unsure how to proceed here. You can close this if you want but you can also leave this around for later ;)

### DIFF
--- a/algorithms/linfa-linear/Cargo.toml
+++ b/algorithms/linfa-linear/Cargo.toml
@@ -17,18 +17,27 @@ keywords = ["machine-learning", "linfa", "ai", "ml", "linear"]
 categories = ["algorithms", "mathematics", "science"]
 
 [features]
-blas = ["ndarray-linalg", "linfa/ndarray-linalg", "argmin/ndarray-linalg"]
+blas = ["ndarray-linalg", "linfa/ndarray-linalg"]
+serde = ["serde_crate", "ndarray/serde", "argmin-math/ndarray_v0_15-nolinalg-serde" ]
+
+[dependencies.serde_crate]
+package = "serde"
+optional = true
+version = "1.0"
+default-features = false
+features = ["std", "derive"]
 
 [dependencies]
 ndarray = { version = "0.15", features = ["approx"] }
 linfa-linalg = { version = "0.1", default-features = false }
 ndarray-linalg = { version = "0.15", optional = true }
 num-traits = "0.2"
-argmin = { version = "0.4.6", features = ["ndarray", "ndarray-rand"] }
-serde = { version = "1.0", default-features = false, features = ["derive"] }
+argmin = { version = "0.7.0", default-features = false }
+argmin-math = { version = "0.2.0", default-features = false, features = ["ndarray_v0_15-nolinalg"] }
+
 thiserror = "1.0"
 
-linfa = { version = "0.6.0", path = "../..", features=["serde"] }
+linfa = { version = "0.6.0", path = "../.." }
 
 [dev-dependencies]
 linfa-datasets = { version = "0.6.0", path = "../../datasets", features = ["diabetes"] }
--- a/algorithms/linfa-linear/src/float.rs
+++ b/algorithms/linfa-linear/src/float.rs
@@ -1,20 +1,12 @@
-use argmin::prelude::{ArgminAdd, ArgminDot, ArgminFloat, ArgminMul, ArgminNorm, ArgminSub};
-use ndarray::{Array1, NdFloat};
+use argmin::core::ArgminFloat;
+use ndarray::NdFloat;
 use num_traits::float::FloatConst;
 use num_traits::FromPrimitive;
-use serde::{Deserialize, Serialize};
 
 // A Float trait that captures the requirements we need for the various places
-// we need floats. There requirements are imposed y ndarray and argmin
+// we need floats. There requirements are imposed by ndarray and argmin
 pub trait Float:
-    ArgminFloat
-    + FloatConst
-    + NdFloat
-    + Default
-    + Clone
-    + FromPrimitive
-    + ArgminMul<ArgminParam<Self>, ArgminParam<Self>>
-    + linfa::Float
+    ArgminFloat + FloatConst + NdFloat + Default + Clone + FromPrimitive + linfa::Float
 {
     const POSITIVE_LABEL: Self;
     const NEGATIVE_LABEL: Self;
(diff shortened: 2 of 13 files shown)


algorithms/linfa-linear/src/float.rs
```diff
@@ -1,20 +1,12 @@
-use argmin::prelude::{ArgminAdd, ArgminDot, ArgminFloat, ArgminMul, ArgminNorm, ArgminSub};
-use ndarray::{Array1, NdFloat};
+use argmin::core::ArgminFloat;
+use ndarray::NdFloat;
 use num_traits::float::FloatConst;
 use num_traits::FromPrimitive;
-use serde::{Deserialize, Serialize};
 
 // A Float trait that captures the requirements we need for the various places
 // we need floats. There requirements are imposed y ndarray and argmin
```
YuhanLiin, line 7: Change `y` to `by`

algorithms/linfa-linear/src/glm/mod.rs
```diff
@@ -5,24 +5,32 @@ mod hyperparams;
...
 impl<F: Float, D: Data<Elem = F>, T: AsSingleTargets<Elem = F>>
     Fit<ArrayBase<D, Ix2>, T, LinearError<F>> for TweedieRegressorValidParams<F>
+where
+    Array1<F>: ArgminAdd<Array1<F>, Array1<F>>,
+    Array1<F>: ArgminSub<Array1<F>, Array1<F>>,
+    Array1<F>: ArgminMul<Array1<F>, Array1<F>>,
+    F: ArgminMul<Array1<F>, Array1<F>>,
+    Array1<F>: ArgminMul<F, Array1<F>>,
+    Array1<F>: ArgminDot<Array1<F>, F>,
+    Array1<F>: ArgminNorm<F>,
```
YuhanLiin, line 33: Are all these new bounds required? I still prefer these bounds over the extra `ArgminParam` code, since it's a net decrease in LOC, but I'd like to minimize the bounds as well.

algorithms/linfa-linear/Cargo.toml
```diff
@@ -17,15 +17,15 @@ keywords = ["machine-learning", "linfa", "ai", "ml", "linear"]
...
 linfa-linalg = { version = "0.1", default-features = false }
 ndarray-linalg = { version = "0.14", optional = true }
 num-traits = "0.2"
-argmin = { version = "0.4.6", features = ["ndarray", "ndarray-rand"] }
-serde = { version = "1.0", default-features = false, features = ["derive"] }
+argmin = { version = "0.6.0", default-features = false }
+argmin-math = { version = "0.1.0", features = ["ndarray_v0_15"] }
 thiserror = "1.0"
 
 linfa = { version = "0.6.0", path = "../..", features=["serde"] }
```
YuhanLiin, line 31: Remove `serde` feature

algorithms/linfa-linear/src/ols.rs
```diff
@@ -9,12 +9,11 @@ use linfa_linalg::qr::LeastSquaresQrInto;
 use ndarray::{concatenate, s, Array, Array1, Array2, ArrayBase, Axis, Data, Ix1, Ix2};
 #[cfg(feature = "blas")]
 use ndarray_linalg::LeastSquaresSvdInto;
-use serde::{Deserialize, Serialize};
 
 use linfa::dataset::{AsSingleTargets, DatasetBase};
 use linfa::traits::{Fit, PredictInplace};
 
-#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
```
YuhanLiin, line 17: Removing the mandatory `serde` dependency is great, but I'd also like to have `serde` support as an optional dependency. Can you introduce an optional `serde` feature into the affected crates like it's done in `linfa-clustering`? Also, instead of removing the `Serialize, Deserialize` annotations like it's done here, feature gate them behind the `serde` feature. You should also add the `linfa-linear/serde` and `linfa-logistic/serde` features to the Serde checking step in `.github/workflows/checking.yml`.

YuhanLiin:
MSRV will need to be bumped to 1.59 I think. Try upgrading to that version in the CI to see if it passes. I'm personally fine with this bump, since that version is already more than a year old. @bytesnake what do you think about bumping MSRV to 1.59?

I want to avoid pulling `ndarray-linalg` because it requires linking with external BLAS linalg libraries, which I want to place behind the `blas` feature. By default we use a pure-Rust linalg library instead. Thus, the dependency tree for a build with just default features should not include `ndarray-linalg` at all (this is enforced by the CI).

Can you also remove `ArgminParams` from `linfa-logistic`? You can probably remove the whole `argmin_params.rs` module. Also, I'd appreciate it if you can apply the same upgrade to `linfa-ftrl`, which also has `argmin` as a dependency. It doesn't need to be in this PR though.

algorithms/linfa-linear/src/glm/mod.rs
stefan-k: I added the traits which where implemented by `ArgminParam` and didn't check which ones where actually needed. Turns out that `Array1<F>: ArgminMul<Array1<F>, Array1<F>>` could be removed.

algorithms/linfa-linear/Cargo.toml
```diff
@@ -17,18 +17,26 @@ keywords = ["machine-learning", "linfa", "ai", "ml", "linear"]
...
 
 [dependencies]
 ndarray = { version = "0.15", features = ["approx"] }
 linfa-linalg = { version = "0.1", default-features = false }
 ndarray-linalg = { version = "0.14", optional = true }
 num-traits = "0.2"
-argmin = { version = "0.4.6", features = ["ndarray", "ndarray-rand"] }
-serde = { version = "1.0", default-features = false, features = ["derive"] }
+argmin = { version = "0.6.0", default-features = false }
+argmin-math = { version = "0.1.0", features = ["ndarray_v0_15"] }
```
YuhanLiin, line 36: Try `default-features = false, features = ["ndarray_0_15"]` to avoid pulling in `ndarray-linalg`. I'm not sure if you need the `primitive` feature as well.

algorithms/linfa-linear/Cargo.toml
```diff
@@ -17,18 +17,26 @@ keywords = ["machine-learning", "linfa", "ai", "ml", "linear"]
 categories = ["algorithms", "mathematics", "science"]
 
 [features]
-blas = ["ndarray-linalg", "linfa/ndarray-linalg", "argmin/ndarray-linalg"]
+blas = ["ndarray-linalg", "linfa/ndarray-linalg"]
```
YuhanLiin, line 20: Add `argmin/ndarray_v0_15`

algorithms/linfa-linear/Cargo.toml
```diff
@@ -17,18 +17,26 @@ keywords = ["machine-learning", "linfa", "ai", "ml", "linear"]
...
+serde = ["serde_crate", "ndarray/serde", "argmin-math/ndarray_v0_15-serde" ]
```
YuhanLiin, line 21: use `argmin-math/ndarray_0_15/serde-1` instead

stefan-k:
@YuhanLiin thanks for your reviews! However, I wasn't done yet with the first review ;) I'll request another one when I'm ready. 

Like I said in the edit of my initial post, I accidentially removed the possibility to turn off the `ndarray-linalg` dependency. Your suggestion to use `ndarray_0_15` instead unfortunately won't compile, because there are no feature gates on `ndarray-linalg` in the code. I could add those (which would not be a breaking change), but I'd rather expose the option to use or not use `ndarray-linalg` in the "high level" features. I'll have to think about how to solve this. I'll properly add the feature gates on `ndarray-lang`, bump the minor version and then add the higher level features in the next major version.

>Can you also remove ArgminParams from linfa-logistic? You can probably remove the whole argmin_params.rs module. Also, I'd appreciate it if you can apply the same upgrade to linfa-ftrl, which also has argmin as a dependency. It doesn't need to be in this PR though.

Sure, I'll do that. My time is a bit limited at the moment, therefore it may take a while until I'll get to it.

stefan-k:
I managed to fix the issue, but I'm not unsure if the fix is a breaking change. To be on the safe side I'll probably release 0.7 soon, hopefully in the course of next week and then I'll continue working on this PR. Sorry about the delay.

stefan-k:
Closing in favor of #289 and other future PRs.

PR #238, opened 2022-08-14
Adding Multi-Task ElasticNet support

YuhanLiin:
Continuation of #194


YuhanLiin:
Running the new multi-task example gives the following output:
```
intercept:  [182.11111111111111, 35.666666666666664, 55.55555555555556]
params: [[-0.9723742003724933, -0.12992938479472216, 0.20256364290951492],
 [0.017231919622246364, -0.00785311972200309, 0.006638074127064588],
 [0.0269082650844912, 0.021197761913871658, -0.027310155988705367]]
z score: Ok([[-1.0608739975723132, -1.6800812346255631, 2.2563865434388126],
 [0.018800267889146287, -0.10154653698276617, 0.0739424949093583],
 [0.029357297568142263, 0.274102444676571, -0.3042118890982639]], shape=[3, 3], strides=[3, 1], layout=Cc (0x5), const ndim=2)
predicted variance: [-47.348308658688005, -2.278139252532177, -50.96366188947603]
```
The variance looks pretty high, but I'm not sure if that's an issue.

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -189,6 +319,92 @@ fn coordinate_descent<'a, F: Float>(
...
+fn block_coordinate_descent<'a, F: Float>(
+    x: ArrayView2<'a, F>,
+    y: ArrayView2<'a, F>,
+    tol: F,
+    max_steps: u32,
+    l1_ratio: F,
+    penalty: F,
+) -> (Array2<F>, F, u32) {
+    let n_samples = F::cast(x.shape()[0]);
+    let n_features = x.shape()[1];
```
bytesnake, line 331: using `ncols` / `nrows` is a bit more expressive

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -189,6 +319,92 @@ fn coordinate_descent<'a, F: Float>(
...
+            let norm_old_w_j = old_w_j.dot(&old_w_j).sqrt();
+            if abs_diff_ne!(norm_old_w_j, F::zero()) {
+                general_mat_mul(
+                    F::one(),
+                    &x_j.view().insert_axis(Axis(1)),
+                    &old_w_j.view().insert_axis(Axis(0)),
+                    F::one(),
+                    &mut r,
+                );
+                //for i in 0..x.shape()[0] {
```
bytesnake, line 361: is this `general_mat_mul` previous version? If so then throw it away and add a comment before

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -189,6 +319,92 @@ fn coordinate_descent<'a, F: Float>(
...
+            if gap < tol {
+                break;
+            }
+        }
+    }
+
+    (w, gap, n_steps)
+}
+
+fn block_soft_thresholding<F: Float>(x: ArrayView1<F>, threshold: F) -> Array1<F> {
```
bytesnake, line 399: can you add a comment to the block soft thresholding operator (e.g. https://math.stackexchange.com/questions/2045579/deriving-block-soft-threshold-from-l-2-norm-prox-operator) so that people have a reference

algorithms/linfa-elasticnet/src/algorithm.rs
```diff
@@ -213,28 +429,71 @@ fn duality_gap<'a, F: Float>(
...
+    penalty: F,
+) -> F {
+    let half = F::cast(0.5);
+    let n_samples = F::cast(x.shape()[0]);
+    let l1_reg = l1_ratio * penalty * n_samples;
+    let l2_reg = (F::one() - l1_ratio) * penalty * n_samples;
+    let xta = x.t().dot(&r) - &w * l2_reg;
+
+    let dual_norm_xta = xta.map_axis(Axis(1), |x| x.dot(&x).sqrt()).norm_max();
+    let r_norm2 = r.iter().map(|rij| rij.powi(2)).sum();
```
bytesnake, line 453: `rij * rij` ?

algorithms/linfa-elasticnet/src/hyperparams.rs
```diff
@@ -12,19 +12,26 @@ use super::Result;
     derive(Serialize, Deserialize),
     serde(crate = "serde_crate")
 )]
-/// A verified hyper-parameter set ready for the estimation of a ElasticNet regression model
-///
-/// See [`ElasticNetParams`](crate::ElasticNetParams) for more informations.
 #[derive(Clone, Debug, PartialEq)]
-pub struct ElasticNetValidParams<F> {
+pub struct ElasticNetValidParamsBase<F, const MULTI_TASK: bool> {
```
bytesnake, line 16: should the multi task flag not be derived from the dataset?

bytesnake (commented):
:+1: can you point me to the test with high variance?

algorithms/linfa-elasticnet/src/hyperparams.rs
YuhanLiin: I should be able to encapsulate multi-task use case in another `Fit` impl on the same type

YuhanLiin:
> +1 can you point me to the test with high variance?

It's not a test, but the new example I added

algorithms/linfa-elasticnet/src/algorithm.rs
YuhanLiin: I feel like the current naming better expresses the properties of the dataset (# of rows = # of samples, # of cols = # of features). Plus this naming convention is used basically everywhere in this crate.

algorithms/linfa-elasticnet/src/hyperparams.rs
YuhanLiin: I just tried to do that and it didn't work. The single and multitarget `Fit` impls bound the target data type with `AsSingleTarget` and `AsMultiTarget` respectively. If I put both impls on a unified param type then I get a conflicting impl error, even though `AsSingleTarget` and `AsMultiTarget` are implemented on completely different types. This idea would be doable if we made the target types `Array1` and `Array2` instead of generics bounded by traits.

algorithms/linfa-elasticnet/src/algorithm.rs
bytesnake: nah I meant the method name, not the variable

algorithms/linfa-elasticnet/src/hyperparams.rs
bytesnake: the conflicting bounds are probably introduced here: https://github.com/rust-ml/linfa/blob/master/src/dataset/impl_targets.rs#L25-L26 have you also tried bounding the type with `T: AsTargets<Ix = Ix?>` directly?

YuhanLiin: Tried it. Still the same error. Rust compiler probably isn't smart enough to realize that those two bounds are completely disjoint.

bytesnake:
reviewed the example and made two changes to make the explained variance more usable

 1. use more than two samples for validation, otherwise the second class has zero variance
 2. compare validation dataset to estimated values (R2 is not symmetric)

```diff
--- a/algorithms/linfa-elasticnet/examples/multitask_elasticnet.rs
+++ b/algorithms/linfa-elasticnet/examples/multitask_elasticnet.rs
@@ -3,7 +3,7 @@ use linfa_elasticnet::{MultiTaskElasticNet, Result};

 fn main() -> Result<()> {
     // load Diabetes dataset
-    let (train, valid) = linfa_datasets::linnerud().split_with_ratio(0.90);
+    let (train, valid) = linfa_datasets::linnerud().split_with_ratio(0.80);

     // train pure LASSO model with 0.1 penalty
     let model = MultiTaskElasticNet::params()
@@ -18,7 +18,7 @@ fn main() -> Result<()> {

     // validate
     let y_est = model.predict(&valid);
-    println!("predicted variance: {}", valid.r2(&y_est)?);
+    println!("predicted variance: {}", y_est.r2(&valid)?);

     Ok(())
 }
```
which gives
```
predicted variance: [-4.143623744690414, -0.2630142563112303, -0.2542410304293199]
```
so worse than taking the average, but the dataset is really small 😅

PR #242, opened 2022-08-30
Add bernoulli naive bayes implementation

bytesnake:
Continuation of https://github.com/rust-ml/linfa/pull/226

### DIFF
--- a/algorithms/linfa-bayes/src/bernoulli_nb.rs
+++ b/algorithms/linfa-bayes/src/bernoulli_nb.rs
@@ -1,14 +1,14 @@
 use linfa::dataset::{AsSingleTargets, DatasetBase, Labels};
 use linfa::traits::{Fit, FitWith, PredictInplace};
 use linfa::{Float, Label};
-use ndarray::{Array1, ArrayBase, ArrayView2, Axis, Data, Ix2, OwnedRepr};
+use ndarray::{Array1, ArrayBase, ArrayView2, CowArray, Data, Ix2};
 use std::collections::HashMap;
 use std::hash::Hash;
 
-use crate::base_nb::{filter, NaiveBayes, NaiveBayesValidParams};
+use crate::base_nb::{NaiveBayes, NaiveBayesValidParams};
 use crate::error::{NaiveBayesError, Result};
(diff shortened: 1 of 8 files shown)


bytesnake:
> error: package `time v0.3.14` cannot be built because it requires rustc 1.59.0 or newer, while the currently active rustc version is 1.57.0

not sure which dependency causes this issue

YuhanLiin:
Seems like we need to bump MSRV to 1.59 (or limit the `time` dependency to a lower version)

algorithms/linfa-bayes/src/bernoulli_nb.rs
```diff
@@ -0,0 +1,285 @@
...
+    where
+        D: Data<Elem = F>,
+    {
+        let thr = self.binarize.unwrap_or_else(F::zero);
+        let xbin = x.map(|v| if v > &thr { F::one() } else { F::zero() });
+        if self.binarize.is_some() {
+            CowArray::from(xbin)
+        } else {
+            CowArray::from(x)
+        }
```
YuhanLiin, line 181: Can you write the logic like:
```rust
if let Some(thr) = self.binarize {
   ...
} else {
  ...
}
```

algorithms/linfa-bayes/src/hyperparams.rs
```diff
@@ -175,3 +175,104 @@ impl<F: Float, L> ParamGuard for MultinomialNbParams<F, L> {
...
+///
+/// See [`BernoulliNb`](crate::bernoulli_nb::BernoulliNb) for information on the model and [`BernoulliNbParams`](crate::hyperparams::BernoulliNbParams) for information on hyperparameters.
+#[derive(Debug, Clone, PartialEq)]
+pub struct BernoulliNbValidParams<F, L> {
+    // Required for calculation stability
+    alpha: F,
+    // Threshold for binarization
+    binarize: Option<F>,
+    // Phantom data for label type
+    label: PhantomData<L>,
```
YuhanLiin, line 189: Do we need this phantomdata?

relf:
Superseded by #388

PR #243, opened 2022-09-01
Update anyhow to version 1.0.58

gkobeaga:
The incompatibility was fixed for intel-mkl-tool https://github.com/rust-math/intel-mkl-src/issues/68

The current version of linfa is not compatible with `ndarray-linalg/intel-mkl-static`.

### DIFF
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -9,8 +9,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        toolchain: 
-          - 1.57.0
+        toolchain:
+          - 1.59.0
           - stable
           - nightly
         os:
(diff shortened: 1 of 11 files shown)


YuhanLiin:
I think you can remove the anyhow dependency altogether

gkobeaga:
Yes, much better. In order to maintain the commit history clean and get rid of this last commit, do you have any preference? Or is it ok if I add a new commit on top of it?

YuhanLiin:
New commit is fine. Also can you update the MSRV in CI to 1.59 to avoid the CI errors?

gkobeaga:
It seems that `ndarray-linalg` needs to be updated to version 0.15 (https://github.com/rust-ndarray/ndarray-linalg/pull/320), to support `intel-mkl-tool` version 0.3.0,  which fixes `use anyhow::*;` (https://github.com/rust-math/intel-mkl-src/pull/70).

I will update it and see if this does not break anything else.

YuhanLiin (approved):

PR #244, opened 2022-09-02
Fix `max_n_iterations` in k_means algorithm 

yottabytt:
**Issue**

If we set `max_n_iterations` as `i` and if the algo hasn't converged based on tolerance at `i`-th iteration, it errors with `KMeansError::NotConverged`.  Even if we don't set to a specific `i` and if it manages to reach `300`-th iteration (the default), it would still error i think. 

**Fix**

This patch fixes the issue and makes the behavior as expected from the already existing documentation. Also, have added a test.

### DIFF
--- a/algorithms/linfa-clustering/src/k_means/algorithm.rs
+++ b/algorithms/linfa-clustering/src/k_means/algorithm.rs
@@ -236,20 +236,17 @@ impl<F: Float, R: Rng + Clone, DA: Data<Elem = F>, T, D: Distance<F>>
 
         let mut min_inertia = F::infinity();
         let mut best_centroids = None;
-        let mut best_iter = None;
         let mut memberships = Array1::zeros(n_samples);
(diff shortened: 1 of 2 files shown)


algorithms/linfa-clustering/src/k_means/algorithm.rs
```diff
@@ -248,7 +248,8 @@ impl<F: Float, R: Rng + Clone, DA: Data<Elem = F>, T, D: Distance<F>>
                 self.init_method()
                     .run(self.dist_fn(), self.n_clusters(), observations, &mut rng);
             let mut converged_iter: Option<u64> = None;
-            for n_iter in 0..self.max_n_iterations() {
+            let mut n_iter = 0;
+            while n_iter < self.max_n_iterations() {
```
YuhanLiin, line 252: Use `loop` here, since we already have an exit condition at the bottom

YuhanLiin (approved):

Issue #245, opened 2022-09-13
Broken link in linfa-logistic documentation


quietlychris:
I was just scrolling through a part of of the `linfa-logistic`, and came across a documentation linking error for the link to https://docs.rs/linfa-logistic/latest/linfa_logistic/struct.LogisticRegression.html from the crate's index page. It might make sense to add a CI check to run a linkchecker over the `index.html` files created by `cargo doc` in each of the sub-crates to help keep an eye on this. 

One option for this could be the [`lychee`](https://github.com/lycheeverse/lychee) crate, which seems to have a number of installation options (although `cargo install` probably isn't a good option, since it has something like 400 deps to build from source).

gkobeaga:
Perhaps `#![deny(rustdoc::broken_intra_doc_links)]` can be set for the crates?

https://doc.rust-lang.org/rustdoc/lints.html?highlight=link#broken_intra_doc_links

gkobeaga:
Could it be closed?

YuhanLiin:
Yep

PR #246, opened 2022-09-28
Automatically generate the intra doc links

gkobeaga:
Addresses #245

The doc links are changed from `[comment](struct.StructName.html)` to `[comment](StructName)`.

I realized that the documentation for TweedieRegressor was not generating. To fix it and to match the public API of `linfa-linear` with the other crates (e.g. `KMeans` and `KMeansParams` from linfa_clustering), I made the following changes:
- The documentation for `TweedieRegressorValidParams` (which is not public), is moved for `TweedieRegressor`
- `TweedieRegressorParams` is now public

Small fixes:
- `FittedCountVectorizer` -> `CountVectorizer` in linfa-preprocessing

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,7 +2,7 @@
 on:
   push:
     branches:
-    - master 
+    - master
   pull_request:
 
 name: Build website with Zola, build rust docs and publish to GH pages
@@ -14,12 +14,18 @@ jobs:
     steps:
       - name: 'Checkout'
         uses: actions/checkout@master
+
       - name: 'Build only' 
         uses: shalzz/zola-deploy-action@master
         env:
           BUILD_DIR: docs/website/
           TOKEN: ${{ secrets.TOKEN }}
           BUILD_ONLY: true
+
+      - name: Build Documentation
+        run: cargo doc --workspace --all --no-deps
+        env:
+          RUSTDOCFLAGS: -D warnings
 
   build_and_deploy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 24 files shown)


YuhanLiin:
I think `TweedieRegressorValidParams` should also be made public.

Also, do you mind setting `#![deny(rustdoc::broken_intra_doc_links)]` for the crates?

gkobeaga:
Instead of adding  `#![deny(rustdoc::broken_intra_doc_links)]` for the crates, how about checking the broken links in the CI? I tested it with [c00c9af](https://github.com/rust-ml/linfa/pull/246/commits/c00c9af82fe5e037f6772975a1cc8e2b8afc4674), and the CI is failing now for that commit.

I also added the `--workspace` flag to `cargo doc` in the CI to build the documents of all the crates.

`TweedieRegressorValidParams` is public now, as you suggested.

.github/workflows/docs.yml
```diff
@@ -14,13 +14,19 @@ jobs:
...
+
       - name: 'Build only' 
         uses: shalzz/zola-deploy-action@master
         env:
           BUILD_DIR: docs/website/
           TOKEN: ${{ secrets.TOKEN }}
           BUILD_ONLY: true
 
+      - name: Build Documentation
+        run: cargo doc --workspace --all --no-deps
```
YuhanLiin, line 26: Delete the `-all` flag, which is redundant with the `--workspace` flag

.github/workflows/docs.yml
```diff
@@ -34,7 +40,9 @@ jobs:
           components: rustfmt, rust-src
 
       - name: Build Documentation
-        run: cargo doc --all --no-deps
+        run: cargo doc --workspace --all --no-deps
```
YuhanLiin, line 43: Here too

YuhanLiin (approved):

PR #249, opened 2022-10-17
Fix confusion matrix using only predictions as source for labels

levkk:
Fix confusing matrix incorrectly using labels from predict only instead of using labels from predict and ground truth. Ideally we should expose the Scikit-like API that passes in all the labels, in case the labels in the test set are not all inclusive (which would be a mistake in train/test partitioning, but can happen).

I'm somewhat confused by the way the API is written because the argument for the `confusion_matrix` method is called `ground_truth`, but shouldn't it be the predicted points instead?


YuhanLiin:
The argument is `ground_truth` because `self` is the predicted points. The point about using labels from both sources still stands though.

src/dataset/mod.rs
```diff
@@ -323,6 +323,18 @@ pub trait Labels {
     fn labels(&self) -> Vec<Self::Elem> {
         self.label_set().into_iter().flatten().collect()
```
YuhanLiin, line 324: For some reason this method doesn't dedup the final vector. It should do something like union all `HashSet` together. Or we can just the return type to `HashSet`, but that might be too invasive.

src/dataset/mod.rs
```diff
@@ -323,6 +323,18 @@ pub trait Labels {
...
     }
+
+    fn combined_labels(&self, other: Vec<Self::Elem>) -> Vec<Self::Elem> {
```
YuhanLiin, line 327: Better to have this method take `&impl Labels` or `&Self` as input. Then you can call `label_set` on both `self` and the input and union all the hashsets before converting it into a `Vec`.

songever:
> the confusion matrix doesn't work on Array: no method named `confusion_matrix` found for struct `ArrayBase` in the current scope method not found in `ArrayBase<OwnedRepr<u8>, Dim<[usize; 1]>>`

 let accuracies = dataset.cross_validate_single(5, &models, 
        |prediction: Array1<u8>, truth: Array1<u8>| {
            let cm = prediction.confusion_matrix(&truth).unwrap();
            Ok(cm.accuracy())
    })?;
    
    the method wasn't in the scope of closure?

PR #250, opened 2022-10-22
Add MAPE to regression metrics

bernardo-sb:
Add MAPE (mean absolute percentage error) to regression metrics

### DIFF
--- a/src/metrics_regression.rs
+++ b/src/metrics_regression.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 use ndarray::prelude::*;
 use ndarray::Data;
-use std::ops::{Sub, Div};
+use std::ops::{Div, Sub};
 
 /// Regression metrices trait for single targets.
 ///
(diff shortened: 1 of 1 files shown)


src/metrics_regression.rs
```diff
@@ -69,6 +69,16 @@ pub trait SingleTargetRegression<F: Float, T: AsSingleTargets<Elem = F>>:
         }
     }
 
+       /// Mean absolute percentage error between two continuous variables
+       fn mean_absolute_percentage_error(&self, compare_to: &T) -> Result<F> {
+        self.as_single_targets()
+            .sub(&compare_to.as_single_targets())
+            .mapv(|x| x.abs())
+            .div(&self.as_single_targets().mapv(|x| x.abs().max(F::min_positive_value())))
```
YuhanLiin, line 77: You can call `mapv(|x| x.abs())` after the `div` call and remove the `mapv` call inside the `div`, since you only need to call `abs` once. The `.max(F::min_positive_value())` part isn't necessary since you'll end up with `INF` either way if `self` is 0.

src/metrics_regression.rs
```diff
@@ -69,6 +69,16 @@ pub trait SingleTargetRegression<F: Float, T: AsSingleTargets<Elem = F>>:
         }
     }
 
+       /// Mean absolute percentage error between two continuous variables
```
YuhanLiin, line 72: Put the MAPE formula in the docs of both methods.

src/metrics_regression.rs
bernardo-sb: > You can call mapv(|x| x.abs()) after the div call and remove the mapv call inside the div, since you only need to call abs once.
Totally! Good call
>The .max(F::min_positive_value()) part isn't necessary since you'll end up with INF either way if self is 0.
Not exactly. Using `.max(F::min_positive_value())`results in very large number and not `inf`; conversely dividing by `0` results in `inf` - you can try it out in tests

YuhanLiin: IMO if `self` is 0 then the output should be `INF`, since you're literally dividing by 0

bernardo-sb: Changed. Allowing division by 0 and calling `mapv(|x| x.abs())` after `div`  and removed inner `mapv`

YuhanLiin (approved):

YuhanLiin:
Thanks for the MR

Issue #228, opened 2022-06-13
Compare BLAS and non-BLAS performance


YuhanLiin:
Now that we've made BLAS support optional on several `linfa` crates, we should compare the performance of those crates with and without BLAS. Doing this requires those crates to have a complete set of benchmarks that represent realistic workloads. If BLAS turns out to have no performance improvements, we can even remove BLAS support, improving code quality.

Benchmark status for each crate that supports BLAS:
- [ ] `linfa-clustering`:
  - [x] Benchmarks (pretty much complete, might want to add more workloads)
  - [ ] Compare BLAS and non-BLAS perf
- [ ] `linfa-ica`:
  - [ ] Benchmarks
  - [ ] Compare BLAS and non-BLAS perf
- [ ] `linfa-reduction`:
  - [ ] Benchmarks
  - [ ] Compare BLAS and non-BLAS perf
- [ ] `linfa-preprocessing`:
  - [x] Benchmarks
  - [ ] Compare BLAS and non-BLAS perf
- [ ] `linfa-pls`:
  - [ ] Benchmarks
  - [ ] Compare BLAS and non-BLAS perf
- [ ] `linfa-linear`:
  - [ ] Benchmarks
  - [ ] Compare BLAS and non-BLAS perf
- [ ] `linfa-elasticnet`:
  - [ ] Benchmarks
  - [ ] Compare BLAS and non-BLAS perf

I'm not 100% sure about removing the optional BLAS support, because that's technically a breaking change. That type of change would be unacceptable on a crate like `serde`, even on a breaking release, but we might be able to get away with it.

YuhanLiin:
Intel MKL libraries have their own ASM routines with architecture-specific optimizations, so it's not viable to completely reproduce in pure Rust. There's definitely space of improvement in the Rust library though.

ruchira:
Thank you, that's what I thought. Perhaps someday we can get Intel itself to provide a Rust + assembly version.

oojo12:
with guidance I am willing to do a blas vs non blas comparison for linfa-clustering.

YuhanLiin:
> with guidance I am willing to do a blas vs non blas comparison for linfa-clustering.

Since benchmarks already exist for `linfa-clustering`, run the benchmarks with and without the BLAS feature flags and post the results. Also post your commands and the specs of your machine.

oojo12:
@YuhanLiin Linfa Clustering results attached as zip.

Commands:

[...]

Issue #103, opened 2021-03-19
Improving and extending benchmarks


bytesnake:
One area where we are lacking right now is the benchmarking coverage. I would like to improve that in the coming weeks.

## Infrastructure for benchmarking

Benchmarks are an essential part of linfa. They should give feedback for contributors on their implementations and users confidence that we're doing good work. In order to automate the process we have to employ an CI system which creates a benchmark report on (a) PR (b) commits to master branch. This is [difficult](https://bheisler.github.io/post/benchmarking-in-the-cloud/) with wall-clock benchmarks (aka criterio.rs) but possible with [valgrind](https://pythonspeed.com/articles/consistent-benchmarking-in-ci/).

 * [ ] use [iai](https://github.com/bheisler/iai) for benchmarking
 * [ ] add a workflow executing the benchmark on PR/commits to master and create reports in JSON format
 * [ ] build a script parsing reports and posting it as comments to PR (see [here](https://stackoverflow.com/questions/58066966/commenting-a-pull-request-in-a-github-action))
 * [ ] add a page to the website which displays reports in a human-readable way
 * [ ] (pro) use polynomial regression to find influence of predictors (e.g. #weights, #features, #samples, etc.) to targets (e.g. L1 cache misses, cycles etc.) and post the algorithmic complexity as well

YuhanLiin:
While Iai is more consistent, it also hasn't been updated in almost 2 years and it also can't exclude setup code from the benchmarks. I'm also not sure how valid instruction-count benchmarks would be for multithread code. I believe we should stick with Criterion for future benchmarks. For most changes we can just do benchmark comparisons manually. For CI usage I'd rather set up a dedicated benchmarking machine like `rustc-perf`.

PR #253, opened 2022-10-26
Iai ols benchmark

oojo12:
This PR adds an iai benchmark for the OLS algorithm that is part of linfa-linear.

It has the following benchmark characteristics:

n = 1_000, 10_000, and 100_000
num_features = 7
values = Non unique (all are **f64:** 7)

If this is accepted, I will post the results in #228

This is also partly addresses #103


oojo12:
That's not a problem, though iai might be feature complete at this time it seems to use Valgrind and cachegrind on the backend for the benchmarking. I'm also not sure excluding setup code is more important than the consistency provided by iai.

If you still prefer criterion I can change it to that by next week and report the benchmark results on the other issue.

YuhanLiin:
Iai still has a bunch of open issues and PRs in its repo, so it's definitely not feature-complete. The author is still active so I'm not sure what the exact status of the project is. Setup code is not a big deal in longer benchmarks but could sway the percentage significantly in shorter benchmarks. Another concern with iai is whether it adds all executed cycles for multithread code, since that would make it seem as if everything was run sequentially.

I've asked the Zulip chat about this, so do you mind waiting on this for a week to see if anyone says anything? If I don't get a response then we go with Criterion.

oojo12:
Heard back from the author via email and thought I'd drop it here. 

According to Brook:


"""
Yeah, it's more or less dead. I certainly haven't had the energy or time to maintain it properly for a while now. I don't expect that to change in the near future either.
"""

YuhanLiin:
I think we should just go ahead and use Criterion.

oojo12:
Gotcha will do tomorrow.

datasets/src/generate.rs
```diff
@@ -66,21 +67,23 @@ fn make_blob(
...
     num_rows: usize,
     num_feats: usize,
-    num_targets: usize,
     feat_distr: X,
     target_distr: Y,
-) -> Dataset<f64, f64>
+) -> DatasetBase<
+    ArrayBase<OwnedRepr<f64>, Dim<[usize; 2]>>,
+    ArrayBase<OwnedRepr<f64>, Dim<[usize; 1]>>,
+>
```
YuhanLiin, line 80: You can put `Dataset<f64, f64, Ix1>` instead.

datasets/src/generate.rs
```diff
@@ -66,21 +67,23 @@ fn make_blob(
 /// use linfa_datasets::generate::make_dataset;
 /// let feat_distr = Laplace::new(0.5, 5. ).unwrap();
 /// let target_distr = DiscreteUniform::new(0, 5).unwrap();
-/// make_dataset(5, 5, 2, feat_distr, target_distr);
+/// make_dataset(5, 5, feat_distr, target_distr);
 /// ```
 pub fn make_dataset<X, Y>(
     num_rows: usize,
     num_feats: usize,
-    num_targets: usize,
```
YuhanLiin, line 74: You're changing this to single-target data right?

algorithms/linfa-linear/benches/ols_bench.rs
```diff
@@ -0,0 +1,28 @@
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use linfa::traits::Fit;
+use linfa_datasets::generate::make_dataset;
+use linfa_linear::LinearRegression;
+use statrs::distribution::{DiscreteUniform, Laplace};
+
+fn perform_ols(num_rows: usize) {
+    let feat_distr = Laplace::new(0.5, 5.).unwrap();
+    let target_distr = DiscreteUniform::new(0, 5).unwrap();
+    let num_feats: usize = 5;
```
YuhanLiin, line 10: Can you also run the benchmark with different numbers of features (don't make it too large though).

algorithms/linfa-linear/benches/ols_bench.rs
```diff
@@ -0,0 +1,28 @@
...
+    let dataset = make_dataset(num_rows, num_feats, feat_distr, target_distr);
```
YuhanLiin, line 11: Move `make_dataset` outside of this function so that it's not included in the benchmark times.

YuhanLiin:
Can you also add a benchmark for the GLM algorithm?

datasets/src/generate.rs
oojo12: will do.

datasets/src/generate.rs
oojo12: yes I am.

algorithms/linfa-linear/benches/ols_bench.rs
oojo12: Not a problem, something like 5, 10, 20 sound good? Just want to know what the standard should be.

oojo12:
yep. Just to make sure we are on the same page, following the glm.rs example will be sufficient?

algorithms/linfa-linear/benches/ols_bench.rs
YuhanLiin: Only really need 5 and 10

algorithms/linfa-linear/benches/ols_bench.rs
oojo12: gotcha.

datasets/src/generate.rs
YuhanLiin: Wouldn't it be more flexible to support multi-target generation? Single-target is just a special-case for multi-target. A target with dimensions `[n, 1]` can be converted into single-target.

YuhanLiin:
Yes GLM example is good

datasets/src/generate.rs
YuhanLiin: I suggest adding a dataset converter method for this in `linfa/src/dataset/impl_dataset.rs`

oojo12: I agree. At the time when I pushed the change I thought linfa algs only supported single_target datasets. What I can do is leave it as a multi-target generation and make the suggested converter method in another PR. 

The question is would you rather I finish this work first or the dataset converter method?

YuhanLiin: Dataset converter, which this work depends on.

YuhanLiin (approved):

PR #256, opened 2022-10-31
Target naming

oojo12:
An attempt to add target names per #70. Not really sure what to do with lines 83-98 in iter.rs. Any guidance there would be appreciated.

Line 83 was just added to see if it would build correctly.

### DIFF
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -30,6 +30,7 @@ impl<R: Records, S> DatasetBase<R, S> {
             targets,
             weights: Array1::zeros(0),
             feature_names: Vec::new(),
+            target_names: Vec::new(),
         }
     }
 
@@ -70,6 +71,16 @@ impl<R: Records, S> DatasetBase<R, S> {
         }
     }
 
+    /// Returns target names
+    ///
+    /// A target name gives a human-readable string describing the purpose of a single target.
+    pub fn target_names(&self) -> Vec<String> {
+        if !self.target_names.is_empty() {
+            self.target_names.clone()
+        } else {
+            "class"
+        }
+    }
     /// Return records of a dataset
     ///
     /// The records are data points from which predictions are made. This functions returns a
@@ -81,13 +92,14 @@ impl<R: Records, S> DatasetBase<R, S> {
     /// Updates the records of a dataset
     ///
     /// This function overwrites the records in a dataset. It also invalidates the weights and
-    /// feature names.
+    /// feature/target names.
     pub fn with_records<T: Records>(self, records: T) -> DatasetBase<T, S> {
         DatasetBase {
             records,
             targets: self.targets,
             weights: Array1::zeros(0),
             feature_names: Vec::new(),
+            target_names: Vec::new(),
         }
     }
 
@@ -100,6 +112,7 @@ impl<R: Records, S> DatasetBase<R, S> {
             targets,
             weights: self.weights,
             feature_names: self.feature_names,
+            target_names: self.target_names,
         }
     }
 
@@ -115,6 +128,15 @@ impl<R: Records, S> DatasetBase<R, S> {
         let feature_names = names.into_iter().map(|x| x.into()).collect();
 
         self.feature_names = feature_names;
+
+        self
+    }
+    
+    /// Updates the target names of a dataset
+    pub fn with_target_names<I: Into<String>>(mut self, names: Vec<I>) -> DatasetBase<R, S> {
+        let target_names = names.into_iter().map(|x| x.into()).collect();
+
+        self.target_names = target_names;
 
         self
     }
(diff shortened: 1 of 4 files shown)


src/dataset/mod.rs
```diff
@@ -164,6 +164,7 @@ impl Deref for Pr {
 /// * `targets`: a two-/one-dimension matrix with dimensionality (nsamples, ntargets)
 /// * `weights`: optional weights for each sample with dimensionality (nsamples)
 /// * `feature_names`: optional descriptive feature names with dimensionality (nfeatures)
+/// * `target_names`: optional descriptive target names with dimensionality (1)
```
YuhanLiin, line 167: Should be `(ntargets)` not `(1)`

src/dataset/iter.rs
```diff
@@ -77,16 +77,17 @@ where
         if self.target_or_feature && self.dataset.nfeatures() <= self.idx {
             return None;
         }
-
         let mut records = self.dataset.records.view();
         let mut targets = self.dataset.targets.as_targets();
         let feature_names;
+        let mut target_names = vec!["class".to_string()];
```
YuhanLiin, line 83: Just write `let mut target_names;`

src/dataset/iter.rs
```diff
@@ -77,16 +77,17 @@ where
...
         let weights = self.dataset.weights.clone();
 
         if !self.target_or_feature {
             // This branch should only run for 2D targets
             targets.collapse_axis(Axis(1), self.idx);
             feature_names = self.dataset.feature_names.clone();
+            target_names = self.dataset.target_names.clone();
```
YuhanLiin, line 90: This branch means we're iterating through the targets, so we should return each target name separately:
```
if self.data.target_names.is_empty() {
  target_names = Vec::new();
} else {
  target_names = vec![self.dataset.target_names[self.idx]];
}
```

src/dataset/iter.rs
```diff
@@ -77,16 +77,17 @@ where
...
         } else {
             records.collapse_axis(Axis(1), self.idx);
```
YuhanLiin, line 92: This branch means we're iterating through the features, so target names can be cloned:
```
target_names = self.dataset.target_names.clone();
```

src/dataset/mod.rs
```diff
@@ -333,6 +335,13 @@ mod tests {
     use ndarray::{array, Array1, Array2, Axis};
     use rand::{rngs::SmallRng, SeedableRng};
 
+    #[test]
+    fn set_target_name() {
```
YuhanLiin, line 339: Can you also add target name verification to one or two tests with `target_iter` just to verify that your iterator changes work correctly?

src/dataset/impl_dataset.rs
```diff
@@ -153,6 +166,20 @@ impl<L, R: Records, T: AsTargets<Elem = L>> DatasetBase<R, T> {
...
+    /// Returns target names
+    ///
+    /// A target name gives a human-readable string describing the purpose of a single target.
+    pub fn target_names(&self) -> Vec<String> {
+        if !self.target_names.is_empty() {
+            self.target_names.clone()
+        } else {
+            (0..self.ntargets())
+                .map(|idx| format!("class-{}", idx))
+                .collect()
```
YuhanLiin, line 182: If there are no names, this method should instead just return empty list. The signature should be `fn target_names(&self) -> &[String]` to prevent cloning. Apply this change to `feature_names` as well for consistency.

src/dataset/impl_dataset.rs
```diff
@@ -118,6 +121,15 @@ impl<R: Records, S> DatasetBase<R, S> {
...
     }
+
+    /// Updates the target names of a dataset
+    pub fn with_target_names<I: Into<String>>(mut self, names: Vec<I>) -> DatasetBase<R, S> {
+        let target_names = names.into_iter().map(|x| x.into()).collect();
+
+        self.target_names = target_names;
+
+        self
+    }
```
YuhanLiin, line 132: This method should check the length of the input vector so that it's equal to number of targets. The input should be Vec<String> so that it can be assigned directly. You can also try implementing this change for `feature_names`, but that might require more work to change.

src/dataset/mod.rs
oojo12: addressed lines 551-556 test this.

src/dataset/impl_dataset.rs
oojo12: the input for the function is already **Vec(I)** where **I** implements **Into(String)**. Unless I am mistaken you want something like?

```
pub fn with_target_names(mut self, names: Vec<String>) -> DatasetBase<R, S> {
        if names.len() == self.ntargets() {
            self.target_names = names;
        } else {
            // raise some error to user stating the number of targets is X or default to class_{0..ntargets}?
        }
        self
    }
```

YuhanLiin: Yeah pretty much. I don't mind panicking here, so you can just `assert` the condition.

oojo12: Gotcha

YuhanLiin:
This will close #248

PR #266, opened 2022-11-07
enable convert_targets method

### LINKED ISSUE #264: Add DataSet Converter Method
This method will be used to reshape the target dimensions from Ix2 to Ix1 and allow the make dataset function to be more generalizable.

oojo12:
Adds a converter method to Dataset<X,Y> type. It requires that the dimensions of the dataset be of the form [X, 1] to properly flatten.

Resolves #264

### DIFF
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -120,6 +120,16 @@ impl<R: Records, S> DatasetBase<R, S> {
     }
 }
 
+impl<X, Y> Dataset<X, Y> {
+    // Converts Ix2 targets to Ix1. Only works for targets with shape of form [X, 1]
+    pub fn convert_targets(self) -> Dataset<X, Y, Ix1> {
+        let nsamples = self.records.nsamples();
+        let targets = self.targets.into_shape(nsamples).unwrap();
+        let features = self.records;
+        Dataset::new(features, targets)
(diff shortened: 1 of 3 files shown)


src/dataset/impl_dataset.rs
```diff
@@ -120,6 +120,16 @@ impl<R: Records, S> DatasetBase<R, S> {
     }
 }
 
+impl<X, Y> Dataset<X, Y> {
+    // Converts Ix2 targets to Ix1. Only works for targets with shape of form [X, 1]
+    pub fn convert_targets(self) -> Dataset<X, Y, Ix1> {
```
YuhanLiin, line 125: Name this `into_single_target`

src/dataset/impl_dataset.rs
```diff
@@ -120,6 +120,16 @@ impl<R: Records, S> DatasetBase<R, S> {
     }
 }
 
+impl<X, Y> Dataset<X, Y> {
+    // Converts Ix2 targets to Ix1. Only works for targets with shape of form [X, 1]
```
YuhanLiin, line 124: `Convert 2D targets to 1D. Only works for targets with shape of form [X, 1], panics otherwise.`

YuhanLiin (approved):

PR #273, opened 2022-11-09
Add Pprof Profiling to Existing Benchmarks

### LINKED ISSUE #272: Add Profiling Hook to Criterion Benchmarks
Per #254 and #262 we can get ahead of these issues and any optimization related ones by adding a profiling hook to our benchmarks. This is a feature that Criterion [supports](https://bheisler.github.io/criterion.rs/book/user_guide/profiling.html#implementing-in-process-profiling-hooks). Additionally, the profiler will only be run when running with the `--profile-time` flag. I also found this [post](https://www.jibbow.com/posts/criterion-flamegraphs/#:~:text=Criterion%20is%20a%20well-known%20and%20often-used%20benchmarking%20framework,to%20use%20a%20custom%20profiler%20while%20running%20benchmarks.) on how to do so with [pprof](https://crates.io/crates/pprof). Pprof also produces flamegraphs if that is of interest.

oojo12:
Resolves #272.

By the way running the benchmarks in the background and will update #228 once if this is approved.


oojo12:
The windows failures are because pprof won't compile correctly there. I had to use wsl to compile. Is there something like bench_dependencies for cargo so we can resolve this? Tried googling but didn't find anything

oojo12:
Actually just found docs

YuhanLiin:
Do the benchmarks also output a text report?

oojo12:
actually, making the change still yielded compile time errors. So, I didn't push it but this is what is more or less was

```
[package]
name = "linfa-clustering"
version = "0.6.0"
edition = "2018"
authors = [
    "Luca Palmieri <rust@lpalmieri.com>",
    "xd009642 <danielmckenna93@gmail.com>",
    "Rémi Lafage <remi.lafage@onera.fr>"
]
description = "A collection of clustering algorithms"
license = "MIT/Apache-2.0"

repository = "https://github.com/rust-ml/linfa/"
readme = "README.md"

keywords = ["clustering", "machine-learning", "linfa", "k-means", "unsupervised"]

[features]
default = []
serde = ["serde_crate", "ndarray/serde", "linfa-nn/serde"]

[dependencies.serde_crate]
package = "serde"
optional = true
version = "1.0"
default-features = false
features = ["std", "derive"]

[dependencies]
ndarray = { version = "0.15", features = ["rayon", "approx"]}
linfa-linalg = { version = "0.1", default-features = false }
ndarray-linalg = { version = "0.15", optional = true }
ndarray-rand = "0.14"
ndarray-stats = "0.5"
num-traits = "0.2"
rand_xoshiro = "0.6"
space = "0.12"
thiserror = "1.0"
partitions = "0.2.4"
linfa = { version = "0.6.0", path = "../.." }
linfa-nn = { version = "0.6.0", path = "../linfa-nn" }
noisy_float = "0.2.0"

[dev-dependencies]
ndarray-npy = { version = "0.8", default-features = false }
linfa-datasets = { version = "0.6.0", path = "../../datasets", features = ["generate"] }
criterion = "0.4.0"
serde_json = "1"
approx = "0.4"
lax = "0.15.0"

[target.'cfg(not(windows))'.dev-dependencies]
pprof = { version = "0.11.0", features = ["flamegraph", "criterion"] }

[target.'cfg(not(windows))'.bench]
name = "k_means"
harness = false

[target.'cfg(not(windows))'.bench]
name = "dbscan"
harness = false

[target.'cfg(not(windows))'.bench]
name = "appx_dbscan"
harness = false

[target.'cfg(not(windows))'.bench]
name = "gaussian_mixture"
harness = false
```

Adding the corresponding cfg to the actual bench files didn't resolve the issue either.

algorithms/linfa-clustering/benches/gaussian_mixture.rs
```diff
@@ -46,9 +48,15 @@ fn gaussian_mixture_bench(c: &mut Criterion) {
...
 criterion_group! {
-  name = benches;
-  config = Criterion::default();
+  name = not_win_benches;
+  config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
   targets = gaussian_mixture_bench
 }
-criterion_main!(benches);
+#[cfg(target_os = "windows")]
+criterion_group!(win_benches, gaussian_mixture_bench);
```
YuhanLiin, line 58: If you give these two groups the same name, then you won't need to use any `cfg` below. Also insert a space after this line for readability.

YuhanLiin:
The zip file only contains the flamegraphs. I guess the profiler doesn't output any extra text right? Also, Criterion still outputs graphs just like before right?

As for the format, I still prefer flamegraphs since that's what I'm used to, but ideally I'd like to have both flamegraphs and the proto format.

YuhanLiin:
As part of this PR, can you also update the Linfa benchmark standards? We should do it once the profiling code is finalized.

oojo12:
> The zip file only contains the flamegraphs. I guess the profiler doesn't output any extra text right? Also, Criterion still outputs graphs just like before right?
> 
> As for the format, I still prefer flamegraphs since that's what I'm used to, but ideally I'd like to have both flamegraphs and the proto format.

It is one or the other you either output flamegraphs or use criterion as the measurement method. See `[here](https://bheisler.github.io/criterion.rs/book/user_guide/profiling.html#--profile-time)` for documentation evidence. As for the format I believe it is either flamegraph format or proto format. I don't think you can do both without running the benchmark twice. Not a problem to add just not sure if the value add is worth it.

oojo12:
> As part of this PR, can you also update the Linfa benchmark standards? We should do it once the profiling code is finalized.

yep once we agree on this PR being ready for merge I plan to.

algorithms/linfa-clustering/benches/gaussian_mixture.rs
YuhanLiin: I meant:
```rust
#[cfg(not(target_os = "windows"))]
criterion_group! {
    name = benches;
    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
    targets = appx_dbscan_bench
}
#[cfg(target_os = "windows")]
criterion_group!(benches, appx_dbscan_bench);
```

CONTRIBUTE.md
```diff
@@ -169,3 +170,34 @@ It is important to the project that we have benchmarks in place to evaluate the
...
+2. Power saving mode
+3. If the computer was idle during benchmark
+4. If the computer was overheating
+5. Hardware specs
+
+### Bench Command Run (Non Blas)
+bench results (code format)
+
+### Bench Command Run (Blas)
+bench results (code format)
```
YuhanLiin, line 200: We don't always want both BLAS and non-BLAS. It's mainly for #228.

oojo12: yeah that is why I added it since that is still on-going. Do you want this removed?

algorithms/linfa-clustering/benches/gaussian_mixture.rs
oojo12: ooo seems I just typod a lot and was mindless when I tested this bit. Making the change now.

CONTRIBUTE.md
YuhanLiin: Yeah I don't want to have to change it again later

oojo12: gotchu fam

algorithms/linfa-linear/benches/ols_bench.rs
```diff
@@ -57,5 +59,15 @@ fn bench(c: &mut Criterion) {
...
+    name = not_win_benches;
+    config = Criterion::default().with_profiler(PProfProfiler::new(100, Output::Flamegraph(None)));
+    targets = bench
+}
+#[cfg(not(target_os = "windows"))]
+criterion_main!(not_win_benches);
+#[cfg(target_os = "windows")]
+criterion_group!(win_benches, bench);
+#[cfg(target_os = "windows")]
+criterion_main!(win_benches);
```
YuhanLiin, line 73: Apply benchmark change to rest of the benchmarks.

YuhanLiin (approved):

PR #275, opened 2022-11-09
Rewrite binary labeling in logistic-regression

### LINKED ISSUE #252: linfa-logistic can't fit/predict very simple example
I originally posted this on the rust-ml zulip [here](https://rust-ml.zulipchat.com/#narrow/stream/233471-general/topic/linfa/near/305040135) and another user suggested I open an issue on this repo as it may be a bug.

------

I'm trying out linfa-logistic and wanted to get a simple case running, however the output confuses me and I'm not sure if I'm doing something wrong. I was attempting to have the model just map [1.0] to 1 and [0.0] to 0, but it seems to predict [1,1,1,1]? Any help would be appreciated :)
```
use linfa::{traits::{Fit, Predict}, DatasetBase};
use linfa_logistic::{LogisticRegression};
use ndarray::array;

fn main() {
    let x = array![[1.0], [0.0], [1.0], [0.0]];
    let y = array![1,0,1,0];
    let dataset = DatasetBase::new(x,y);
    let model = LogisticRegression::default().fit(&dataset).unwrap();

    println!("model: {:?}", model);
    println!();
    let pred = model.predict(&dataset.records);
    println!("pred: {:?}", pred);
}
```
````
model: FittedLogisticRegression { threshold: 0.5, intercept: 0.8989829201964533, params: [0.4214817977184791], shape=[1], strides=[1], layout=CFcf (0xf), const ndim=1, labels: [ClassLabel { class: 1, label: 1.0 }, ClassLabel { class: 0, label: -1.0 }] }

pred: [1, 1, 1, 1], shape=[4], strides=[1], layout=CFcf (0xf), const ndim=1
```

bytesnake:
Fixes #252


algorithms/linfa-pls/src/lib.rs
```diff
@@ -44,8 +44,6 @@ pub use pls_generic::*;
 mod pls_svd;
 mod utils;
 
-use crate::pls_generic::*;
```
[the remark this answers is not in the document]
YuhanLiin: Another currently-open PR changes this line to `pub use`, so I think you should revert this change to avoid a conflict.

algorithms/linfa-logistic/src/lib.rs
```diff
@@ -969,6 +977,17 @@ mod test {
         );
     }
 
+    #[test]
+    fn simple_example_3() {
```
YuhanLiin, line 981: Can you also add a multinomial version of this test to see if multinomila logistic regression suffers from the same issue?

algorithms/linfa-logistic/src/lib.rs
```diff
@@ -273,46 +273,54 @@ where
...
-            if *last_item != item {
-                use_negative_label = !use_negative_label;
+    let y = y.as_single_targets();
+
+    // find binary classes of our target dataset
+    let classes = y.iter().fold(Some((None, None)), |map, elm| match map {
+        Some((Some((n, val)), x)) if n == elm => Some((Some((n, val + 1)), x)),
+        Some((x, Some((n, val)))) if n == elm => Some((x, Some((n, val + 1)))),
+        Some((Some(x), None)) => Some((Some(x), Some((elm, 1)))),
+        Some((None, None)) => Some((Some((elm, 1)), None)),
```
YuhanLiin, line 283: I think we should make this 2-tuple into a simple struct with named fields, so that this code is more understandable.

YuhanLiin (approved):

Issue #295, opened 2023-03-08
linfa-clustering depends on partitions 0.2.4, which throws future-incompatibilities warnings


eric-seppanen:
When testing rust 1.68.0, I see the following warning:
```
warning: the following packages contain code that will be rejected by a future version of Rust: partitions v0.2.4
```
That crate is a dependency of `linfa-clustering`.

I would guess that the `partitions` crate is no longer maintained. It hasn't been touched in 5 years and there hasn't been any response to the [issue](https://github.com/DDOtten/partitions/issues/2).

YuhanLiin:
Resolved by #298

PR #298, opened 2023-03-14
Remove appx DBSCAN impl and replace with type aliases

YuhanLiin:
Addresses #295

Approx DBSCAN was slower than normal DBSCAN and also brings in `partitions`, which is an unmaintained crate that will not compile in a future version. Since there's not easy replacement for `partitions`, I decided to temporarily remove the Approx DBSCAN implementation and replace the public API with equivalent type aliases. The files for Approx DBSCAN have been kept behind in case we want to restore the implementation.


YuhanLiin:
@quietlychris Is this fine with you?

quietlychris (approved):
Just took a look (apologies for the delay!); this all seems good to me, and I think that leaving things as a placeholder is fine. 

I also looked at the `partitions` crate; it definitely does seem abandoned, and as far as I can tell, the maintainer (Daniel Otten) does not have any active online presence. It seems we'd probably need to fork and update `partitions` in the future if we wanted to use it again; I'd need to look into the standard convention in the Rust community for handling something like that, but it is an option if you'd like to go that route in the future.

YuhanLiin:
Approx DBSCAN has been slower than DBSCAN for a long time. If there's no replacement for `partitions` then I'd be willing to axe approx DBSCAN on the next breaking release.

Issue #299, opened 2023-03-14
Incompatibility of the partitions dependency with future versions of Rust


gilad-rosenblatt:
Would you be able to replac/update the `partitions v0.2.4` dependency for `linfa-clustering` so that I can continue using `linfa` with future versions of Rust? The existing version has some lint issues that are incompatible w/ future versions of Rust (see below).

# Description

While building my project I get a future incompatibility warning coming from `partitions v0.2.4` (a dependency introduced by `linfa-clustering`). 
```
warning: the following packages contain code that will be rejected by a future version of Rust: partitions v0.2.4
```

# Details

On running `cargo report future-incompatibilities --id 4` it seems the that `partitions v0.2.4` is expected to have some linting issues.
```The package `partitions v0.2.4` currently triggers the following future incompatibility lints:
> warning: trailing semicolon in macro used in expression position
>    --> /home/maticd/.cargo/registry/src/github.com-1ecc6299db9ec823/partitions-0.2.4/src/lib.rs:69:51
>     |
> 69  |         bit_vec::BitVec::from_elem($len, $element);
>     |                                                   ^
>     |
>    ::: /home/maticd/.cargo/registry/src/github.com-1ecc6299db9ec823/partitions-0.2.4/src/partition_vec.rs:528:24
>     |
> 528 |         let mut done = bit_vec![false; self.len()];
>     |                        --------------------------- in this macro invocation
>     |
>     = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
>     = note: for more information, see issue #79813 <https://github.com/rust-lang/rust/issues/79813>
>     = note: `#[allow(semicolon_in_expressions_from_macros)]` on by default
>     = note: this warning originates in the macro `bit_vec` (in Nightly builds, run with -Z macro-backtrace for more info)
> 
> warning: trailing semicolon in macro used in expression position
>     --> /home/maticd/.cargo/registry/src/github.com-1ecc6299db9ec823/partitions-0.2.4/src/lib.rs:69:51
>      |
> 69   |         bit_vec::BitVec::from_elem($len, $element);
>      |                                                   ^
>      |
>     ::: /home/maticd/.cargo/registry/src/github.com-1ecc6299db9ec823/partitions-0.2.4/src/partition_vec.rs:1297:19
>      |
> 1297 |             done: bit_vec![false; len],
>      |                   -------------------- in this macro invocation
>      |
[...]

PR #304, opened 2023-04-19
Reactivate appx DBSCAN impl with new disjoint set dependency

jogru0:
Hi everyone!

I saw the issues with the partitions dependency described in #295 and #299, and finally the removal of the appx DBSCAN in #298.

Coincidentally, I recently was looking for a disjoint set crate to use for myself, and when I wasn't satisfied with what I found on crates.io (all existing ones seem to be not maintained and/or not documented/not tested), I decided to publish my own: https://crates.io/crates/disjoint.

So it turned out there were just some small additional features necessary in my crate to enable it as a possible replacement for partitions for you. I implemented these now, and therefore can now provide this pull request to reactivate your appx DBSCAN.

I split the implementation in two commits:

1. Reverted #298. This just reintroduces the appx DBSCAN impl without any further changes.
2. Replaced dependencies: partitions -> disjoint.

If you just look at the second commit, you can see that it's really almost a drop-in replacement. Only a few lines change due to the slightly different interface.

### DIFF
--- a/algorithms/linfa-clustering/Cargo.toml
+++ b/algorithms/linfa-clustering/Cargo.toml
@@ -37,7 +37,7 @@ num-traits = "0.2"
 rand_xoshiro = "0.6"
 space = "0.12"
 thiserror = "1.0"
-#partitions = "0.2.4" This one will break in a future version of Rust and has no replacement
+partitions = "0.2.4"
 linfa = { version = "0.6.1", path = "../.." }
 linfa-nn = { version = "0.6.1", path = "../linfa-nn" }
 noisy_float = "0.2.0"
@@ -57,6 +57,10 @@ harness = false
 
 [[bench]]
 name = "dbscan"
+harness = false
+
+[[bench]]
+name = "appx_dbscan"
 harness = false
 
 [[bench]]
--- a/algorithms/linfa-clustering/Cargo.toml
+++ b/algorithms/linfa-clustering/Cargo.toml
@@ -37,7 +37,7 @@ num-traits = "0.2"
 rand_xoshiro = "0.6"
 space = "0.12"
 thiserror = "1.0"
-partitions = "0.2.4"
+disjoint = "0.5.0"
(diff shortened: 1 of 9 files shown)


quietlychris:
Any chance you're willing to run/post some benchmarks? I believe part of the rationale for removing Appx. DBSCAN, beyond the dependency issue, was that it didn't really add any new capability since the implementation itself was actually slower than just standard DBSCAN, leaving little practical reason why someone would choose to use it. @YuhanLiin, any thoughts?

jogru0:
So, I benchmarked my disjoint set implementation itself, as I wanted it to be not slower than existing alternatives, and I wanted to choose implementation details to optimize for speed.

For that, I looked at how fast Kruskal's algorithm runs on an example graph with 1 million vertices and ~8.8 million edges, already sorted by weight, with measures including creating the disjoint set data structure and allocating memory for the result Vec on demand.

The imo most fair comparison to the partition crate is to use DisjointSetVec<()> for that, even though you could use DisjointSet directly (which turns out to not even be faster as Rust handles Vec<T> with size_of<T>() == 0 very efficiently), and to not use the Quick-Union variant of the algorithm, as that's not possible with partition.

With these restrictions, on my local  AMD Ryzen 5 3600, I get the following results:

- partition: 290ms +/- 10ms
- disjoint: 57ms +/- 1ms

That being said, of course this is a very specific benchmark, and I assume Appx. DBSCAN is a very different algorithm, so this doesn't necessarily mean anything for the performance of that. I don't know anything about what Appx. DBSCAN even is, to be honest (I just drop-in replaced the old dependency with a new one), but my gut feeling would be that the performance of the disjoint set data structure doesn't even matter that much here, as there is so much other stuff happening. But I really don't know and probably someone with more knowledge about clusting algorithms should do benchmarks for that, if you are interested in that.

YuhanLiin:
For a quick comparison can you run the normal DBSCAN and Approx DBSCAN benchmarks and post the results? If Approx is slower than we should just get rid of it.

PR #309, opened 2023-07-12
Bump argmin to 0.8.1 and serde optional

### LINKED ISSUE #303: Bump `argmin` dependency to 0.8
Now that our MSRV is higher than 1.62, we can use `argmin` v0.8. This update will also allow `linfa-ftrl` and `linfa-logistic` to have `serde` as an optional feature.

pablf:
It should fix #303.


Congyuwang:
Can we have this merged?

quietlychris (approved):
These all make sense to me, and I don't see much of this being particularly controversial.

Issue #274, opened 2022-11-09
Convert linfa-preprocessing Benchmark to Criterion


oojo12:

oojo12:
@YuhanLiin Can add "help wanted" and good first issue" labels to this?

lucas-montes:
Hi, I've changed the whitening benchs to criterion. If it looks good I could open a PR and keep going with the rest of benches. You can find the commit in my fork at this link https://github.com/rust-ml/linfa/compare/master...lucas-montes:linfa:master

PR #314, opened 2023-08-16
 Convert linfa-preprocessing Benchmark to Criterion #274 

lucas-montes:
I currently only started with whitening to check if the changes are according to what is expected

### DIFF
--- a/algorithms/linfa-preprocessing/benches/whitening_bench.rs
+++ b/algorithms/linfa-preprocessing/benches/whitening_bench.rs
@@ -1,38 +1,47 @@
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion};
+use linfa::benchmarks::config;
 use linfa::traits::Fit;
 use linfa::traits::Transformer;
+use linfa_datasets::generate::make_dataset;
 use linfa_preprocessing::whitening::Whitener;
-use ndarray::Array2;
-use ndarray_rand::{
-    rand::distributions::Uniform, rand::rngs::SmallRng, rand::SeedableRng, RandomExt,
-};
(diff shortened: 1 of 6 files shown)


algorithms/linfa-preprocessing/benches/whitening_bench.rs
```diff
@@ -1,38 +1,47 @@
...
+        (Whitener::pca(), "pca"),
+    ] {
+        for nfeatures in (10..100).step_by(10) {
+            let dataset = make_dataset(size, nfeatures, 1, feat_distr, target_distr);
+            benchmark.bench_function(
+                BenchmarkId::new(fn_name, format!("{}x{}", nfeatures, size)),
+                |bencher| {
+                    bencher.iter(|| {
+                        whitener
+                            .fit(black_box(&dataset.clone()))
```
YuhanLiin, line 28: unnecessary clone

algorithms/linfa-preprocessing/benches/whitening_bench.rs
```diff
@@ -1,38 +1,47 @@
...
+                            .unwrap()
+                            .transform(black_box(dataset.clone()));
```
YuhanLiin, line 30: I think you can put `dataset.view()` here to prevent unnecessary clone

lucas-montes: Done

algorithms/linfa-preprocessing/benches/whitening_bench.rs
lucas-montes: Done

YuhanLiin (approved):

PR #325, opened 2023-12-22
Logistic Regression: Binary class labels getter

DGPardo:
Expose a reference to the mapping from the ground truth data label to the assigned binary label used in the logistic regression model.

This goes in line with the output provided by `MultiFittedLogisticRegression` that has a method to access `classes`

### DIFF
--- a/algorithms/linfa-logistic/src/lib.rs
+++ b/algorithms/linfa-logistic/src/lib.rs
@@ -613,6 +613,12 @@ impl<F: Float, C: PartialOrd + Clone> FittedLogisticRegression<F, C> {
         &self.params
     }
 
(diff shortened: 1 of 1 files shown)


quietlychris (changes requested):
Since these methods/structs are being made public, can you add doc comments for each? Otherwise this looks good, although you may need to rebase onto master. The CI isn't passing right now bc of what looks like a Windows linking issue, but I'll don't think that should gate this being merged.

DGPardo:
Thanks for your review @quietlychris 
I think I have addressed your requests, otherwise please let me know

quietlychris (approved):

quietlychris:
Looks good, thank you!

PR #327, opened 2024-01-02
Expose references to the PCA components and column means

hamishs:
Adds methods to `Pca` that return references to the fitted embedding matrix (named components) and the mean. The components are useful for understanding the relative importance of each feature in the decomposition. The mean is included for completeness as this is in addition to the singular values for which a method already exists.


quietlychris (approved):
Looks good, thank you!

PR #329, opened 2024-01-11
Add wasm-bindgen feature to enable argmin/wasm-bindgen

atsuik:
To complie linfa with wasm-bindgen and use it from Browser, I add a new feature flag `wasm-bindgen` to `Cargo.tom` for
- linfa-ftrl
- linfa-linear
- linfa-logistic
, which crates depend on `argmin` crate.

The new feature looks like this.
```
[features]
wasm-bindgen = ["argmin/wasm-bindgen"]
```

`argmin` has feature flag `wasm-bindgen` to use this crate with wasm-bindgen.


quietlychris (approved):
Cool to see this potentially getting used in a WASM environment!

PR #330, opened 2024-01-17
Improve docstring of LinearRegression

rainij:
- Fixed some "markdown errors".
- Made docstring of `LinearRegression` more consistent with the docstrings of `fit` and `predict` (and variable names in code).
- Removed comment on Moore-Penrose inverse because it does not really help to understand what the struct does. Instead it is explained more precisely what the goal of _ordinary_ least squares for linear models is (briefly).
- It is emphasized that the model is _univariate_ (scalar targets).


quietlychris (approved):
Looks good, thank you!

PR #332, opened 2024-02-17
Feat: implement Random Projections

GBathie:
This PR implements random projection techniques for dimensionality reduction, as seen in the [`sklean.random_projection` module of scikit-learn](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.random_projection)


quietlychris:
I've done a quick review; this looks good to me, but have requested @bytesnake also give it a look as he is probably more familiar with the algorithm side of things.

algorithms/linfa-reduction/src/random_projection/sparse/algorithms.rs
```diff
@@ -0,0 +1,101 @@
...
+            SparseRandomProjectionParamsInner::Dimension { target_dim } => *target_dim,
+            SparseRandomProjectionParamsInner::Precision { precision } => {
+                johnson_lindenstrauss_min_dim(n_samples, *precision)
+            }
+        };
+
+        let scale = (n_features as f64).sqrt();
+        let p = 1f64 / scale;
+        let dist = SparseDistribution::new(F::cast(scale), p);
+        let mut rng = thread_rng();
```
relf, line 39: You can pass a `Rng` as you do for `GaussianRandomProjection` in order to be able to strictly reproduce runs.

GBathie: I'm not sure what is the best way to do this.
The `fit` function of the `Transform` trait takes `&self` as parameter, so the only way to use the `Rng` is to `clone()` it. This means that using the same `GaussianProjectionParamters` struct twice in a row to create a projection of a given size will result in the **same** projection. Is this the expected behavior?

Another issue is that few algorithms in `linfa` can use a custom `Rng`, so you would only have reproducibility for the projection part.

relf: > Is this the expected behavior?

Yes, any algorithm using randomness should offer a way to be reproducible for debugging/testing purpose.  
For `Rng` what I have in mind was things like in `linfa-clustering`, check out `kmeans` or `gaussian mixture` for instance where the random generator is not an option.

algorithms/linfa-reduction/src/random_projection/common.rs
```diff
@@ -0,0 +1,11 @@
+/// Compute a safe dimension for a projection with precision `eps`,
+/// using the Johnson-Lindestrauss Lemma.
+///
+/// References:
+/// - [D. Achlioptas, JCSS](https://www.sciencedirect.com/science/article/pii/S0022000003000254)
+/// - [Li et al., SIGKDD'06](https://hastie.su.domains/Papers/Ping/KDD06_rp.pdf)
+pub(crate) fn johnson_lindenstrauss_min_dim(n_samples: usize, eps: f64) -> usize {
```
relf, line 7: You could add a test against values from [scikit-learn](https://scikit-learn.org/stable/modules/random_projection.html#the-johnson-lindenstrauss-lemma)

algorithms/linfa-reduction/src/random_projection/gaussian/algorithms.rs
```diff
@@ -0,0 +1,62 @@
...
+impl<F, Rec, T, R> Fit<Rec, T, ReductionError> for GaussianRandomProjectionValidParams<R>
+where
+    F: Float,
+    Rec: Records<Elem = F>,
+    R: Rng + Clone,
+    StandardNormal: Distribution<F>,
+{
+    type Object = GaussianRandomProjection<F>;
+
+    fn fit(&self, dataset: &linfa::DatasetBase<Rec, T>) -> Result<Self::Object, ReductionError> {
```
relf, line 28: Also add some tests at least to check dimension reduction, it could be in the documentation like in [scikit-learn](https://scikit-learn.org/stable/modules/random_projection.html#gaussian-random-projection).

bytesnake:
> I've done a quick review; this looks good to me, but have requested @bytesnake also give it a look as he is probably more familiar with the algorithm side of things.

thank you for reviewing @relf @quietlychris

algorithms/linfa-reduction/src/random_projection/common.rs
GBathie: Thank you for the suggestion, I added tests and found a bug!

algorithms/linfa-reduction/src/random_projection/sparse/algorithms.rs
GBathie: Updated in [3569c10](https://github.com/rust-ml/linfa/pull/332/commits/3569c1035f83e1818704eb14d582ee143268a3e4), both structs now have RNGs; this field is no longer an option, and `Xoshiro256Plus` is used by default if not provided.

algorithms/linfa-reduction/src/random_projection/gaussian/algorithms.rs
GBathie: Added in [a5f1ad3](https://github.com/rust-ml/linfa/pull/332/commits/a5f1ad3957775e90c9a898e7ce2d1ff02f26653e).

GBathie:
Thank you for the reviews, and @relf for the suggestions, I have implemented them.

Changes:
- The `rng` field for both random projections structs  is no longer optional, and defaults to Xoshiro256Plus with a fixed seed if not provided by the user.
- Renamed `precision` parameter to `eps`, as increasing this parameter results in a lower dimension embedding and often yields lower classification performance.  
- Added a check that the projections reduce the dimension of the data, returning an error otherwise. Added tests for this behavior.
- Added a test for the function that computes the embedding dimension from a given epsilon, against values from scikit-learn .
- Fixed reference issue in docs and other minor issues.

GBathie:
> Thanks for your contribution and the changes. Now, gaussian and sparse random projection codes look very alike, I am wondering if you could not refactor even further by using zero-sized types and a unique `RandomProjection` generic type, something like:
> 
> ```rust
> struct Gaussian;
> struct Sparse;
> 
> pub struct RandomProjectionValidParams<RandomMethod, R: Rng + Clone> {
>     pub params: RandomProjectionParamsInner,
>     pub rng: Option<R>,
>     pub method: std::marker::PhantomData<RandomMethod>,
> }
> 
> pub struct RandomProjectionParams<RandomMethod, R: Rng + Clone>(
>     pub(crate) RandomProjectionValidParams<RandomMethod, R>,
> );
> 
> pub struct RandomProjection<RandomMethod, F: Float> {
>     projection: Array2<F>,
>     method: std::marker::PhantomData<RandomMethod>,
> }
> 
> pub struct GaussianRandomProjection<F: Float> = RandomProjection<Gaussian, F: Float>;
> pub struct SparseRandomProjection<F: Float> = RandomProjection<Sparse, F: Float>;
> 
> impl<F, Rec, T, R> Fit<Rec, T, ReductionError> for RandomProjectionValidParams<Gausssian, R>
> where
>     F: Float,
>     Rec: Records<Elem = F>,
>     StandardNormal: Distribution<F>,
>     R: Rng + Clone,
> {
>     type Object = RandomProjection<Gaussian, F>;
> 
>     fn fit(&self, dataset: &linfa::DatasetBase<Rec, T>) -> Result<Self::Object, ReductionError> {...}
> }
> 
> impl<F, Rec, T, R> Fit<Rec, T, ReductionError> for RandomProjectionValidParams<Sparse, R>
> where
>     F: Float,
>     Rec: Records<Elem = F>,
>     StandardNormal: Distribution<F>,
>     R: Rng + Clone,
> {
>     type Object = RandomProjection<Sparse, F>;
> 
>     fn fit(&self, dataset: &linfa::DatasetBase<Rec, T>) -> Result<Self::Object, ReductionError> {...}
> }
> 
> ...
> ```
> 
> What do you think?

I think that's a very good suggestion, it will be easier to maintain than the previous approach using a macro to avoid code duplication. [6b9c2a4](https://github.com/rust-ml/linfa/pull/332/commits/6b9c2a4c0d059ae29fa88e66759951b14ce56293) implements a variation of this idea: all the logic has been refactored, and behavior depending on the projection method has been encapsulated in the `ProjectionMethod` trait. It also makes implementing other projection methods significantly easier.

relf (approved):

quietlychris (approved):

PR #263, opened 2022-11-06
Pls bench

oojo12:
This adds PLS benchmarking code. Albeit, I made a mistake and made it off the iai_ols_bench branch so it is dependent on a merge of that branches changes to the make_dataset function. The criterion diagrams are attached as [linfa_pls.zip](https://github.com/rust-ml/linfa/files/9946094/linfa_pls.zip)

### Context
1. Laptop idle except bench runs
2. Laptop mostly full charge but running on battery (not plugged in)
3. Rust Version - rustc 1.65.0 (897e37553 2022-11-02)

### Bench Results cargo bench -p linfa-pls
```
     Running benches\pls.rs (target\release\deps\pls-e5f47fa7aee48589.exe)
Linfa_pls/SVD/1000      time:   [519.14 µs 535.71 µs 553.07 µs]
Found 14 outliers among 100 measurements (14.00%)
  4 (4.00%) high mild
  10 (10.00%) high severe
Linfa_pls/SVD/10000     time:   [5.1048 ms 5.1683 ms 5.2393 ms]
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) high mild
  6 (6.00%) high severe
Benchmarking Linfa_pls/SVD/100000: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.3s, or reduce sample count to 70.
Linfa_pls/SVD/100000    time:   [60.800 ms 62.163 ms 63.621 ms]
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
```

### Bench Results cargo bench -p linfa-pls -F intel-mkl-static
```
     Running benches\pls.rs (target\release\deps\pls-171ffb7df7822c7b.exe)
Linfa_pls/SVD/1000      time:   [524.84 µs 537.65 µs 552.99 µs]
                        change: [+8.4643% +12.173% +15.848%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
Linfa_pls/SVD/10000     time:   [4.9096 ms 4.9755 ms 5.0472 ms]
                        change: [-5.4899% -3.7303% -1.8338%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
[...]

PR #261, opened 2022-11-03
Ica bench

oojo12:
### Additions
Adds benchmarking code for linfa-ica it is a small refactor of the provided example. Will add benchmark stats if accepted.

PR #355, opened 2024-07-09
use rayon to speed up linfa-logistic

droundy:
This PR speeds up a test logistic regression by a factor of two on my laptop, from 2 minutes and 13 seconds to just 1 minute.

### DIFF
--- a/algorithms/linfa-logistic/src/lib.rs
+++ b/algorithms/linfa-logistic/src/lib.rs
@@ -454,9 +454,9 @@ fn log_sum_exp<F: linfa::Float, A: Data<Elem = F>>(
 /// Computes `exp(n - max) / sum(exp(n- max))`, which is a numerically stable version of softmax
 fn softmax_inplace<F: linfa::Float, A: DataMut<Elem = F>>(v: &mut ArrayBase<A, Ix1>) {
     let max = v.iter().copied().reduce(F::max).unwrap();
-    v.mapv_inplace(|n| (n - max).exp());
+    v.par_mapv_inplace(|n| (n - max).exp());
     let sum = v.sum();
-    v.mapv_inplace(|n| n / sum);
+    v.par_mapv_inplace(|n| n / sum);
 }
 
 /// Computes the logistic loss assuming the training labels $y \in {-1, 1}$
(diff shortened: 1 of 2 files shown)


algorithms/linfa-logistic/src/lib.rs
```diff
@@ -495,8 +495,7 @@ fn logistic_grad<F: Float, A: Data<Elem = F>>(
     let yz = x.dot(&params.into_shape((params.len(), 1)).unwrap()) + intercept;
     let len = yz.len();
     let mut yz = yz.into_shape(len).unwrap() * y;
-    yz.mapv_inplace(logistic);
-    yz -= F::one();
+    yz.par_mapv_inplace(|v| logistic(v) - F::one());
```
bartek-siudeja, line 498: Could you put `* y` from below inside par_mapv_inplace? Plus there is `* y` above too.

algorithms/linfa-logistic/src/lib.rs
```diff
@@ -454,9 +454,9 @@ fn log_sum_exp<F: linfa::Float, A: Data<Elem = F>>(
 /// Computes `exp(n - max) / sum(exp(n- max))`, which is a numerically stable version of softmax
 fn softmax_inplace<F: linfa::Float, A: DataMut<Elem = F>>(v: &mut ArrayBase<A, Ix1>) {
     let max = v.iter().copied().reduce(F::max).unwrap();
-    v.mapv_inplace(|n| (n - max).exp());
+    v.par_mapv_inplace(|n| (n - max).exp());
```
bartek-siudeja, line 457: Is there `par_reduce` by any chance, for max and sum?

algorithms/linfa-logistic/src/lib.rs
bartek-siudeja: ah, `y` is also a vector, so this would need to be more like map on zipped vectors, store in one of them

bartek-siudeja: Maybe another comment, the expression we have right now, isn't it `-logistic(-x)`, so we could avoid possibly numerically bad subtraction.

relf (commented):
Hi, thanks. Would you add some benches then attach the results here in the PR? See #263 or #261 as examples.

droundy:
> Hi, thanks. Would you add some benches then attach the results here in the PR? See #263 or #261 as examples.

Alas, I implemented a benchmark, but it seems that the limiting step (if you have many more samples than "features") is a matrix multiply that I didn't parallelize.  And moreover I don't see a good way to make `ndarray` do the multiply in parallel.  :(

I'll leave this open for now, in case someone wants to use it as a starting point for doing this properly, but feel free to close when you wish to.

PR #368, opened 2025-01-09
CI maintenance

relf:
- [x] Make clippy happy
- [x] Bump MSRV to 1.71.1
- [x] Pin `sprs` to 1.11.1 version (1.11.2 pulls `ndarray 0.16.1` which makes `linfa-kernel` compilation fail)
- [x] Fix code coverage (?)

### DIFF
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         toolchain:
-          - 1.71.0
+          - 1.71.1
           - stable
           - nightly
         os:
--- a/.github/workflows/checking.yml
+++ b/.github/workflows/checking.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         toolchain:
-          - 1.70.0
+          - 1.71.0
           - stable
           - nightly
         os:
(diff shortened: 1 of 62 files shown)


quietlychris (approved):

quietlychris:
Looks good to me! The one comment I have is that switching from `Vec` to `&[]` can sometimes make things a little less flexible/finicky, but I guess it make more sense to be consistent. Feel free to merge whenever works!

quietlychris:
Also, it's interesting that `sprs = 1.11.2` is breaks things. It's dependency is labeled at `ndarray = ">=0.15.0, <0.17"`, so I understand why `ndarray 0.16.1` is an option, but in theory it should just accept `linfa's` existing `ndarray` dependency number for `0.15`. Definitely having a patch bump shouldn't result in a break per SemVer guidelines.

relf:
> Was that change motivated by clippy lints?

Yes ([useless_vec](https://rust-lang.github.io/rust-clippy/master/index.html#useless_vec)), the change was without sensible side-effect, so...

> Also, it's interesting that sprs = 1.11.2 is breaks things [...]  Definitely having a patch bump shouldn't result in a break per SemVer guidelines.

I do not understand why `ndarray 0.15` is not selected by `sprs` dependency, either.

PR #370, opened 2025-01-23
[linfa-svm] Fix SVR nu parameter passing and rework SVR parameterization API

### LINKED ISSUE #317: Is the `nu` parameter used incorrectly, for svm regression?
Was just browsing through the source, and this seemed weird, I could be totally off-base so feel free to just close, but in `fit_nu`, `nu` is [the last param](https://github.com/rust-ml/linfa/blob/master/algorithms/linfa-svm/src/regression.rs#L81) , but when it's called, `nu` is the second-to-last param, and c is the last param: https://github.com/rust-ml/linfa/blob/master/algorithms/linfa-svm/src/regression.rs#L144-L145

relf:
This PR fixes and clarifies the SVR parameterization with regards to [the original libsvm code](https://github.com/cjlin1/libsvm/blob/35e55962f7f03ce425bada0e6b9db79193e947f8/svm.cpp#L1610)  
before:
* `nu_eps(nu, eps)` was wrongly wired: `eps` was used for `nu` and `nu` for `c`
* `c_eps(c, eps)` is confusing:  eps refers is epsilon of the inner loss function while there is also an epsilon used for the solver
after: 
* `nu_svr(nu, Option<c>)` replaces deprecated `nu_eps()`, optional parameter default is 1. as in libsvm
* `c_svr(c, Option<loss_eps>)` replaces deprecated `c_eps()` second optional parameter default is 0.1 as in libsvm
* the `epsilon` of the solver is set through `SolverParams::epsilon()` as before

Fixes #317

### DIFF
--- a/.github/workflows/codequality.yml
+++ b/.github/workflows/codequality.yml
@@ -30,9 +30,10 @@ jobs:
         run: cargo clippy --all-targets -- -D warnings
 
   coverage:
+    needs: codequality
     name: coverage
     runs-on: ubuntu-latest
-    if: github.event.pull_request.draft == false
+    if: github.event.pull_request.draft == false && (github.event_name == 'pull_request' || github.ref == 'refs/heads/master')
 
     steps:
       - name: Checkout sources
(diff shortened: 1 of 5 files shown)


relf:
Not sure to understand why codecov/patch is not happy. The missing lines should be covered by tests in `regression.rs`, so... @quietlychris, @xd009642, any idea?

PR #371, opened 2025-01-29
Upgrade to ndarray 0.16

relf:
- [x] `ndarray 0.16`
- [x] `linfa-llinalg 0.2`
- [x] `narray-linalg 0.17`
- [x] `ndarray-stats 0.6`
- [x] `ndarray-rand 0.15`
- [x] `ndarray-csv 0.5`
- [x] `sprs 0.11`
- [x] `approx 0.5`
- [ ] `argmin > 0.10.0`
- [ ] `argmin-math > 0.4.0`
- [ ] `criterion 0.5`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ approx = "0.5"
 
 ndarray = { version = "0.16", features = ["approx"] }
 ndarray-linalg = { version = "0.17", optional = true }
-sprs = { version = "=0.11.1", default-features = false }
+sprs = { version = "0.11", default-features = false }
 
 thiserror = "1.0"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ serde = ["serde_crate", "ndarray/serde"]
 [dependencies]
 num-traits = "0.2"
 rand = { version = "0.8", features = ["small_rng"] }
-approx = "0.4"
+approx = "0.5"
 
 ndarray = { version = "0.16", features = ["approx"] }
 ndarray-linalg = { version = "0.17", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ num-traits = "0.2"
 rand = { version = "0.8", features = ["small_rng"] }
 approx = "0.4"
 
-ndarray = { version = "0.15", features = ["approx"] }
+ndarray = { version = "0.16", features = ["approx"] }
 ndarray-linalg = { version = "0.16", optional = true }
 sprs = { version = "=0.11.1", default-features = false }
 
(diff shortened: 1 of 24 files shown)


amfaber:
Is anything blocking this other than the release of a new version of argmin that supports ndarray 0.16?

relf:
> Is anything blocking this other than the release of a new version of argmin that supports ndarray 0.16?

No. The PR is currently tested against argmin master. AFAIK, argmin's maintainer will release a new version in a few weeks.

amfaber:
Fantastic! Thanks for the quick response :)

shigedangao:
Any update on that PR ?

relf:
> Any update on that PR ?

Nope... Still waiting for `argmin` new release.

xum001:
So, when will it be released? Today is August 3.

xum001 (approved):

jonaspleyer:
Since this is a foundational library, it might make sense to use a range of versions such as `version= ">=0.15, <=0.16"` in order to provide better compatibility. For me, `linfa` is a secondary dependency via `egobox` and I would like to avoid any future clashes with other crates that use slightly different versions.

relf:
@jonaspleyer, Beside the fact the code should then be compatible/tested with various versions of `ndarray`, I am not sure that such spec solves the problem. As stated [here](https://github.com/PyO3/rust-numpy?tab=readme-ov-file#dependency-on-ndarray) for `PyO3/rust-numpy`, I am afraid you will need to pin the version of `ndarray` anyway.

jonaspleyer:
> @jonaspleyer, Beside the fact the code should then be compatible/tested with various versions of `ndarray`, I am not sure that such spec solves the problem. As stated [here](https://github.com/PyO3/rust-numpy?tab=readme-ov-file#dependency-on-ndarray) for `PyO3/rust-numpy`, I am afraid you will need to pin the version of `ndarray` anyway.

Well ... this problem should be solved by a `1.0` release of `ndarray`. This would mean that we can rely on new versions being compatible. For now, we are stuck in an uncomfortable intermediate state where we can not be sure that `0.17` will be compatible with `0.16` for instance. However, a `1.0` release is not close as far as I can tell.

Both ways which we have discussed have their advantages and challenges. I can understand that you are hesitant to simply allow arbitrary ranges of versions. For me, as a user who relies on many packages that either use `ndarray` or re-export it themselves, the flexibility would be a great relief. I do not mind pinning the version number but I would like to be able to combine multiple packages with as little ease as possible.

stefan-k:
> > * argmin
> 
> Seems to still blocked by the fact that there are no new argmin release...

A new argmin release is out: https://github.com/argmin-rs/argmin/releases/tag/argmin-v0.11.0

Apologies for blocking this, I unfortunately wasn't able to release earlier.

PR #376, opened 2025-02-06
add max_features and tokenizer to CountVectorizer

marco-cloudflare:
add max_features and tokenizer to CountVectorizer (similar to what's available at sklearn). Note that tokenizer and regex as competing parameters, in case sklearn, it disables regex if you pass a tokenizer and gives you a warning, so here we could think of a single parameter that would encompass both.

another caveat is the serialization of the tokenizer function pointer, the workaround I made was not skip it, allow it to be reset after deserialization and keep a guard that will error if you try to use transform after deserialization without resetting a tokenizer

### DIFF
--- a/algorithms/linfa-preprocessing/Cargo.toml
+++ b/algorithms/linfa-preprocessing/Cargo.toml
@@ -32,6 +32,7 @@ encoding = "0.2"
 sprs = { version = "=0.11.1", default-features = false }
 
 serde_regex = { version = "1.1", optional = true }
+itertools = "0.14.0"
 
 [dependencies.serde_crate]
 package = "serde"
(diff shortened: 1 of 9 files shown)


relf (changes requested):
Thanks for your contribution. As you mentioned, it would be great to have only one parameter with an appropriate enum type to unify regexp and tokenizer function. 
Could you also add some tests and maybe an example of the tokenizer function you introduce?

marco-cloudflare:
implemented the enum interface, documentation and testing

relf (changes requested):
Changes look good to me except for clippy linting. Would you mind rebasing your branch? Thanks!

marco-cloudflare:
done

relf (approved):

relf:
Thanks!

Issue #375, opened 2025-02-06
Shuffling erases feature names


Plutone11011:
Hello, while playing around with the iris dataset, I noticed that if I shuffle it, it loses the feature names.

```
let mut rng = rand::thread_rng();

train = linfa_datasets::iris().shuffle(&mut rng);

println!("Feature names {:?}", train.feature_names());
```

prints 
Feature names ["feature-0", "feature-1", "feature-2", "feature-3"]

so I have to use `with_feature_names` again with the iris feature names which is not ideal, in my opinion. Is there a way to maintain feature names during shuffling, without having to explicitly setting them again?

relf:
Thanks, good catch! Indeed [shuffle()](https://github.com/rust-ml/linfa/blob/9744625938e3d68194c871c37c7868c1f174bd9d/src/dataset/impl_dataset.rs#L580) creates a new `Dataset` but forgets to set features names.

You could also use `with_target_names` with the latest version.

relf:
Fixed by #377

PR #377, opened 2025-02-09
Maintain original feature and target names after shuffling

Plutone11011:
Fixing Issue [#375](https://github.com/rust-ml/linfa/issues/375)

### DIFF
--- a/src/dataset/impl_dataset.rs
+++ b/src/dataset/impl_dataset.rs
@@ -577,19 +577,6 @@ where
     /// ### Returns
     ///
     /// A new shuffled version of the current Dataset
-    /// # Example
-    /// ```
-    /// use rand::thread_rng;
-    /// use ndarray::s;
-    /// let dataset = linfa_datasets::iris();
-    /// println!("First 5 rows {:?}", dataset.records.slice(s![0..5,..]));
-    /// println!("Feature names {:?}", dataset.feature_names());
-    /// println!("Target names {:?}", dataset.target_names());
-    /// let mut rng = thread_rng();
-    /// let shuffled = dataset.shuffle(&mut rng);
-    /// println!("First 5 rows after shuffling {:?}", shuffled.records.slice(s![0..5,..]));
-    /// println!("Feature names after shuffling {:?}", shuffled.feature_names());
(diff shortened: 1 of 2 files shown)


src/dataset/impl_dataset.rs
```diff
@@ -577,6 +577,20 @@ where
     /// ### Returns
     ///
     /// A new shuffled version of the current Dataset
+    /// # Example
+    /// ```
+    /// use rand::thread_rng;
```
relf, line 582: In an example, it would be better to have assertions with explicit values rather than `println!()` which requires the user to run the example. 

As the purpose of the function is rather straightforward and as checking names in an example of `shuffle()` seems a bit contrived, here I would simply go with a test in `mod.rs` instead of an example. What do you think?

relf (commented):
Thanks for the fix.

src/dataset/impl_dataset.rs
Plutone11011: I agree, it does seem contrived to check for names in shuffle, but since this is the purpose of this PR, do you think I should still add an assertion for feature and target names in the test in `mod.rs`?

relf: Yes, my point was to remove the example and add a test of `shuffle()` in `mod.rs` testing also you've fixed the bug, something like:
```
    #[test]
    fn test_dataset_shuffle() {
        let mut rng = SmallRng::seed_from_u64(42);
        let f_names = vec!["1", "2", "3"];
        let t_names = vec!["one"];
        let dataset = Dataset::new(
            array![[1., 2., 3.], [4., 5., 6.], [7., 8., 9.]],
            array![0., 1., 3.],
        )
        .with_feature_names(f_names.clone())
        .with_target_names(t_names.clone());

        let shuffled = dataset.shuffle(&mut rng);

        assert_abs_diff_ne!(dataset.records(), shuffled.records());
        assert_abs_diff_ne!(dataset.targets(), shuffled.targets());
        assert_eq!(f_names, shuffled.feature_names());
        assert_eq!(t_names, shuffled.target_names());
    }
```

relf:
Thanks!

PR #381, opened 2025-03-18
Add `predict_proba()` to Gaussian mixture model

### LINKED ISSUE #372: Does GMM have the option to get the cluster probability?
Hey!

After doing the fit:

```
// We fit the model from the dataset setting some options
    let gmm = GaussianMixtureModel::params(n_clusters)
                .n_runs(10)
                .tolerance(1e-4)
                .with_rng(rng)
                .fit(&dataset).expect("GMM fitting");

    // Then we can get dataset membership information, targets contain **cluster indexes**
    // corresponding to the cluster infos in the list of GMM means and covariances
    let blobs_dataset = gmm.predict(dataset);
```

Can I get the probability of it belonging to one of the clusters? It will assign it to the cluster with the highest probability (I assume) however, I want to only assign if the probability is higher than a specific threshold.

I'm not familiar with linfa at all, so perhaps there is a standard way of doing this.

Thanks for building all this!

techyorzl:
This PR is intended to add the `predict_proba` method to Gaussian Mixture and includes tests.

Closes #372

### DIFF
--- a/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
+++ b/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
@@ -211,6 +211,11 @@ impl<F: Float> GaussianMixtureModel<F> {
         self.means()
     }
 
+    pub fn predict_proba<D: Data<Elem = F>>(&self, observations: &ArrayBase<D, Ix2>) -> Array2<F> {
+        let(_, log_resp) = self.estimate_log_prob_resp(observations);
+        log_resp.mapv(F::exp)
+    }
+
     #[allow(clippy::type_complexity)]
     fn estimate_gaussian_parameters<D: Data<Elem = F>>(
         observations: &ArrayBase<D, Ix2>,
@@ -483,6 +488,7 @@ impl<F: Float, D: Data<Elem = F>> PredictInplace<ArrayBase<D, Ix2>, Array1<usize
 #[cfg(test)]
 mod tests {
     use super::*;
+    use rand_xoshiro::Xoshiro256Plus;
     use approx::{abs_diff_eq, assert_abs_diff_eq};
     use linfa_datasets::generate;
     use linfa_linalg::LinalgError;
@@ -493,6 +499,7 @@ mod tests {
     use ndarray_rand::rand_distr::Normal;
     use ndarray_rand::rand_distr::{Distribution, StandardNormal};
     use ndarray_rand::RandomExt;
+    use ndarray::Array;
 
     #[test]
     fn autotraits() {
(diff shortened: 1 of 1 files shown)


algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -746,4 +753,33 @@ mod tests {
...
+            .expect("Failed to fit GMM");
+    
+        // getting probabilities
+        let proba = gmm.predict_proba(dataset.records());
+    
+        // checking output shape is correct or not
+        assert_eq!(proba.dim(), (total_samples, n_clusters));
+    
+        // checking for each sample, the sum of probabilities is 1
+        for sample in proba.outer_iter() {
```
relf, line 780: You can use [sum_axis(Axis(1))](https://docs.rs/ndarray/0.15.6/ndarray/struct.ArrayBase.html#method.sum_axis) and check against `Array::ones(n_samples)`

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -211,6 +211,11 @@ impl<F: Float> GaussianMixtureModel<F> {
         self.means()
     }
 
+    pub fn predict_proba<D: Data<Elem = F>>(&self, observations: &ArrayBase<D, Ix2>) -> Array2<F> {
```
relf, line 214: Add a doc string, something like : "Returns the responsabilities as a (n_obs, n_clusters) array where each row gives the probabilities for the corresponding ith observation to belong to the jth cluster."

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -746,4 +753,33 @@ mod tests {
...
+
+    #[test]
+    fn test_predict_proba() {
+        let mut rng = Xoshiro256Plus::seed_from_u64(42);
+        let centroids = array![[0.0, 1.0], [-10.0, 20.0], [-1.0, 10.0]];
+        let n_samples_per_cluster = 1000;
+        let dataset = DatasetBase::from(generate::blobs(n_samples_per_cluster, &centroids, &mut rng));
+        let n_clusters = centroids.len_of(Axis(0));
+        // total samples = (n_samples_per_cluster * n_clusters)
+        let total_samples = n_samples_per_cluster * n_clusters;
```
relf, line 765: Maybe `n_samples` instead of `total_samples` to be consistent

relf (changes requested):
Thanks for your contribution. See below some minor fixes. You have to format (`cargo fmt`) and lint (`cargo clippy`) the file.

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
techyorzl: Ok Thank You so much for the help, I was wondering where I was wrong. I will do these updates and update the code.

relf (approved):
LGTM, thanks!

PR #387, opened 2025-05-03
linfa-preprocessing::countgrams: RefCell is replaced with OnceLock

kakserpom:
Simple fix to allow for concurrent access.

### DIFF
--- a/algorithms/linfa-preprocessing/src/countgrams/hyperparams.rs
+++ b/algorithms/linfa-preprocessing/src/countgrams/hyperparams.rs
@@ -1,8 +1,8 @@
 use crate::PreprocessingError;
 use linfa::ParamGuard;
 use regex::Regex;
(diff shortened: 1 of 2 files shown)


relf:
`OnceLock` is not serde. What about cloning to use multi-threading?

relf:
With `OnceLock` we cannot serialize/deserialize `CountVectorizerValidParams` which is a no go (cf. CI tests errors). 

I meant you can just clone `CountVectorizerParams` and use an independent copy in each thread (if it was your initial problem with the multi-threading).

kakserpom:
@relf Well, I guess we could solve the issue by implementing serialize and deserialize manually, right?

relf:
Maybe... But there is no problem in the first place if you just clone `CountVectorizerParams` so no need of a special serde code complication.

PR #410, opened 2025-10-17
Implement Random Forest as a sub case to EnsembleLearner 

ladezai:
Premise: a Random Forest is just Bagging with the additional constraint that each weak predictor knows only a subset of the feature of the dataset. In this PR, we updated the code of `EnsembleLearner` to account for that key constraint.

Implementation details: 
* Adds to `src/dataset/impl_dataset.rs` 3 new methods: `bootstrap_with_indices`, `bootstrap_samples_with_indices`, `bootstrap_features_with_indices`. Indices knowledge is fundamental to know which feature should each weak classifier.
* Adds the parameter `feature_proportion` to `EnsembleLearnerValidParams`.
* Adds `model_features` to `EnsembleLearner` to keep track of which feature should each weak learner use.
* Updates the `examples/bagging_iris.rs` -> `examples/ensemble.rs`. 

Hope this implementation meets the library standards. If any further modification is required, just ping me or feel free to modify it :)

### DIFF
--- a/algorithms/linfa-ensemble/src/lib.rs
+++ b/algorithms/linfa-ensemble/src/lib.rs
@@ -9,8 +9,8 @@
 //! several decision trees (see `linfa-trees`) trained on different samples subset of the training dataset.
 //!
 //! ## Random Forest
-//! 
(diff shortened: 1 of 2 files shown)


relf:
Thanks!

PR #422, opened 2025-12-06
fix: resolve ndarray version mismatch

rathideep22:
Resolves compilation errors due to ndarray version conflict (0.16 vs 0.17) by pinning sprs and ndarray-csv.


relf (approved):
Thanks! Indeed better than CI patch.

Issue #411, opened 2025-10-19
Add boosting support to linfa-ensemble


yfnaji:
I was looking into the `linfa-ensemble` crate and noticed that it currently does not include a boosting mechanism. The current ensemble implementation appears to be quite generalised, so we could either extend `EnsembleLearner` to support boosting methods or (preferably) create a new struct dedicated to boosting, while keeping `EnsembleLearner` as the dedicated bagging struct (and rename `EnsembleLearner` -> `Bagging`).

I would like your thoughts on this - I’d be happy to work on this and write an initial implementation.

PR #430, opened 2026-02-27
Add generic `ResidualChain` composing method

Feiyang472:
Hi! Wanted to float the idea of a small crate for chaining models together additively, fitting each model with the residuals of the previous one, like below

```
LinearRegression::default()
    .stack_with(some_svm)
    .stack_with(LinearRegression::default())
    .fit(&dataset)?
```

feels like a natural ergonomic addition which could sit as a crate in linfa.

Happy to hear if this belongs somewhere else (folded into an existing crate, or not at all).


relf:
Hi! It looks like a nice addition! I think you should move it in the `linfa` crate under the existing `composing` module which gathers composite models.

Feiyang472:
Hi @relf, absolutely composing is the better place for this to be! I have moved it to composing/residual-sequence.

src/composing/residual_chain.rs
```diff
@@ -0,0 +1,337 @@
...
+/// Wraps the error from whichever of the two model fits failed, keeping them
+/// distinguishable without requiring both models to share the same error type.
+#[derive(Debug, thiserror::Error)]
+pub enum ResidualSequenceError<E1, E2> {
+    #[error("first model: {0}")]
+    First(E1),
+    #[error("second model: {0}")]
+    Second(E2),
+    // Satisfies the `Fit` trait's `E: From<linfa::error::Error>` bound.
+    #[error(transparent)]
```
Feiyang472, line 142: Wondering if we could loosen the trait bound in traits.rs for Fit

from
```
pub trait Fit<R: Records, T, E: std::error::Error + From<crate::error::Error>>
```

to 
```
pub trait Fit<R: Records, T, E: std::error::Error>
```

although errors raised by the composite model could be naturally converted into a linfa error (by just discarding information on which part of the fitting went wrong), given a linfa error, we cannot know whether the first model fit or the residual model fit went wrong. But in order to satisfy the trait bound I came up with this third enum member which never gets used explicitly

src/composing/residual_sequence.rs
```diff
@@ -0,0 +1,340 @@
...
+    Linfa(#[from] crate::error::Error),
+}
+
+/// Fits two models sequentially on the residuals of the first.
+///
+/// `first` is fit on the original dataset. `second` is fit on the residuals
+/// `Y - first.predict(X)`. See the [module docs](self) for details.
+#[derive(Debug, Clone)]
+pub struct ResidualSequence<F1, F2> {
+    pub first: F1,
```
relf: Prefer accessors instead of public fields.

src/composing/residual_sequence.rs
```diff
@@ -0,0 +1,340 @@
+//! Residual sequence model composition for the linfa ML framework.
+//!
+//! This crate provides [`ResidualSequence`], which fits models sequentially on
```
relf: Is "residual sequence" a well known terminology? Any reference paper?

src/composing/residual_sequence.rs
```diff
@@ -0,0 +1,340 @@
...
+
+impl<'a, R1, R2, D> Predict<&'a Arr2<D>, Array1<D::Elem>> for FittedResidualSequence<R1, R2>
+where
+    D: Data,
+    D::Elem: Copy + Add<Output = D::Elem>,
+    Arr2<D>: Records,
+    for<'b> R1: Predict<&'b Arr2<D>, Array1<D::Elem>>,
+    for<'b> R2: Predict<&'b Arr2<D>, Array1<D::Elem>>,
+{
+    fn predict(&self, x: &'a Arr2<D>) -> Array1<D::Elem> {
```
relf: You have to implement `PredictInplace` instead of `Predict` as stated in the [guidelines](https://github.com/rust-ml/linfa/blob/b1f9ddb73d2e5537e664b4f0c632f147d3060606/CONTRIBUTE.md)

src/composing/residual_sequence.rs
```diff
@@ -0,0 +1,340 @@
...
+pub struct FittedResidualSequence<R1, R2> {
+    pub first: R1,
+    pub second: R2,
+}
+
+impl<F1, F2, D, T, E1, E2> Fit<Arr2<D>, T, ResidualSequenceError<E1, E2>>
+    for ResidualSequence<F1, F2>
+where
+    D: Data + RawDataClone,
+    D::Elem: Copy + Sub<Output = D::Elem>,
```
relf: Why not use `F: Float` ?

```rust
...
where
    F: Float,
    D: Data<Elem = F> + RawDataClone,
...
```

src/composing/residual_sequence.rs
```diff
@@ -0,0 +1,340 @@
...
+            .map_err(ResidualSequenceError::First)?;
+
+        let y_pred = first.predict(dataset.records());
+        let residuals = dataset
+            .targets()
+            .as_targets()
+            .iter()
+            .zip(y_pred.iter())
+            .map(|(y, p)| *y - *p)
+            .collect::<Array1<D::Elem>>();
```
relf: You can write
```rust
let targets = dataset.targets().as_targets();
let residuals = &targets - &y_pred;
```

relf (changes requested):
Ok, I think you should take a look at the [contribution guidelines](https://github.com/rust-ml/linfa/blob/b1f9ddb73d2e5537e664b4f0c632f147d3060606/CONTRIBUTE.md). Specially the sections about the implementation of `PredictInplace` trait and the optional `serde` support. As there is no hyperparameters (at the moment?), I guess you can skip the `ParamGuard` pattern.

src/composing/residual_chain.rs
relf: Well there is no need per se here but the idea of the bound `From<crate::error::Error>>` in the first place is that sub-crates should handle errors caused by the base crate `linfa`.

src/composing/residual_sequence.rs
Feiyang472: Good idea! have done as the guidelines recommended

src/composing/residual_sequence.rs
Feiyang472: yes, fixed this

src/composing/residual_sequence.rs
Feiyang472: that's what I have called it so far, after a bit research I think the statisticians would refer to this as forward stagewise additive model https://nyu-ds1003.github.io/mlcourse/2021/lectures/lec11/11a.forward-stagewise-annotated.pdf (with vm=1 for all m). I could also allow configuring v_m parameter here. that would allow me to meaningfully implement ParamGuard as well

src/composing/residual_sequence.rs
Feiyang472: have switched to accessors instead. 

as an alternative to first/second as field names, would you prefer

- base / corrector (stats)
- left / right (because this in general feels like a tree)

?

src/composing/residual_sequence.rs
Feiyang472: Thanks! I have switched to PredictInplace. I saw the guidelines mention new models should implement PredictInplace. Do you expect any models to implement Predict but not PredictInplace? If not I can change the bound on the component models to be `PredictInplace` as well.

src/composing/residual_sequence.rs
```diff
@@ -0,0 +1,349 @@
...
+    #[test]
+    fn deep_chain_accessors() {
+        let model = MeanParams
+            .stack_with(MeanParams)
+            .stack_with(MeanParams)
+            .stack_with(MeanParams);
+        let dataset = DatasetBase::new(array![[0.0_f64], [1.0]], array![1.0, 3.0]);
+        let fitted = model.fit(&dataset).unwrap();
+
+        assert_eq!(fitted.first().first().first().0, 2.0); // params trained on original targets
```
Feiyang472: this feels very verbose but I am not sure if there is a better syntax given models are in generous heterogeneous

src/composing/residual_sequence.rs
relf: Yeah, I was thinking about that, first and second are not very informative. I would prefer base and corrector as first and second does not play the same role.

src/composing/residual_sequence.rs
relf: There are a blanket implementations for `Predict`, the user has just to implement `PredictInplace`.

src/composing/residual_sequence.rs
relf: You cannot do better with just `ResidualSequence` which is "just" a chain link.

src/composing/residual_sequence.rs
relf: Thanks for the reference. `Residual Sequence` now is like one iteration of FSAM. What do you think of `ResidualChain`? It feels more rusty and is even shorter.

Yes, the configuration with `vm` would be nice along with the reference link (maybe try to find a paper or book link)

Feiyang472: cool, will try get to it later today. thanks @relf

Feiyang472: I have added a shrinkage parameter

src/composing/residual_sequence.rs
Feiyang472: renamed to base/corrector. I think it makes more sense

src/composing/residual_chain.rs
```diff
@@ -0,0 +1,558 @@
...
+    /// by calling `.stack_with(...)` on the returned [`ResidualChainParams`].
+    fn stack_with<C, F: Float>(self, corrector: Shrunk<C, F>) -> ResidualChainParams<Self, C, F>;
+    /// Wrap `self` in a [`Shrunk`] with learning rate `shrinkage` ∈ (0, 1],
+    /// making it ready to pass as the `corrector` argument to [`Stagewise::stack_with`].
+    ///
+    /// The bound `Self: Fit<Array2<F>, Array1<F>, E>` ensures at compile time
+    /// that the model's element type matches the shrinkage type `F`.
+    fn shrink_by<F: Float, E>(self, shrinkage: F) -> Shrunk<Self, F>
+    where
+        Self: Fit<Array2<F>, Array1<F>, E>,
```
Feiyang472, line 236: I think this means that shrink_by only works when fitting on Owned arrays, but this is probably fine because the residual data is the result of a subtraction which will always be owned? I am not entirely sure, happy to hear thoughts

src/composing/residual_chain.rs
```diff
@@ -0,0 +1,558 @@
...
+        E: std::error::Error + From<crate::error::Error>;
+}
```
Feiyang472, line 238: could also add another method which just shrinks by `F::one()` (no shrinking). not sure how to name it

src/composing/mod.rs
relf: Should be renamed `residual_chain`

src/composing/residual_chain.rs
relf: You require `<Model>Params` to fit on owned arrays which is ok for me

src/composing/residual_chain.rs
relf: Well, the `.shrink_by(1.0)` is noisy... With a conversion we could have:
```rust
let model = MeanParams.stack_with(MeanParams.into()));
```
otherwise you can add another method:
```rust
let model = MeanParams.stack_with_full(MeanParams);
```
It is better if the most common use case has the shorter name. 
We could also have just `stack(MeanParams)` and `stack_shrunk(MeanParams, 0.5)`.

Feiyang472: Hi @relf thanks for reviewing. I propose the two methods:


```
chain(corrector: C)
chain_shrunk(corrector: Shrunk<C, F>)
```

I went with chain over stack since "stack" is so strongly associated with the fundamental data structure that it could be confusing in a library context. `chain` also echoes the `ResidualChain` struct name you suggested.

I kept chain_shrunk accepting a pre-wrapped Shrunk<C, F> rather than an inline shrinkage value (as in chain_shrunk(corrector, 0.5)) so that a Shrunk corrector can be constructed once and reused across multiple pipelines.

relf: Excellent!

relf (approved):
Thanks for this contribution!

relf:
This addition may help address #411.

Feiyang472:
Thanks for reviewing!

PR #432, opened 2026-03-17
Fix label ordering in binary logistic regression

### LINKED ISSUE #393: linfa-logistic: labels are not Ord ordered, 'neg' is the one first found in data during training
Hey,

I'm using linfa-logistic for pet project. I needed some predictions and run into weird problem, where model.predict_probabilities(&input) would give reversed order.

So I was looking into it, also reading code. The documentation says that neg and pos classes are defined depending on Ord of class values, BUT actually they are defined on Ord of count(*) of particular class in data set (ie., larger class is always pos).

This was confusing as heck for me, but I'm not a ML expert (so it might be usual way). Anyhow, I used this workaround to always get probabilities for specific class (I'm using 1 & -1 as targets):

`

            let prob = model.predict_probabilities(&features);
            let labels = model.labels();
            if labels.neg.class == 1 {
                return Some(prob.into_iter().map(|p| 1.0 - p).collect_vec());
            }

            Some(prob.into_iter().collect_vec())
`

Is this a bug or intended? At least this should be documented and (ideally) allow lib users to fix-classes (regardless of input dataset sizes) so that predict_probabilities is stable.

espenloov:
Fixes #393. It now assigns positive/negative labels based on Ord ordering of the label values rather than encounter order in the training data.

### DIFF
--- a/algorithms/linfa-logistic/src/lib.rs
+++ b/algorithms/linfa-logistic/src/lib.rs
@@ -301,32 +301,33 @@ where
 {
     let y = y.as_single_targets();
 
-    // counts the instances of two distinct class labels
     let mut binary_classes = [None, None];
-    // find binary classes of our target dataset
     for class in y {
         binary_classes = match binary_classes {
-            // count the first class label
             [None, None] => [Some((class, 1)), None],
-            // if the class has already been counted, increment the count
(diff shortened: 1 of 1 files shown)


relf (approved):
Thanks for the fix!

PR #437, opened 2026-03-29
feat: add symmetric mean absolute percentage error (sMAPE)

### LINKED ISSUE #436: Feature: Add Symmetric Mean Absolute Percentage Error (sMAPE) to Regression Metrics
Currently, metrics_regression.rs provides Mean Absolute Percentage Error (MAPE), but lacks its symmetric counterpart, sMAPE.

sMAPE is a useful regression metric, particularly in time-series forecasting, because it addresses the asymmetry of standard MAPE (which penalizes over-forecasting more heavily than under-forecasting) and handles values near zero more gracefully by summing the absolute values of both the actual and predicted targets in the denominator.

It would be beneficial to add sMAPE to the Regression Metrics.

oglego:
Fixes #436 

This PR adds the Symmetric Mean Absolute Percentage Error (sMAPE) metric.

sMAPE is a regression metric that addresses the asymmetry found in standard MAPE by normalizing the absolute error by the average of the actual and predicted values. This implementation is numerically stabilized with a small epsilon to handle cases where both the ground truth and prediction are zero.

Verified sMAPE testing with 

```rust
cargo test metrics_regression
```

I tried to keep the formatting and style consistent with MAPE but if there is anything that needs to be changed just let me know!

### DIFF
--- a/src/metrics_regression.rs
+++ b/src/metrics_regression.rs
@@ -94,7 +94,7 @@ pub trait SingleTargetRegression<F: Float, T: AsSingleTargets<Elem = F>>:
             .fold(F::cast(0.0), |acc, &yi, &yhi| {
                 let num = (yhi - yi).abs();
                 let den = yi.abs() + yhi.abs();
-                if den <= F::cast(f64::EPSILON) {
+                if den <= F::epsilon() {
                     acc
                 } else {
                     acc + (num / den)
--- a/src/metrics_regression.rs
+++ b/src/metrics_regression.rs
@@ -8,7 +8,7 @@ use crate::{
     Float,
 };
 use ndarray::prelude::*;
-use ndarray::Data;
+use ndarray::{Data, Zip};
 use std::ops::{Div, Sub};
 
 /// Regression metrices trait for single targets.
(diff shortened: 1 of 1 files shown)


src/metrics_regression.rs
```diff
@@ -80,6 +80,29 @@ pub trait SingleTargetRegression<F: Float, T: AsSingleTargets<Elem = F>>:
...
+        let y_hat = compare_to.as_single_targets();
+        if y.is_empty() {
+            return Err(Error::NotEnoughSamples);
+        }
+        let sum: F = Zip::from(&y)
+            .and(&y_hat)
+            .fold(F::cast(0.0), |acc, &yi, &yhi| {
+                let num = (yhi - yi).abs();
+                let den = yi.abs() + yhi.abs();
+                if den <= F::cast(f64::EPSILON) {
```
relf:
```suggestion
                if den <= F::epsilon() {
```

relf (changes requested):
Thanks for this contribution. After the minor fix below, it will be ok for me.

oglego:
Thank you for reviewing this I really appreciate it!  I've made the recommended change, thanks again!

PR #438, opened 2026-04-17
Add browser WASM feature support

LucaCappelletti94:
I want to use this crate as part of dioxus web apps in my lab, so I added the few things needed to make it compile in WASM settings. I have also added a CI job to verify that indeed it does compile in WASM.


relf (commented):
Thanks for your contribution. Could you rebase on current master where lint errors have been fixed now?

relf (approved):

PR #440, opened 2026-04-21
Add Wasserstein distance (aka Earth Mover's distance) to linfa-nn

mrsanor:
Add Earth Mover's Distance, aka. Wasserstein to linfa-nn

### DIFF
--- a/algorithms/linfa-nn/src/distance.rs
+++ b/algorithms/linfa-nn/src/distance.rs
@@ -118,6 +118,27 @@ impl<F: Float> Distance<F> for LpDist<F> {
     }
 }
 
+/// Wasserstein or [Earth Mover's](https://en.wikipedia.org/wiki/Earth_mover%27s_distance) distance
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(crate = "serde_crate")
+)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EarthMoverDist;
+impl<F: Float> Distance<F> for EarthMoverDist {
+    #[inline]
+    fn distance<D: Dimension>(&self, a: ArrayView<F, D>, b: ArrayView<F, D>) -> F {
+        let mut cumulative_diff = F::zero();
+        let mut emd = F::zero();
+        Zip::from(&a).and(&b).for_each(|&a, &b| {
+            cumulative_diff += a - b;
+            emd += cumulative_diff.abs()
+        });
+        emd
+    }
+}
+
 /// Computes a similarity matrix with gaussian kernel and scaling parameter `eps`
 ///
 /// The generated matrix is a upper triangular matrix with dimension NxN (number of observations) and contains the similarity between all permutations of observations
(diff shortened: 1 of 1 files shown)


algorithms/linfa-nn/src/distance.rs
```diff
@@ -204,4 +226,45 @@ mod test {
...
+        let ab = dist.distance(a.view(), b.view());
+        assert_abs_diff_eq!(ab, 0.8, epsilon = 1e-5);
+
+        let a = arr1(&[0.3, 0.2, 0.15, 0.10, 0.25]);
+        let b = arr1(&[0.1, 0.2, 0.05, 0.20, 0.45]);
+        let ab = dist.distance(a.view(), b.view());
+        assert_abs_diff_eq!(ab, 0.9, epsilon = 1e-5);
+
+        let a = arr1(&[0.35, 0.15, 0.15, 0.10, 0.25]);
+        let b = arr1(&[0.1, 0.20, 0.05, 0.20, 0.45]);
```
relf: Slight variations in input data, maybe add a comment to explain what we should see as a result. You can also test against scipy examples

algorithms/linfa-nn/src/distance.rs
```diff
@@ -204,4 +226,45 @@ mod test {
...
+        let ab = dist.distance(a.view(), b.view());
+        assert_abs_diff_eq!(ab, 0.95, epsilon = 1e-5);
+
+        let a = arr2(&[[0.3, 0.2, 0.15, 0.10, 0.25], [0.35, 0.15, 0.15, 0.10, 0.25]]);
+        let b = arr2(&[[0.1, 0.2, 0.05, 0.20, 0.45], [0.1, 0.20, 0.05, 0.20, 0.45]]);
+        let ab = dist.distance(a.view(), b.view());
+        assert_abs_diff_eq!(ab, 0.9 + 0.95, epsilon = 1e-5);
```
relf: You could use previous a, b, ab variables

algorithms/linfa-nn/src/distance.rs
```diff
@@ -118,6 +118,27 @@ impl<F: Float> Distance<F> for LpDist<F> {
...
 }
 
+/// Wasserstein or [Earth Mover's](https://en.wikipedia.org/wiki/Earth_mover%27s_distance) distance
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(crate = "serde_crate")
+)]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EarthMoverDist;
```
relf: I would go with `WassersteinDist` as you put it first above and it will be consistent with [scipy impl](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.wasserstein_distance.html).

relf (changes requested):
Thanks for the contribution. Some minor requests...

algorithms/linfa-nn/src/distance.rs
mrsanor: I updated it to WassersteinDist and changed the tests to be according to SciPy where relevant since the APIs are a bit different

relf: Ok I was not aware of the difference of API. Maybe specify once the version of scipy you checked against.

relf (commented):
Could you rebase your changes on current master?

mrsanor:
> Could you rebase your changes on current master?

Done, and I added the SciPy version I compared against to comments

relf (approved):
Thanks!

PR #443, opened 2026-05-28
fix(gmm): use log-sum-exp in estimate_log_prob_resp to prevent overflow

### LINKED ISSUE #442: Numerical overflow in linfa-clustering GMM E-step: estimate_log_prob_resp uses unstable exp-sum-ln normalization
Summary
The GaussianMixtureModel implementation in linfa-clustering still uses a numerically unstable normalization path in estimate_log_prob_resp in release 0.8.1 (and current master).
It computes ln(sum(exp(x))) directly, which overflows when weighted log-probabilities are large positive values.

Affected code
Release 0.8.1:
[https://github.com/rust-ml/linfa/blob/0.8.1/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs#L328-L350](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Master:
[https://github.com/rust-ml/linfa/blob/master/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs#L328-L350](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html)

Current logic in estimate_log_prob_resp:

weighted_log_prob.mapv(|x| x.exp()).sum_axis(Axis(1)).mapv(|x| x.ln())
Why this is a bug
The above is mathematically correct but numerically unstable.
For large positive x, exp(x) overflows to inf.
Then log_prob_norm becomes inf and log_resp can become -inf or nan, which destabilizes EM updates.

Minimal reproducer of the exact numerical issue
This is the same normalization pattern used in estimate_log_prob_resp, isolated:
```
use ndarray::{array, Axis};

fn main() {
    let weighted_log_prob = array![[1000.0_f64, 999.0_f64]];
    let log_prob_norm = weighted_log_prob
        .mapv(|x| x.exp())
        .sum_axis(Axis(1))
        .mapv(|x| x.ln());
    let log_resp = &weighted_log_prob - &log_prob_norm.insert_axis(Axis(1));

    println!("log_prob_norm = {:?}", log_prob_norm); // inf
    println!("log_resp = {:?}", log_resp);           // -inf / nan
}
```
Expected behavior
Normalization in estimate_log_prob_resp should remain finite and stable for large magnitudes, using row-wise log-sum-exp.

Proposed fix
Replace direct exp-sum-ln with log-sum-exp:

```
let weighted_log_prob = self.estimate_weighted_log_prob(observations);
let log_max = weighted_log_prob.map_axis(Axis(1), |row| {
    *row.max().unwrap_or(&F::neg_infinity())
});
let shifted = &weighted_log_prob - &log_max.to_owned().insert_axis(Axis(1));
let log_prob_norm =
    shifted.mapv(|x| x.exp()).sum_axis(Axis(1)).mapv(|x| x.ln()) + &log_max;
let log_resp = weighted_log_prob - log_prob_norm.to_owned().insert_axis(Axis(1));
```
Additional context
I am currently using a local patch with exactly this log-sum-exp change, and it prevents overflow in high-dimensional GMM workloads.

Would maintainers accept a PR with:

- this fix in estimate_log_prob_resp
- a regression test that stresses large weighted log-probability values to ensure finite log_prob_norm and stable responsibilities?

schliffen:
Fixes #442

## Problem

`estimate_log_prob_resp` computed the per-sample log-normaliser as:

    ln(Σ exp(xᵢ))

For tight Gaussians (small covariance, observation near the mean) the weighted
log-probabilities xᵢ can exceed ~709, causing `f64::exp()` to overflow to `inf`.
The subsequent `ln(inf)` produces an infinite `log_prob_norm`, and every `log_resp`
becomes `-inf`, collapsing all responsibilities to NaN / zero and destabilising EM.

## Fix

Replace with the **log-sum-exp trick**:

    max + ln(Σ exp(xᵢ − max))

where `max = max_i(xᵢ)` is computed independently per sample row via
`map_axis(Axis(1), ...)`. After the shift the largest exponent is 0, so
`exp()` never exceeds 1.

**Proof of equivalence:**

    max + ln(Σ exp(xᵢ − max))
    = max + ln(Σ exp(xᵢ) · exp(−max))     [exp(a−b) = exp(a)·exp(−b)]
    = max + ln(exp(−max) · Σ exp(xᵢ))     [factor out constant exp(−max)]
    = max + ln(exp(−max)) + ln(Σ exp(xᵢ)) [ln(a·b) = ln(a)+ln(b)]
    = max + (−max) + ln(Σ exp(xᵢ))        [ln(exp(−max)) = −max]
    = ln(Σ exp(xᵢ))                        ∎

Both the old and new code operate per-row (`Axis(1)`); only the intermediate
numerical range changes.

## Regression test

`test_large_weighted_log_prob_numerical_stability` builds a 70-feature,
1-component GMM with covariance `1e-10 · I`. A point at the cluster mean
produces a log-Gaussian probability of ~741 — above the f64 overflow
threshold of ~709.78. The test asserts `log_prob_norm` is finite and
responsibilities sum to 1.

### DIFF
--- a/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
+++ b/algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
@@ -341,10 +341,13 @@ impl<F: Float> GaussianMixtureModel<F> {
         observations: &ArrayBase<D, Ix2>,
     ) -> (Array1<F>, Array2<F>) {
         let weighted_log_prob = self.estimate_weighted_log_prob(observations);
-        let log_prob_norm = weighted_log_prob
-            .mapv(|x| x.exp())
-            .sum_axis(Axis(1))
-            .mapv(|x| x.ln());
+        // Log-sum-exp trick: shift by per-row max before exponentiating to prevent
+        // overflow when weighted log-probabilities are large (e.g. tight Gaussians).
+        // Mathematically: ln(Σ exp(xᵢ)) = max + ln(Σ exp(xᵢ - max))
+        let log_max = weighted_log_prob
+            .map_axis(Axis(1), |row| row.fold(F::neg_infinity(), |a, &b| a.max(b)));
+        let shifted = &weighted_log_prob - &log_max.clone().insert_axis(Axis(1));
+        let log_prob_norm = shifted.mapv(|x| x.exp()).sum_axis(Axis(1)).mapv(|x| x.ln()) + &log_max;
         let log_resp = weighted_log_prob - log_prob_norm.to_owned().insert_axis(Axis(1));
         (log_prob_norm, log_resp)
     }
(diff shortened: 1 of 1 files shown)


algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -341,10 +341,13 @@ impl<F: Float> GaussianMixtureModel<F> {
         observations: &ArrayBase<D, Ix2>,
     ) -> (Array1<F>, Array2<F>) {
         let weighted_log_prob = self.estimate_weighted_log_prob(observations);
-        let log_prob_norm = weighted_log_prob
-            .mapv(|x| x.exp())
-            .sum_axis(Axis(1))
-            .mapv(|x| x.ln());
+        // Log-sum-exp trick: shift by per-row max before exponentiating to prevent
+        // overflow when weighted log-probabilities are large (e.g. tight Gaussians).
+        // Mathematically: ln(Σ exp(xᵢ)) = max + ln(Σ exp(xᵢ - max))
```
relf, line 346:
```suggestion
        // Log-sum-exp trick to avoid overflow: ln(Σ exp(xᵢ)) = max + ln(Σ exp(xᵢ - max))
```

algorithms/linfa-clustering/src/gaussian_mixture/algorithm.rs
```diff
@@ -341,10 +341,13 @@ impl<F: Float> GaussianMixtureModel<F> {
...
+        let log_max = weighted_log_prob
```
relf, line 347: The lines 346-350 are enought.

relf (changes requested):
While there is a valid fix, I will not accept this verbose Claude PR and AI as co-author. Please could you close this PR and create one yourself with the fix (see below). Just reference the closed PR in yours.

schliffen:
Since in the current PR AI was a coauthor, this is not mergeable, so I'm going to close this branch and address the request in a different branch referring to the current branch.

PR #446, opened 2026-07-01
Bump bhtsne from 0.5.4 to 0.7 in linfa-tsne

LucaCappelletti94:
Bumps `bhtsne` from `0.5.4` to `0.7` in `linfa-tsne`. The `0.6` release switched the embedding dimensionality from a builder parameter to a const generic `tSNE<T, U, const D: usize>`, so the wrapper now dispatches at runtime over `D = 1, 2, 3` for exact t-SNE and `D = 2, 3` for Barnes-Hut (the `Morton<D>` bound limits Barnes-Hut to those dimensions).

bhtsne 0.7 is considerably faster: the optimization loop is parallelized (roughly 4x per epoch at 2k points, scaling with N), the Barnes-Hut tree is now a Morton linear quadtree in a single contiguous arena (repulsive pass 2–3.6x faster, peak memory down by about a third), and there is an FFT-accelerated FIt-SNE fitting path available for future use. All existing tests pass.


LucaCappelletti94:
Closing as I did not realize you had pinned 1.87.

PR #447, opened 2026-07-08
fix(metrics): correct explained_variance residual-variance term

teddytennant:
## What

`SingleTargetRegression::explained_variance` computed the residual-variance numerator as `diff.mapv_into(|x| x * x).sum() - mean_error`, i.e. `Σ(diffᵢ²) − mean(diff)`. That subtracts the raw residual mean (units of y) from a sum of squared residuals (units of y²), which is dimensionally inconsistent.

Explained variance is defined as:

```
EV = 1 − Var(y_true − y_pred) / Var(y_true)
```

so the numerator must be the sum of squared deviations of the residuals from their mean, `Σ(diffᵢ − mean_error)²`. `mean_error = diff.mean()` was already computed but misused. This PR uses it correctly. Numerator and denominator are both sums over the same n samples, so the biased-variance `1/n` factors cancel and the existing `+ 1e-10` denominator guard is preserved.

## Why

For `y_true = [0, 0.1, 0.2, 0.3, 0.4]`, `y_pred = [0.1, 0.3, 0.2, 0.5, 0.7]`:

| quantity | value |
|---|---|
| old `explained_variance` | 0.8 (matches neither R² nor EV) |
| `r2` | −0.8 |
| corrected `explained_variance` | 0.48 |
| `sklearn.metrics.explained_variance_score` | 0.48 |

The old result of 0.8 corresponded to no standard metric. The corrected value matches scikit-learn's reference implementation.

## Testing

- Updated `test_explained_variance_for_single_targets` to assert the corrected value `0.48` (was `0.8`), with a comment citing the scikit-learn reference.
- `test_same` (`explained_variance(ones, ones) == 1.0`) still passes: `diff == 0` ⇒ numerator `0` ⇒ `1 − 0/1e-10 = 1.0`.
- `cargo test -p linfa explained_variance` and the full `cargo test -p linfa --lib metrics_regression` suite (16 tests) pass.
- `cargo fmt --check` and `cargo clippy -p linfa --lib --tests -- -D warnings` are clean.


quietlychris:
I'm generally onboard with this change, but when referencing the sklearn implementation, but would like to see a link to the referenced `sklearn` implementation and docs just to verify alignment there before approving.

teddytennant:
Links:

- docs: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.explained_variance_score.html
- source: https://github.com/scikit-learn/scikit-learn/blob/53302ff996818d108851e31d38a421425026bba5/sklearn/metrics/_regression.py#L1144-L1152

The part that matters:

```python
y_diff_avg = _average(y_true - y_pred, weights=sample_weight, axis=0, xp=xp)
numerator = _average(
    (y_true - y_pred - y_diff_avg) ** 2, weights=sample_weight, axis=0, xp=xp
)

y_true_avg = _average(y_true, weights=sample_weight, axis=0, xp=xp)
denominator = _average(
    (y_true - y_true_avg) ** 2, weights=sample_weight, axis=0, xp=xp
)
```

sklearn averages both terms where we sum both terms, so the 1/n cancels and the results agree. Ran it rather than eyeballing it, sklearn 1.9.0 on the test's own data:

```
>>> explained_variance_score([0.0,0.1,0.2,0.3,0.4], [0.1,0.3,0.2,0.5,0.7])
0.4800000000000002
>>> r2_score([0.0,0.1,0.2,0.3,0.4], [0.1,0.3,0.2,0.5,0.7])
-0.7999999999999994
```

0.48 is what the test now asserts, and -0.8 is what `r2` already returned here before this PR, so the two metrics line up with sklearn on both counts. Put the doc link in the test comment too.

Also rebased on master, it had picked up three commits. `cargo test --release --workspace` is 547 passed, 0 failed. fmt and clippy clean.

PR #448, opened 2026-07-21
Fix multiclass macro F-score (mean of per-class F, not F of the means)

binggao1230:
Multiclass `f_score` / `f1_score` returned the F-beta of the macro-averaged precision and recall, but the macro F-score is the mean of the per-class F-beta scores. Those agree only when every class shares the same precision and recall, so multiclass `f1_score` was wrong for most inputs.

Repro (`y_true=[0,0,1,1,2,2]`, `y_pred=[0,1,1,1,2,2]`): `f1_score()` returned 0.860215; the macro F1 is 0.822222 (per-class F1 = [2/3, 4/5, 1], matching sklearn `f1_score(average="macro")`). Macro precision and recall were already correct, which isolates the error to the F combination.

The fix mirrors the existing `precision`/`recall` macro averaging: average the per-class one-vs-all F-beta. A class the model never gets right has precision = recall = 0 (a 0/0 F); it now counts as 0 (matching sklearn's `zero_division=0`) instead of turning the whole average into NaN. Binary matrices are unchanged.

Two regression tests added; full `cargo test --workspace`, `cargo fmt` and `cargo clippy` pass.

One caveat: for beta != 1 the aggregation is now correct, but linfa's per-class precision and recall are oriented opposite to sklearn's, so `f_score(beta)` lines up with sklearn's `fbeta_score(beta=1/beta)`. That orientation is pre-existing (it affects the binary path too) and left out of scope here; f1 is symmetric so it is unaffected.


quietlychris (approved):
Thanks for looking into this, approved!

Issue #449, opened 2026-07-27
serde support for `linfa-ensemble`


dbrundu:
Hi all. When using `linfa-ensemble` I noticed the two fitted-model types (`EnsembleLearner<M>` and `AdaBoost<M, L>`) cannot be currently persisted, even though the underlying model `M` is, e.g. `EnsembleLearner<DecisionTree<F, L>>`.
The extension of serialization to the ensemble types should involve, in principle, only derive bounds with no manual `impl Serialize`. I'm working on this in my fork and I would be happy to contribute if you want.

PR #450, opened 2026-07-27
Support serde serialization of ensemble models

dbrundu:
A fitted ensemble could not be persisted, even though the weak learners it wraps, like DecisionTree, already can (#449).

This PR adds the `Serialize`/`Deserialize` derives to `EnsembleLearner<M>` and `AdaBoost<M, L>`, following the pattern already used in `linfa-trees` and `linfa-clustering`, without manual impl. The `serde` feature needs to forward to `linfa-trees/serde`, which is required in order to get `RandomForest` serializable. Also, linfa-ensemble/serde is added to the serde cargo check in checking.yml.

Three unit tests are added to cross-check round-trip predictions and parameters between the original model and the serialized/reloaded one, by using serde-json. The prediction comparison for the `RandomForest` model was done by iterating over all the trees by hand, to get reproducible and meaningful comparison and not be dependent by the tree majority vote, which is dependent by the HashMap iteration order.


quietlychris (approved):

quietlychris:
Looks good, thanks for the contribution!

